PR #28, opened 2022-09-18
feat: Added sshnpd systemd script and README.md file 

cconstab:
**- What I did**
Added some details around how to have sshnpd run as a systemd service

**- How I did it**
Created a .service file and README.md

**- How to verify it**
Checked everything works on my test rig 9it does)

Added sshnpd systemd script and README.md file


systemd/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+This directory contains a systemd unit definitions. It runs a the sshnpd component
+of the ssh! no ports in a GNU screen, which can be attached to (for logging etc.)
+with `screen -r`
+
+## Installation
+
+The `sshnpd.service` files should be placed in `/etc/systemd/system` (as root).
+
+Modify the `sshnpd.service` unit to use the appropriate sender and target
```
cpswan: May also need to modify the user if not running on a pi

systemd/README.md
```diff
@@ -0,0 +1,40 @@
...
+The services will then start at the next reboot, or can be started manually
+with:
+
+```bash
+sudo systemctl start sshnpd.service
+```
+
+## Usage
+
+When running there will be three detached screens. `screen -r` will list them:
```
cpswan: There will only be one, and if there is only one `screen -r` will attach to it directly rather than listing it.

cpswan (changes requested):

systemd/README.md
cconstab: agreed

pubspec.lock
```diff
@@ -7,14 +7,14 @@ packages:
       name: _fe_analyzer_shared
```
cpswan: pubspec.lock shouldn't be committed (and should be in .gitignore)

cpswan (changes requested):

pubspec.lock
cconstab: removed and added to .gitignore

cpswan (approved):

PR #29, opened 2022-09-23
fix: changed eol to LF (Unix style) for systemd service file

cconstab:
**- What I did**
Change format of sshnpd.service to Unix style (LF) from windows style (CRLF)

**- How I did it**
VS Code
**- How to verify it**
Checked in editor
fix: changed eol to LF (Unix style) for systemd service file


nickelskevin (approved):
looks good

PR #35, opened 2022-11-29
feat: dockerized sshnpd to allow sshnp into Docker container running anywhere.

cconstab:
*Reason for work*
Docker/Containers at the edge are becoming an easy way to deploy applications and sshnpd in a container allows remotes and secure access to those environments without the normal pain of firewalls/vpns/static IP's. If the container can see the the Internet then you will be able to log into that container.. This is really useful if the device is on a foreign network or cellular network where the IP address cannot be static, or were getting NAT/Firewall ports open is troublesome.

**- What I did**
 Dockerized sshnpd - 
 Created Dockerfile that can be used to build a docker image.
 Docker Image can then be run under Docker/OCI and sshnp used to remotely log into the container remotely
 Added an option to add options, this is really useful if you want to port forward to beyond the container, for instances getting to a GUI/Web interface

**- How I did it**
Created a Dockerfile
Created a GitHub workflow to create armv7/v8/amd64 images automatically
Created a simple DependaBot config to raise PR's as things move

**- How to verify it**
Run extensive tests locally, but would welcome help to create an end to end test to ensure things work before they are pushed

feat: dockerized sshnpd to allow sshnp into Docker container running anywhere.


cpswan (changes requested):
Please take a look at recent commits to see what was needed to bring this into alignment with conventions (terminology and OSSF Scorecard requirements).

Then please update the CHANGELOG.md, and we should be ready to go.

cpswan (approved):

PR #105, opened 2023-03-20
feat: Adding the at_activate application to the sshnoprts repo

cconstab:
**- What I did**
Including the at_activate tool into the sshnp repo so it can be used to activate an generate keys for new sshnp installs.

**- How I did it**
Pulled code form the at_libraries repo
**- How to verify it**
tested by hand

### DIFF
--- a/bin/activate_cli.dart
+++ b/bin/activate_cli.dart
@@ -1,0 +1,6 @@
+import 'package:at_onboarding_cli/src/activate_cli/activate_cli.dart'
+    as activate_cli;
+
+Future<void> main(List<String> args) async {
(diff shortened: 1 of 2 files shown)


cpswan (changes requested):
I think the pubspec needs to be bumped to (at least)`at_onboarding_cli ^1.1.2` to ensure that activate_cli is available.

cconstab:
good catch THANKS!! I actually need to bump it to ^1.2.5 so the keys files get placed in ~/.atsign/keys

cpswan (approved):

PR #115, opened 2023-03-30
fix: Set provenance to false for Docker builds

cpswan:
Docker image files with OCI manifests have been causing problems with routers, so we need to build older style images.

**- What I did**

* Set `provenance: false` so that OCI images aren't used
* Allow the builds to be triggered by `workflow_despatch` so that a new release doesn't have to be cut to rebuild images.

**- How to verify it**

Run the workflow manually and check that routers can use the new images.

fix: Set provenance to false for Docker builds


cconstab (approved):
Perfect thanks Chris

PR #138, opened 2023-05-11
feat: Add notification to workflow

cpswan:
The `other_build` job presently takes around 30m, so nobody should be sat waiting for it. Instead they can get a notification when the job(s) are complete.

**- What I did**

Added a notification job that depends on the existing jobs

**- How I did it**

Copied from at_dockerfiles

**- How to verify it**

[This run](https://github.com/atsign-foundation/sshnoports/actions/runs/4946769866) should generate a notification when it's done in around 30m.

feat: Add notification to workflow

### DIFF
--- a/.github/workflows/multibuild.yaml
+++ b/.github/workflows/multibuild.yaml
@@ -53,3 +53,14 @@ jobs:
         with:
           name: other_binaries
           path: upload
+
+  notify_on_completion:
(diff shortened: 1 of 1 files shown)


cpswan:
Verified as working 
image

cconstab (approved):
Neat

PR #140, opened 2023-05-14
feat: v3.1.0 Includes rendezvous point atSign service migration to ECC ssh keys and more

cconstab:
**- What I did**

### rendezvous service
Added native way of creating a TCP rendezvous for sshnp/sshnpd the new daemon sshrvd runs on a machine with a Public IP and when asked will create a virtual crossover cable on two TCP ports. Those ports are returned to sshnp and in turn sshnpd and sshrv to make the connection. Once established the command to ssh to the remote machine is returned to stdout in the normal manner.
 The -h option is overloaded if the option starts with an `@` it is treated as the sshrvd's atSign if not then old usage model remains. This is kept as if the sshnpd and sshnp can "see" each other then the direct connection will likley have less latency.

### ECC ssh keys
Migration to ECC ssh keys as default, sshnp uses ssh-keygen to create  ed25519 keys using ` -a 100 ` , RSA4098 keys can also be used by using the --rsa option.

### Auto builds
Added binaries to auto build GitHub Actions scripts

**- How to verify it**
Extensive testing by hand, but please also check!

v3.1.0 Includes rendezvous service and migration to ed25519 ssh keys by default


README.md
```diff
@@ -183,155 +157,12 @@ by cron using:
...
-```
-
-Where:
-
-* `-p 3456` corresponds to `-l 3456` from the `sshnp` invocation
-* `-i ~/.ssh/key_for_device.key` is presenting a private key that's trusted
-by the device in its `~/.ssh/authorized_keys`
-* `deviceuser` is the username for the device
-
-### Tunnels in tunnels, an illustration
```
cpswan, line 279: It might be useful to keep the ASCII art illustrations, even if we clear out the ngrok references.

pubspec.yaml
```diff
@@ -4,21 +4,22 @@ version: 3.0.0
 homepage: https://docs.atsign.com/
 
 environment:
-  sdk: '>=2.16.1 <3.0.0'
+  sdk: '>=3.0.0 <4.0.0'
 
 
 dependencies:
```
cpswan, line 10: As this app is using packages rather than publishing a package it would be good to get the dependencies pinned down.

cpswan (dismissed):

pubspec.yaml
```diff
@@ -4,21 +4,22 @@ version: 3.0.0
 homepage: https://docs.atsign.com/
 
 environment:
-  sdk: '>=2.16.1 <3.0.0'
+  sdk: '>=3.0.0 <4.0.0'
```
cpswan, line 7: This will need #136 and #139 to be merged so that Dart 3 is in place for the binaries and Docker images. It might be worth merging their branches into this so it all gets done at once.

cpswan: I've rolled those two PRs into this one

cpswan (dismissed):

cpswan (dismissed):

README.md
cconstab: Will raise a new issue for new diagrams

pubspec.yaml
cconstab: Lets chat about a company wide strategy for this

gkc:
pubsoec sdk should be >=2.12.0 <4.0.0 , shouldn't be requiring minimum sdk of 3.0.0 unless we are using dart-3 features

nickelskevin (approved):

PR #154, opened 2023-05-24
docs: mention sshd

### LINKED ISSUE #151: Mention `sshd`
Documentation should mention that `sshd` needs to be started on `sshnp` client

 'your machine running sshnp has to be running sshd even if only listening on localhost on port 22` - @cconstab

JeremyTubongbanua:
## What I did

Mention that `sshd` has to be running on client `sshnp`
Also mention it has to be running on `sshnp`

closes #151

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Remember to start the daemon on start up using rc.local script or similar, examp
 --device <iot_device_name> -u -s
 ```
 
-Once that has started up you can run the client code from another machine. The client machine has to be running sshd even if only listening on localhost on port 22.
+Once that has started up you can run the client code from another machine. The client machine has to be running sshd even if only listening on localhost on an open port.
 
 ```
 ./sshnp --from <@your_manager_atsign> --to <@your_devices_atsign>  \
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,7 +47,7 @@ Remember to start the daemon on start up using rc.local script or similar, examp
 --device <iot_device_name> -u -s
 ```
 
-Once that has started up you can run the client code from another machine.
+Once that has started up you can run the client code from another machine. The client machine has to be running sshd even if only listening on localhost on port 22.
```
cpswan: It doesn't have to be on port 22, so perhaps better not to mention that (I quite often use 2222).

cpswan (changes requested):

README.md
JeremyTubongbanua: Done in https://github.com/atsign-foundation/sshnoports/pull/154/commits/2a61a1bd7b7d8a4b5d42e17c95ddcad5e15a550c

cpswan (approved):

PR #177, opened 2023-06-07
fix: add armv7l for rpi to the ARCH list

xavierchanth:
**- What I did**

**- How I did it**

**- How to verify it**

fix: add armv7l for rpi to the ARCH list


cpswan (approved):

xavierchanth:
@cconstab @cpswan note that the install_* commands are now updated, however the update_* scripts will not work until the next release. Maybe the install scripts should download the update scripts from github instead of including them in the release

PR #197, opened 2023-06-12
chore: Update LICENSE to 2023

xavierchanth:
**- What I did**

**- How I did it**

**- How to verify it**


cconstab (approved):

cpswan:
The general practice here is for the date to be when code was first created. So we don't need to bump this one, but maybe should bump the archetype (and any recently created repos from it)

PR #216, opened 2023-06-23
fix: fix `Unable to find ssh public key file : false` if -s option is not provided

gkc:
**- What I did**
fix: fix problem where you get an error message `Unable to find ssh public key file : false` if the -s option is not provided

**- How I did it**
Add guard in init to check if value is 'false' before checking if the file exists

**- How to verify it**
Run sshnp without providing the -s option. Should not fail with this error message.


cconstab (approved):
Yay!

PR #222, opened 2023-06-26
feat: Updated docker file, removed errors and put in place error checking

cconstab:
**- What I did**
Investigated https://github.com/orgs/atsign-foundation/projects/8/views/139?pane=issue&itemId=31494613
Found that with debian the command in not needed (used to be in Ubuntu)

**- How I did it**
Remove offending line from Dockerfile
Moved to bash being the default shell
Put in place `set -eux ; ` before each set of bash commands so on any errors the Docker build stops. 

**- How to verify it**
Ran with command still in place and saw the build fail. Removed command and docker build ran fine
Updated docker file, removed errors and put in place bash exit on error, so future errors are caught.


cconstab:
Note container is run as root so sshd can be started in the the startup script. Other ideas welcome.

cpswan (approved):

PR #226, opened 2023-06-28
fix: sshnpd updater patches

xavierchanth:
**- What I did**

- moved post_install function to run before restarting sshnpd
- service restarter now restarts like so:
  `killall -qu "$SSHNP_USER" -r "$BINARY_NAME$"`
  which ensures that the restarter loop keeps running in the background

**- How I did it**

**- How to verify it**

fix: sshnpd updater patches

### DIFF
--- a/scripts/install_sshnpd
+++ b/scripts/install_sshnpd
@@ -259,12 +259,7 @@ setup_service() {
 
 # used to restart the service after an update
 restart_service() {
-  SERVICE_LIST_FILE="$HOME_PATH/.$BINARY_NAME/.service_list";
-  while read -r SERVICE;
-  do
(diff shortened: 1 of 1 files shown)


cconstab:
I get an odd error on pi

`spi0(183): Operation not permitted`

running with bash -x

```
Installation complete!
+ '[' update = update ']'
+ restart_service
+ killall -qu pi -r 'sshnpd$'
spi0(183): Operation not permitted
pi@xmt-pi:~ $
```

should read 

` killall -q -u pi -r 'sshnpd$'`

typo on line 262

cconstab (approved):
tested on Pi works fine

PR #230, opened 2023-06-29
fix: pass all options on sshnp@device to sshnp binary

### LINKED ISSUE #186: Carry over all flags from generated sshnp@<device>
Not all flags in the generated binaries for sshnp and sshnpd work... we should make them configurable.

xavierchanth:
closes #186 
**- What I did**

Rather than manually parsing the arguments on this script, just pass them down to sshnp.
Using sshnp's usage as this script's usage, but I've used grep to remove lines with "(mandatory)" since those lines are filled with default values in this script.

**- How I did it**

**- How to verify it**

Make a copy of sshnp-full.sh and replace the values in FROM, TO, HOST, PUBLIC_KEY.

Example:
```sh
#!/bin/bash
#v2.0.0
BINARY_NAME="sshnp";
FROM="@xavierchanth";
TO="@xchan";
HOST=""; #<ommitted intentionally in this example (this is the rendezvous atSign)>
PUBLIC_KEY="sshnp@xchan.pub";

usage() {
  "$HOME/.local/bin/$BINARY_NAME" --help | grep -v -e "(mandatory)" -e "FormatException";
  echo "Note: previously device name was a positional argument, please specify it with -d."
}

if [ "$#" -eq 0 ]; then
  usage;
  exit 1;
fi

if [ "$1" = "--help" ]; then
  usage;
  exit 0;
fi

"$HOME/.local/bin/$BINARY_NAME" -f "$FROM" -t "$TO" -h "$HOST" -s "$PUBLIC_KEY" "$@";
```

Run the above script with -d <devicename>

fix: pass all options on sshnp@device to sshnp binary

### DIFF
--- a/templates/client/sshnp-full.sh
+++ b/templates/client/sshnp-full.sh
@@ -11,4 +11,14 @@ usage() {
   echo "Note: previously device name was a positional argument, please specify it with -d."
 }
 
+if [ "$#" -eq 0 ]; then
+  usage;
+  exit 1;
+fi
+
+if [ "$1" = "--help" ]; then
+  usage;
+  exit 0;
+fi
+
 "$HOME/.local/bin/$BINARY_NAME" -f "$FROM" -t "$TO" -h "$HOST" -s "$PUBLIC_KEY" "$@";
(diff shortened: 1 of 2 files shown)


xavierchanth:
Note: -v is an inversion flag in grep, disposes of matching lines instead of keeping them

JeremyTubongbanua (approved):

PR #240, opened 2023-07-05
feat: config files

xavierchanth:
**- What I did**

Added config file support which follows the [dotenv spec](https://hexdocs.pm/dotenvy/dotenv-file-format.html):
- Option: `--config-file`
- Expected value: `path to file` (relative or absolute)

Notes:

If a config file is supplied, it is read and parsed first, then the rest of the args are read.
So command line args will always override any duplicate args in the file.

**- How I did it**

**- How to verify it**

Demo:

Note: the second attempt to sshnp, the error is expected since I don't own the atSign that I passed into the config file. Third attempt shows that command line args override the value in the file contents.

https://github.com/atsign-foundation/sshnoports/assets/33691921/60dadb6b-b629-4ea9-acd5-081c9260fc4d

feat: config files

### DIFF
--- a/packages/sshnoports/lib/config_util.dart
+++ b/packages/sshnoports/lib/config_util.dart
@@ -1,0 +1,45 @@
+import 'dart:io';
+
+import 'package:sshnoports/sshnp_args.dart';
+
+List<String> parseConfigFile(String fileName) {
+  List<String> args = <String>[];
+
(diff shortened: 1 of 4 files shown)


JeremyTubongbanua:
Very cool

gkc (approved):
V cool! LGTM

PR #259, opened 2023-07-17
feat: SSHNPD Python

Xlin123:
**- What I did**
I've made decent progress and I'm hoping to get some feedback. I don't think I'm setting up the SSH tunnel right... 

**- How I did it**
Using at_python notifications
Using paramiko for SSHClient

**- How to verify it**

sshnpd python


JeremyTubongbanua:
@Xlin123 Just so I can start thinking about the tests, what will the NPD container end up doing?

Is it just a `python sshnpd.py` that I would have to run with args?

Xlin123:
> @Xlin123 Just so I can start thinking about the tests, what will the NPD container end up doing?
> 
> Is it just a `python sshnpd.py` that I would have to run with args?

yea `python3 sshnpd.py --atsign .....` etc.

JeremyTubongbanua:
> > @Xlin123 Just so I can start thinking about the tests, what will the NPD container end up doing?
> > Is it just a `python sshnpd.py` that I would have to run with args?
> 
> yea `python3 sshnpd.py --atsign .....` etc.

Sounds good. Great work @Xlin123 :)

cpswan:
Do we even need a container for tests @JeremyTubongbanua @Xlin123 or can this be run directly?

Also can we `#!/usr/bin/env python3` at the start of the script and make it executable?

JeremyTubongbanua:
> Do we even need a container for tests @JeremyTubongbanua @Xlin123 or can this be run directly?
> 
> Also can we `#!/usr/bin/env python3` at the start of the script and make it executable?

Not sure, will have to try. Is there a disadvantage to just running inside of a docker container? I am currently running everything in a docker container so I can run the test simply with `docker compose up` then let the ENTRYPOINTS handle the test. Then there's a nifty flag `--exit-code-from=<service>` which gets the exit code from a specific container.

gkc:
@Xlin123 you may wish to keep the paramiko sshclient as a fallback in case openssh isn't available or isn't at the version you need. See also [this ticket](https://github.com/atsign-foundation/sshnoports/issues/299) and [this comment](https://github.com/atsign-foundation/sshnoports/issues/299#issuecomment-1665385052)

Xlin123:
> @Xlin123 you may wish to keep the paramiko sshclient as a fallback in case openssh isn't available or isn't at the version you need. See also [this ticket](https://github.com/atsign-foundation/sshnoports/issues/299) and [this comment](https://github.com/atsign-foundation/sshnoports/issues/299#issuecomment-1665385052)

Sounds good, I am making it a separate method. I'll try fixing the paramiko version for now and look into that next week.

Xlin123:
I'm done refactoring so I'm adding reviewers, take a look when you can. Still doing work on the sys call version instead of paramiko.

JeremyTubongbanua:
Merge the most recent version of trunk to your branch @Xlin123 ..

@XavierChanth extended the time it takes for sync to happen as v3.2.0 and v3.1.2 sshnpd's use sync. Was fixed in #309

.gitignore
```diff
@@ -1,14 +1,27 @@
 # Files and directories created by pub.
 .dart_tool/
 .packages
+
+#vscode
+.vscode
```
cpswan: We might need something more targeted here per the discussion in #311

packages/sshnpdpy/lib/sshnpdclient.py
```diff
@@ -0,0 +1,314 @@
...
+                
+                if event_type == AtEventType.UPDATE_NOTIFICATION:
+                    queue.put(at_event)
+                    sleep(1)
+                if event_type != AtEventType.DECRYPTED_UPDATE_NOTIFICATION:
+                    continue
+                key = event_data["key"].split(":")[1].split(".")[0]
+                decrypted_value = str(event_data["decryptedValue"])
+                if key == "privatekey":
+                    # print(
```
cpswan: Even if they're commented, let's get rid of print statements that have been used during development/testing.

Anything that does need to print should use a logger.

cpswan (changes requested):
This is great progress @Xlin123. Good work getting it this far.

I've left a few comments on nits that should be resolved before this gets merged.

Can you please make this PR Draft for now, and then use the removal of the Draft marker as a sign that you're ready for it to be merged.

.gitignore
Xlin123: Updated to ignore all files except launch.json.

packages/sshnpdpy/lib/sshnpdclient.py
Xlin123: Added a logger.

packages/sshnpdpy/lib/sshnpdclient.py
```diff
@@ -0,0 +1,318 @@
...
+
+    def _reverse_ssh_client(self, ssh_list: list, private_key: str):
+        local_port = ssh_list[0]
+        port = ssh_list[1]
+        username = ssh_list[2]
+        hostname = ssh_list[3]
+        self.logger.info("ssh session started for " + username + " @ " + hostname + " on port " + port)
+        ssh_client = SSHClient()
+        ssh_client.load_system_host_keys(f"{self.ssh_path}/known_hosts")
+        ssh_client.set_missing_host_key_policy(AutoAddPolicy())
```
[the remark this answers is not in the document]
Xlin123: Whoa. Thanks mr github-code-scanning. I'll check it out.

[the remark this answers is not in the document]
Xlin123: The only other option I have with it is the WarningPolicy, which will still accept the connection but won't write the keys on the device key file. It also prints out this and I can't make it quieter. 
```
/home/xavier/.local/lib/python3.10/site-packages/paramiko/client.py:889: UserWarning: Unknown ssh-ed25519 host key for [85.239.53.170]:35869: b'fc4ea32434216548fae44d14e6525db3'
  warnings.warn(
  ```
  
  My opinion is to change it to WarningPolicy. 

  Thoughts? @cpswan

**Comment by Xlin123** [comment]
@XavierChanth @cconstab review when you can, ik you're both very busy.

File: .vscode/launch.json [hunk 4]
```diff
@@ -12,6 +12,22 @@
...
+      "program": "sshnpd.py",
+      "console": "integratedTerminal",
+      "justMyCode": true,
+      //replace your arguments with the ones you want to use
+      "args": [
+        "--atsign", "@chess69",
+        "--manager", "@chess69lovely",
+        "--device", "dev",
+        "--username", "xavier"
+      ]
```
xavierchanth: Is there a way to configure this in a gitignored .env file?

packages/sshnpdpy/lib/sshnpdclient.py
```diff
@@ -0,0 +1,318 @@
...
+        elif os.name == "nt":  # Windows
+            home_dir = os.path.expanduser("~")
+        else:
+            raise NotImplementedError("Unsupported operating system")
+        self.ssh_path = f"{home_dir}/.ssh"
+        
+    def start(self):
+        username_key = SharedKey(
+        "username", AtSign(self.atsign), AtSign(self.manager_atsign))
+        self.at_client.put(username_key, self.username)
```
xavierchanth: This should only happen if the -u flag is supplied

xavierchanth (commented):
Worth having a discussion with @gkc as to some of the recent behaviour changes related to the -u flag

packages/sshnpdpy/lib/sshnpdclient.py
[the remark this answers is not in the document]
Xlin123: Changing to WarningPolicy.

packages/sshnpdpy/lib/sshnpdclient.py
Xlin123: Made changes, will reach out to Gary today to talk about it.

.vscode/launch.json
Xlin123: Pushed changed for this, let me know if it works...

xavierchanth (changes requested):
Two things:

1. For the launch.json's .env, it would be best to match the format used for [config files](https://github.com/atsign-foundation/sshnoports/blob/trunk/packages/sshnoports/templates/config/sshnp-config-template.env) that way they are reusable.

2. -u in dart is a boolean flag not a string input, with username being derived by the program automatically, for consistency, my suggestion is that we try to do the same in python.

```dart
String? getUserName({bool throwIfNull = false}) {
  Map<String, String> envVars = Platform.environment;
  if (Platform.isLinux || Platform.isMacOS) {
    return envVars['USER'];
  } else if (Platform.isWindows) {
    return envVars['USERPROFILE'];
  }
  if (throwIfNull) {
    throw ('\nUnable to determine your username: please set environment variable\n\n');
  }
  return null;
}
```

Xlin123:
@XavierChanth let me know if those changes are proper.

packages/sshnpdpy/sshnpd-config-template.env
```diff
@@ -0,0 +1,11 @@
+#Template Config File for SSHNPD Python
+
+#Receving or (device) atsign
+ATSIGN=@chess69
```
xavierchanth: change to TO=

packages/sshnpdpy/sshnpd-config-template.env
```diff
@@ -0,0 +1,11 @@
...
+
+#Sending or (manager) atsign
+MANAGER=@chess69lovely
```
xavierchanth: change to FROM=

packages/sshnpdpy/sshnpd-config-template.env
Xlin123: 👍 🦖

packages/sshnpdpy/sshnpd-config-template.env
Xlin123: 👍 🌮

xavierchanth:
Merge trunk into your branch again @Xlin123

cpswan (approved):

xavierchanth (approved):

PR #265, opened 2023-07-20
fix: dependency upgrades + cleanup error messages in sshnp

### LINKED ISSUE #264: Ugly errors on 3.4.0 when the conection times out
### Describe the bug

on a connection fail event this is what os printed out..

```
cconstab@tarial ~ % $(sshnp@ssh_1 -d raksign1)    
Error: sshnp: connection timeout
Stack Trace: #0      SSHNPImpl.run (package:sshnoports/sshnp/sshnp_impl.dart:359)
<asynchronous suspension>
#1      main (file:///Users/chant/src/af/sshnoports/packages/sshnoports/bin/sshnp.dart:23)
<asynchronous suspension>

cconstab@tarial ~ %
```

Which is ugly and show the developers env which is super ugly too

### Steps to reproduce

1. First  set up a device
2. Then switch off the device
3. And then try to connect to it

### Expected behavior

Simple time out with reason for timeout if known

### Screenshots

_No response_

### Smartphones

- Device: Mac M1
- OS: OSX

_No response_

### Additional context

_No response_

xavierchanth:
**- What I did**

closes #264 

I get the params object in the main function now so I can check if the verbose flag is set.
Then only print stack trace if verbose is true.

**- How I did it**

**- How to verify it**

chore: cleanup error messages in sshnp

### DIFF
--- a/packages/sshnoports/bin/sshnp.dart
+++ b/packages/sshnoports/bin/sshnp.dart
@@ -7,12 +7,16 @@ import 'package:at_utils/at_logger.dart';
 // local packages
 import 'package:sshnoports/sshnp/sshnp.dart';
 import 'package:sshnoports/sshnp/cleanup.dart';
+import 'package:sshnoports/sshnp/sshnp_params.dart';
 
 void main(List<String> args) async {
   AtSignLogger.root_level = 'SHOUT';
   SSHNP? sshnp;
+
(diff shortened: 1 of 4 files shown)


xavierchanth:
Upgraded dependencies due to a breaking change in at_persistence_spec / at_persistence_secondary_server.

trunk tests are expected to fail since they are using the old dependencies.

gkc (approved):

PR #271, opened 2023-07-21
ci: dockerhub

xavierchanth:
**- What I did**
separate dockerhub automated builds from latest tagging

**- How I did it**

**- How to verify it**

ci: dockerhub


cconstab (approved):

xavierchanth:
cc: @cpswan
We made this change because when I did pre-releases of sshnp it got pushed to latest on dockerhub, which subsequently caused a breakage for someone using the image.

Issue #184, opened 2023-06-09
sshnp - list devices


xavierchanth:
### Is your feature request related to a problem? Please describe.

Can't really tell what devices can be connected to, especially in an organization where you may have had an admin setup sshnpd for you.

### Describe the solution you'd like

A subcommand / flag which allows you to list available devices for connecting to.

### Additional context

Depending on the implementation, the client may have to state that only devices with version greater than x.x.x will be listed...
```[tasklist]
### Tasks
- [x] sshnpd - internal schedule with ttl to create/refresh device key (json structure with devicename, metadata of installed version, and room for info)
- [x] sshnpd - monitor for ping notifications and respond with a device name notification
- [x] sshnp - grab key list for all clients
- [x] sshnp - ping notification and wait up to 5 seconds for response
- [ ] https://github.com/atsign-foundation/sshnoports/issues/371
```

xavierchanth:
A notification from clients which daemons respond to, or put an atKey shared with the client/manager atsign

cconstab:
I like the notification route best.. This way only listening sshnpd's respond.. using an atKey could leave some "junk" in place.

gkc:
I'd like to see a list of all of my devices, with the ones which respond marked as such

Junk is a valid concern but could be mitigated with a suitable ttl e.g. N days?

i.e. a combination of
- sshnpd makes a key for the device saying what version it is running and when it was last started
- sshnp retrieves that list
- sshnp also sends a 'ping' notification which all sshnpds listen for
- running sshnpds respond to ping notifications

xavierchanth:
I've started the work on this based on what @gkc listed above. Will resume in this sprint after addressing some of the other issues first.

xavierchanth:
The initial implementation is available in #285 
Enough to allow me to progress with #234

xavierchanth:
Last step is less related to this issue so I've made it a separate issue so that this can be closed

PR #285, opened 2023-07-26
feat: ping/heartbeat (list devices)

xavierchanth:
**- What I did**

### Completed the following tasks from #184
- [x] sshnpd - internal schedule with ttl to create/refresh device key (json structure with devicename, metadata of installed version, and room for info)
- [x] sshnpd - monitor for ping notifications and respond with a device name notification
- [x] sshnp - grab key list for all clients
- [x] sshnp - ping notification and wait up to 5 seconds for response

In addition, adding some of my work on the python management scripts to help with testing this code.

Python scripts:

- Initial (experimental) SSHNP / SSH / SFTP Client written using paramiko
- A script which uses the aforementioned clients to update sshnpd installations through sshnp

**- How I did it**

**- How to verify it**

feat: ping/heartbeat (list devices)


packages/sshnoports/lib/common/utils.dart
```diff
@@ -45,13 +45,10 @@ Future<bool> fileExists(String file) async {
...
+const String asciiMatcher = r'[a-zA-Z0-9_]{0,15}';
+
 bool checkNonAscii(String test) {
-  var extra = test.replaceAll(RegExp(r'[a-zA-Z0-9_]*'), '');
-  if ((extra != '') || (test.length > 15)) {
-    return true;
-  } else {
-    return false;
-  }
+  return RegExp(asciiMatcher).allMatches(test).first.group(0) != test;
```
xavierchanth, line 51: The first match will always be the whole thing (if it matches), so we just compare the first match to test, if there are no matches, the left always resolves to empty string.

gkc:
@XavierChanth Looks good! Will you be adding test(s) to the e2e test pack? Also while I was doing my code walkthrough I noticed something we missed during the refactorings and I pushed [a commit](url) to address that

I will not approve and merge yet as the PR is marked as a draft.

PS Some checks failed with
```
sshnpd  | SEVERE|2023-07-26 11:05:10.923584|OnboardingCli|Caught exception: Exception: Failed connecting to root server url root.atsign.org on port 64
```
I will re-run the checks

xavierchanth:
CleanShot 2023-07-29 at 08 20 53@2x

playpen is my running instance, 
dead is an instance I spun up and then killed to see if the atKey was put in place correctly

Then I switched to interiorwolverine for the device atSign (which is not running any sshnpds) and it shows no devices, as well as a prompt hinting that they should try updating sshnpd

xavierchanth:
Made the output a little bit nicer: 
CleanShot 2023-07-29 at 08 25 47@2x

xavierchanth:
I did not complete the task that ensures device names are unique, Want to make sure we do that one right. Initially we thought of binding a device name to a client ID, but we also have the option of preventing subsequent sshnpd startups if a device name is already in use. i.e. check for an atKey and if there is an atKey try to do a ping (probably 2-3 times to be safe) then startup sshnpd

gkc (approved):

xavierchanth:
@gkc did you confirm this to be working before merging? I still had some testing to do

gkc:
I did not; I only reviewed the code, it looked complete as was not draft PR and tasks had been ticked off 

But now that you mention it I do now remember that you had said something about it being tricky to test

Let's just create another PR for follow-up testing & test automation

xavierchanth:
> I did not; I only reviewed the code, it looked complete as was not draft PR and tasks had been ticked off
> 
> But now that you mention it I do now remember that you had said something about it being tricky to test
> 
> Let's just create another PR for follow-up testing & test automation

My mistake, maybe I did complete testing... since I marked it as ready for review... I honestly don't remember how far along I got with testing. It never hurts to test twice though.

PR #315, opened 2023-08-09
feat: device renaming for sshnpd via installer

xavierchanth:
**- What I did**

Added `--rename` flag to the sshnpd installer.
It also will prompt for client atSign and the new name if not provided via command line args.
(the assumption here is that you only ever need to have 1 sshnpd instance per client atSign running on a given unix user).

I also cleaned up the behaviour of the install, so that if you attempt to install over an existing client it will count as an update + rename for that client.

Lastly, there was a bug where multiple entries for the same client could end up in the service_list tracker file, so I prevented that from happening for both install and rename.

**- How I did it**

**- How to verify it**

feat: device renaming for sshnpd via installer

### DIFF
--- a/packages/sshnoports/scripts/install_sshnpd
+++ b/packages/sshnoports/scripts/install_sshnpd
@@ -25,6 +25,7 @@ usage() {
   echo "Usage: $0 [options]"
   echo "General options:"
   echo "  -u, --update                Update all services instead of installing"
+  echo "      --rename                Rename device for client/device pair with the new name"
   echo "  -l, --local <path>          Install using local zip/tgz"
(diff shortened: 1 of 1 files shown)


cpswan (approved):

cpswan:
@XavierChanth is there going to be a corresponding docs update with an example of how to use that (in the context of why somebody might need that)?

xavierchanth:
> @XavierChanth is there going to be a corresponding docs update with an example of how to use that (in the context of why somebody might need that)?

Yes, I was going to also add the doctor and uninstall scripts before updating the docs so it can be done once

PR #320, opened 2023-08-10
refactor: installers

xavierchanth:
**- What I did**
Refactoring of sshnpd installer:
- Broke up setup_service script into 3 smaller functions which are called by setup_service:
  - setup_service_script
  - setup_service_job
  - setup_service_entry 
- Added comments to every function explaining what it does
- Replaced the download function with get_main_binaries function
- Broke out the various methods of getting the main binaries into their own functions
  - build_dart_binaries when building from the repo
  - copy_local_archive when providing a local tar/zip
  - download_archive when downloading from github releases (default)
  - unpack_archive to unpack the tar/zip of the previous two functions
- Added a guard with parse_requirements / check_requirements functions
  - parse_requirements stores whether each dependency is available on the system
  - check_requirements echos out any errors due to missing dependencies
    - check for ssh, sshd, and sshd is running (by running ssh 0)
    - cron is a must
    - curl if the download_archive method is expected to be called (based on user input)
    - logrotate or tmux depending on the installation method
 

Added similar refactorings to install_sshnp so that it mirrors the changes for install_sshnpd

- Added the sync based end2end tests to their own job below (which is identical to the first one) 
  - The sync based end2end job is limited to max-parallel: 1, only allowing 1 sync to be taking place at a time to hopefully reduce flaking

**- How I did it**

**- How to verify it**

I tested this version of the installer locally, it can be tested by running:
```bash
bash -c "$(curl -fsSL https://raw.githubusercontent.com/atsign-foundation/sshnoports/refactor-install-sshnpd/packages/sshnoports/scripts/install_sshnpd)"
```

refactor: sshnpd installer


xavierchanth:
Went back and disabled the ssh 0 check, since I can't consistently guarantee that it won't try to ask for password authentication. There must be a better way... the obvious solution is netcat, but that isn't necessarily available on all systems.

cpswan (dismissed):

xavierchanth:
I want to do more extensive testing before merging

xavierchanth:
Note that in [438ca4b](https://github.com/atsign-foundation/sshnoports/pull/320/commits/438ca4b2464f0971cee81ca64ccbb75f91911170) I've disabled SC2009 for a couple lines, which suggests to try pgrep if available. (which we are doing and using the ps -> grep as a fallback.)

JeremyTubongbanua:
Did you forget `v3.1.2` for the sync e2e tests?

JeremyTubongbanua (approved):

xavierchanth:
> Did you forget `v3.1.2` for the sync e2e tests?

No, @gkc and I have decided to remove them

PR #325, opened 2023-08-10
ci: Increase wait times

xavierchanth:
**- What I did**

Default time (i.e. no sync)
7s -> 10s

Sync clients
30s -> 60s

**- How I did it**

**- How to verify it**

ci: Increase wait times

### DIFF
--- a/.github/composite/setup_entrypoints/action.yaml
+++ b/.github/composite/setup_entrypoints/action.yaml
@@ -26,6 +26,6 @@ runs:
       shell: bash
       working-directory: tests/end2end_tests/tools/configuration
       run: |
-        ./setup-sshnp-entrypoint.sh ${{ inputs.devicename }} ${{ inputs.sshnp_atsign }} ${{ inputs.sshnpd_atsign }} ${{ inputs.sshrvd_atsign }} ${{ inputs.wait_time }}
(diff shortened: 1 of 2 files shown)


cpswan (dismissed):

xavierchanth:
@gkc @cpswan I'll keep tweaking until I can get 2-3 runs with no fails

xavierchanth:
10s on no sync has passed twice, 60s on the sync clients has passed once, I raised the max-parallel to 100. If this last run of the tests pass then I think we are good.

xavierchanth:
Down to just 1x 10s test failing
Running again

JeremyTubongbanua:
The 10s sleep isn't actually there. the tests without the number at the end aren't sleeping at all at the moment

image

xavierchanth:
> The 10s sleep isn't actually there. the tests without the number at the end aren't sleeping at all at the moment
> 
> image

That's because the 10s is the default fallback value in the composite workflow. i.e. not specifying a time will default to 10s.

xavierchanth:
Oh I see what you mean...

xavierchanth:
CleanShot 2023-08-11 at 13 23 07@2x

Fixed in [1c386ca](https://github.com/atsign-foundation/sshnoports/pull/325/commits/1c386ca2f82e55385a0e66b89af3367ab6011783)

cpswan (approved):

PR #329, opened 2023-08-11
feat: make sshnpd ping/heartbeat an opt-in feature

xavierchanth:
**- What I did**

- Added a negatable `--hidden flag` which is true by default
- You must specify --no-hidden if you want the device to show up in sshnp --list-devices
- Added --no-hidden to the sshnpd installer which will then add the flag to the `sshnpd@<client atSign>` service script
- Updated the sshnpd templates and rename script in accordance with that

Question for @gkc and @cconstab:

- Instead of "hidden" and "no-hidden" should I make the flag "discoverable" or something similar?
Note that discoverable is awkward to type (and -d is already taken).

**- How I did it**

**- How to verify it**

feat: make sshnpd ping/heartbeat an opt-in feature

### DIFF
--- a/packages/sshnoports/lib/sshnpd/sshnpd.dart
+++ b/packages/sshnoports/lib/sshnpd/sshnpd.dart
@@ -33,7 +33,7 @@ abstract class SSHNPD {
   abstract final SupportedSshClient sshClient;
 
   /// Whether this device should be hidden from sshnpd or not
-  abstract final bool shouldBroadcastInfo;
+  abstract final bool isPingable;
 
   /// true once [init] has completed
   @visibleForTesting
@@ -47,7 +47,7 @@ abstract class SSHNPD {
     required String device,
     required String managerAtsign,
     required SupportedSshClient sshClient,
-    bool shouldBroadcastInfo = false,
+    bool isPingable = false,
   }) {
     return SSHNPDImpl(
       atClient: atClient,
@@ -56,7 +56,7 @@ abstract class SSHNPD {
       device: device,
       managerAtsign: managerAtsign,
       sshClient: sshClient,
-      shouldBroadcastInfo: shouldBroadcastInfo,
+      isPingable: isPingable,
     );
   }
 
--- a/packages/sshnoports/lib/sshnpd/sshnpd.dart
+++ b/packages/sshnoports/lib/sshnpd/sshnpd.dart
@@ -32,26 +32,32 @@ abstract class SSHNPD {
   /// The ssh client to use when doing reverse ssh
   abstract final SupportedSshClient sshClient;
 
+  /// Whether this device should be hidden from sshnpd or not
+  abstract final bool isHidden;
+
   /// true once [init] has completed
   @visibleForTesting
   bool initialized = false;
 
-  factory SSHNPD(
-      {
-      // final fields
-      required AtClient atClient,
-      required String username,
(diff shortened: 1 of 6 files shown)


cconstab:
Echo -e  perhaps?

xavierchanth:
Not exactly an echo, it doesn't respond with device info...

What about broadcast? As in: broadcast device info, I believe -b is available

gkc:
--pingable

xavierchanth:
Now --pingable and -p

gkc:
not compiling currently

gkc:
As currently implemented, this prevents clients from discovering which version the sshnpd is running unless we add additional request and response notification handlers

Additionally, there is now an inconsistency, since we are continuing to create the `username.$device.sshnp` record (for backwards compatibility) thus a client atSign can still discover all of the connected devices by doing a `scan` on their atServer

We need to think about this a little more.

@cconstab since we already have the existing sshnpd's creating the `username.$device.sshnp` record, are you OK with not making the ping configurable in v3.4.0 i.e. just going with the current implementation? And then revisiting this topic when we have a little more think time available?

cconstab:
So -u on sshnpd is also optional for the same security reason.

Same with -s

On sshnpd

We should follow the same pattern or just drop the ping for now I think

xavierchanth:
> not compiling currently

Oops, I missed a couple of renames

gkc:
> So -u on sshnpd is also optional for the same security reason.
> 
> Same with -s
> 
> On sshnpd
> 
> We should follow the same pattern or just drop the ping for now I think

ACK

Note that (1) the behaviour for the -s flag may have been designed but has never been implemented (2) the behaviour for the -u flag was broken during refactoring a few months back

I'll do a PR now to resolve (1) and (2) and to also (3) remove the 'ping' notification handler

xavierchanth:
I will resolve these merge conflicts at my next sit-down, it looks like I need to be careful with some of the changes.

gkc:
> I will resolve these merge conflicts at my next sit-down, it looks like I need to be careful with some of the changes.

I'll fold the opt-in into the PR I've just started as my refactoring was merged a little while ago and I am fixing the other stuff also

gkc:
I note that the current implementation is that the client is required to already **know** the device name in order to send the ping request anyway. To put it another way, ping requests are currently unicast, not multicast.

Given that, and the need for clients to be able to ask for information like version numbers, I believe that
- pings should be handled if they are received, but the responses should have a very short ttl and should include only the version number
- device info like username and current version should not be shared unless the `-u` flag is provided (which now means `un-hide` 😀)

gkc:
@XavierChanth I note that there is other stuff in this PR but I'd suggest you back out your changes to the dart programs and also the --pingable flag ... I'm re-using the `-u` flag since it is already related to device visibility, and making it now mean "**u**n-hide" rather than "share **u**sername"

gkc:
Other PR in https://github.com/atsign-foundation/sshnoports/pull/331

xavierchanth:
> @XavierChanth I note that there is other stuff in this PR but I'd suggest you back out your changes to the dart programs and also the --pingable flag ... I'm re-using the `-u` flag since it is already related to device visibility, and making it now mean "**u**n-hide" rather than "share **u**sername"

Will do

xavierchanth:
This PR is made obsolete by #331 and #333

PR #333, opened 2023-08-13
feat: Update sshnpd-service-template and improve renaming / modification

xavierchanth:
**- What I did**

See the discussion in: https://github.com/atsign-foundation/sshnoports/pull/329

Updated the headless sshnpd.sh and sshrvd.sh service templates. The installer for sshnpd now has an additional `--args` option where non-mandatory sshnpd args will be set ("-v -u -s" by default), supplying an empty string to args (`--args ""`) will allow the exclusion of all non-mandatory fields.

**- How I did it**

**- How to verify it**

feat: Update sshnpd-service-template and improve renaming / modification


xavierchanth:
@gkc this extracts my relevant changes out of #329

packages/sshnoports/scripts/install_sshnpd
```diff
@@ -35,6 +35,7 @@ usage() {
   echo "  -d, --device <address>      Device address (e.g. @alice_device)"
   echo "  -n, --name <device name>    Name of the device"
   echo "  -v, --version <version>     Version to install (default: latest)"
+  echo "      --args <args>           Additional arguments to sshnpd (\"-v -s -u\" by default)"
```
gkc: You mentioned in the PR description, `supplying an empty string to args (--args "") will allow the exclusion of all non-mandatory fields.` ... I don't think that will be clear to the user from reading this usage help; maybe add that phrase to the message as well?

gkc: How does a user get to know what `-v`, `-s` and `-u` are for?

Given recent discussions with @cconstab re #329 and #331 I feel like `-s` and `-u` should, by default, _not_ be set, so that the default security posture is the most defensive; but that needs to be accompanied by clear explanations of what those flags do during the install, so that the user gets the information and can then come back to it later on to adjust the sshnpd's arguments

xavierchanth: There's no good way for them to know... unless we duplicate the output of sshnpd into this script

xavierchanth:
Added the additional usage docs, and made it only -v by default

packages/sshnoports/scripts/install_sshnpd
```diff
@@ -35,6 +35,11 @@ usage() {
   echo "  -d, --device <address>      Device address (e.g. @alice_device)"
   echo "  -n, --name <device name>    Name of the device"
   echo "  -v, --version <version>     Version to install (default: latest)"
+  echo "      --args <args>           Additional arguments to sshnpd ("-v" by default)"
+  echo "      Possible args:"
+  echo "        -s, --[no-]sshpublickey      Update authorized_keys to include public key from sshnp"
+  echo "        -u, --[no-]username          Send username to the manager to allow sshnp to display username in command line"
```
gkc: Please use the new description associated with the -u which is now --un-hide (although still also aliased to --username for back compatibility

gkc (changes requested):

packages/sshnoports/scripts/install_sshnpd
gkc: From sshnpd params:

parser.addFlag(
      'un-hide',
      abbr: 'u',
      aliases: const ['username'],
      defaultsTo: false,
      help:
          'When set, makes various information visible to the manager atSign - e.g. username, version, etc',
    );

packages/sshnoports/scripts/install_sshnpd
```diff
@@ -38,7 +38,7 @@ usage() {
   echo "      --args <args>           Additional arguments to sshnpd ("-v" by default)"
   echo "      Possible args:"
   echo "        -s, --[no-]sshpublickey      Update authorized_keys to include public key from sshnp"
-  echo "        -u, --[no-]username          Send username to the manager to allow sshnp to display username in command line"
+  echo "        -u, --[no-]un-hide          When set, makes various information visible to the manager atSign - e.g. username, version, etc"
```
gkc: (Sorry!) Missing a space ... help text output would not be aligned.

gkc (changes requested):

packages/sshnoports/scripts/install_sshnpd
xavierchanth: Good catch, thanks

gkc (approved):

PR #334, opened 2023-08-13
ci: e2e tests for the installer

### LINKED ISSUE #317: SSH No Ports: E2E Test - Installer Script
End-to-end test where we use the installer script to install latest release of ssh no ports and then use the installed binaries and generated custom scripts to test basic functionality of SSH No ports with prod atSigns

xavierchanth:
closes #317 

**- What I did**

Working on e2e tests so we can comfortably merge #320 

Added end2end tests that test the sshnp and sshnpd installers against local and trunk.

Updated the sshnp and sshnpd installers in order to get it running on the debian slim test image, had to cut back on certain shell features which weren't available in the slim environment.

On the sshnp side, it does not test the custom script, I will think about doing a later update once the newer script template is released, since the newer script template allows passing custom args to sshnp.

On the sshnpd side, it tests the custom script, but doesn't test the job used to install the script (i.e. with cron/tmux/etc.).

This is a start, there is room for improvement but it is good enough for now.

**- How I did it**

**- How to verify it**

ci: e2e tests for the installer

### DIFF
--- a/.github/composite/setup_entrypoints/action.yaml
+++ b/.github/composite/setup_entrypoints/action.yaml
@@ -18,6 +18,12 @@ inputs:
   wait_time:
     description: Time to wait for sshnpd to start
     required: false
+  sshnp:
+    description: Source of the sshnp container
+    required: true
+  sshnpd:
+    description: Source of the sshnpd container
+    required: true
 
 runs:
   using: composite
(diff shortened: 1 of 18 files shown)


xavierchanth:
It seems that the slim environment is unhappy with some of the shell syntax used in the install scripts, might take a bit longer than expected to get them working

xavierchanth:
I remember going with some of these solutions before, and shell check recommending a different approach... where it would prefer the cleaner solution which works 90% of the time instead of the dirty 99% solution.

xavierchanth:
The installer - installer test seems to not be running anything, @JeremyTubongbanua maybe you have some insight into what I did wrong?

JeremyTubongbanua:
My only feedback would be to just create a whole new template like "sshnp_installer_entrypoint.sh" then when we copy it over into the context, make sure it's named "entrypoint.sh" 

Like
```sh
cp /../sshnp_installer_entrypoint.sh /../entrypoint.sh
```

JeremyTubongbanua (dismissed):

xavierchanth:
> My only feedback would be to just create a whole new template like "sshnp_installer_entrypoint.sh" then when we copy it over into the context, make sure it's named "entrypoint.sh" 
> 
> 
> 
> Like
> 
> ```sh
> 
> cp /../sshnp_installer_entrypoint.sh /../entrypoint.sh
> 
> ```

Could do, but most of the logic is the same, so I introduced the USE_INSTALLER environment variable to deal with the few differences.

JeremyTubongbanua:
> > My only feedback would be to just create a whole new template like "sshnp_installer_entrypoint.sh" then when we copy it over into the context, make sure it's named "entrypoint.sh"
> > Like
> > ```shell
> > cp /../sshnp_installer_entrypoint.sh /../entrypoint.sh
> > ```
> 
> Could do, but most of the logic is the same, so I introduced the USE_INSTALLER environment variable to deal with the few differences.

Cause when we add more tests, it'll fill up the sshnpX_entrypoint.sh

xavierchanth:
> > > My only feedback would be to just create a whole new template like "sshnp_installer_entrypoint.sh" then when we copy it over into the context, make sure it's named "entrypoint.sh"
> 
> > > Like
> 
> > > ```shell
> 
> > > cp /../sshnp_installer_entrypoint.sh /../entrypoint.sh
> 
> > > ```
> 
> > 
> 
> > Could do, but most of the logic is the same, so I introduced the USE_INSTALLER environment variable to deal with the few differences.
> 
> 
> 
> Cause when we add more tests, it'll fill up the sshnpX_entrypoint.sh

Ah, makes sense

xavierchanth:
@JeremyTubongbanua I'll need a reapproval, feel free to merge if you are happy with the changes.

xavierchanth:
I made it so that the composite action controls the name of the entrypoint template file to use

`.github/composite/setup_entrypoints/action.yaml`

xavierchanth:
Please squash when you merge this PR!

JeremyTubongbanua (approved):

PR #344, opened 2023-08-15
feat: Sshrv using pure dart

xavierchanth:
**- What I did**

Current way that the sshnp uses sshrv is incompatible with @CurtlyCritchlow 's needs for the Flutter app. So we need a pure dart way of calling sshrv (already done in this PR).

However, this raises the question though, why even bother having a separate binary for sshrv?
It would be much more stable knowing that the sshrv code is a part of sshnp / sshnpd. 

[e1cead8](https://github.com/atsign-foundation/sshnoports/pull/344/commits/e1cead841ee794a27d0ae7dd0c495e5a78af5b55) contains the code which still uses the local binaries but allows for fallback to pure dart method.
[Full diff for those changes](https://github.com/atsign-foundation/sshnoports/compare/5400bd80cac30e7935fe4bbc72d0410579e5a605...e1cead841ee794a27d0ae7dd0c495e5a78af5b55). 

**- How I did it**

**- How to verify it**


gkc:
sshrv has to run in a separate process because that allows sshnp itself to then exit without affecting the rv connection 

If there's a pure dart way to fork the process that would work 

Alternatively we could have sshnp have two modes - the command-line mode as is, and the app mode where it runs the sshrv code within the same process

xavierchanth:
> sshrv has to run in a separate process because that allows sshnp itself to then exit without affecting the rv connection
> 
> If there's a pure dart way to fork the process that would work
> 
> Alternatively we could have sshnp have two modes - the command-line mode as is, and the app mode where it runs the sshrv code within the same process

AFAIK we can't run internal dart code in a separate process without having a separate binary.
What I've done instead is added `sshrvGenerator` which accepts either `SSHRV.pureDart` or `SSHRV.localBinary`. It will default to `SSHRV.localBinary` to keep the same behaviour in the CLI. 

@CurtlyCritchlow what this means for you is when you generate an SSHNP object, you must always set sshrvGenerator to `SSHRV.pureDart`:

```dart
import 'package:sshnoports/sshrv/sshrv.dart';

var sshnp = SSHNP(
  //... other parameters,
  sshrvGenerator: SSHRV.pureDart,
  );
```

For now please use this branch @CurtlyCritchlow

xavierchanth:
superceded by #367

PR #354, opened 2023-08-17
feat: add warning message when using default installation args for sshnpd

xavierchanth:
**- What I did**

Added a warning message when installing sshnpd and --args was unset

**- How I did it**

**- How to verify it**
Running the added echos in the terminal:
```
~ 
➜ echo "Warning: For security reasons, as of v3.5.0 the installer no longer sets -u and -s";
    echo "  -s when set, will update authorized_keys to include public key from sshnp";
    echo "  -u when set, makes various information visible to the manager atSign - e.g. username, version, etc";
    echo "  -v when set, enables more logging (on by default)";
    echo "To pass additional arguments to sshnpd at install time, use --args";
    echo "  e.g. <rest of install command> --args \"-s -u -v\"";
Warning: For security reasons, as of v3.5.0 the installer no longer sets -u and -s
  -s when set, will update authorized_keys to include public key from sshnp
  -u when set, makes various information visible to the manager atSign - e.g. username, version, etc
  -v when set, enables more logging (on by default)
To pass additional arguments to sshnpd at install time, use --args
  e.g. <rest of install command> --args "-s -u -v"
```

feat: add warning message when using default installation args for sshnpd

### DIFF
--- a/packages/sshnoports/scripts/install_sshnpd
+++ b/packages/sshnoports/scripts/install_sshnpd
@@ -266,6 +266,12 @@ parse_env() {
   fi
 
   if [ -z "$SSHNP_SERVICE_ARGS_PARSED" ]; then
+    echo "Warning: For security reasons, as of v3.5.0 the installer no longer sets -u and -s"
+    echo "  -s when set, will update authorized_keys to include public key from sshnp"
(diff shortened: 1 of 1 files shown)


xavierchanth:
This needs to be reopened on the new script location

PR #367, opened 2023-08-18
feat: Add support for GUI sshnp client

xavierchanth:
**- What I did**

Arguments:
- Fixed a bunch of merge conflicts related to adding multiple args to SSHNP
  - Added trailing commas to parameter lists to reduce the chances of this kind of merge conflict from reoccurring
- Made legacyDaemon arg optional and provided a default
- Added static const default* arguments to SSHNP class
- Made all port args int type instead of String type

SSHRV:
- Added an abstract class with localBinary and pureDart factory functions
- localBinary runs in a separate process so that CLIs can exit
- pureDart is needed so that the GUI is not dependant on separate binaries elsewhere

Misc:
- Now using part/part of notation for related files
- Added SSHNPResult class, which overrides toString to output the ssh command for CLI, but the class can be intercepted for the GUI

**- How I did it**

**- How to verify it**

feat: Add support for GUI sshnp client


xavierchanth:
Need to merge #366 to fix the rest of the tests

packages/sshnoports/bin/sshnp.dart
```diff
@@ -55,7 +54,10 @@ void main(List<String> args) async {
     }
 
     await sshnp.init();
-    await sshnp.run();
+    SSHNPResult res = await sshnp.run();
+    if (res is SSHCommand) {
+      stdout.write('$res\n');
+    }
     exit(0);
```
gkc, line 64: Don't we need a test for `res is SSHNPFailed` which should (I presume) result in an`exit(1);`?

gkc (changes requested):

packages/sshnoports/bin/sshnp.dart
xavierchanth: Good catch, thanks

gkc (approved):

PR #370, opened 2023-08-20
feat: Add sshnp option for local sshd to run on alternative port than 22

cconstab:
**- What I did**

Added optional parameter to sshnp (-P / --local-sshd-port)

**- How I did it**
Followed existing patterns to add optional parameter that defaults to 22 
Added 3rd (optional) parameter to sshrv to forward port X to sshrvd

**- How to verify it**
Ran tests locally (I need help to make an automated test please! BEFORE Merging)

Added sshnp optional parameter to select local sshd on alternative port (useful on Android/ioS devices that are not rooted)

### DIFF
--- a/packages/sshnoports/lib/sshnp/sshnp.dart
+++ b/packages/sshnoports/lib/sshnp/sshnp.dart
@@ -63,6 +63,11 @@ abstract class SSHNP {
   /// is set to '0' then
   abstract String localPort;
 
+  /// Port that local sshd is listening on localhost interface
+  /// Default set to 22
+
+  abstract String localSshdPort;
+
   // ====================================================================
   // Derived final instance variables, set during construction or init
   // ====================================================================
@@ -137,6 +142,7 @@ abstract class SSHNP {
       required String localPort,
       String? remoteUsername,
       bool verbose = false,
+      required String localSshdPort,
       required bool legacyDaemon}) {
     return SSHNPImpl(
         atClient: atClient,
@@ -151,6 +157,7 @@ abstract class SSHNP {
         host: host,
         port: port,
         localPort: localPort,
+        localSshdPort: localSshdPort,
         remoteUsername: remoteUsername,
         verbose: verbose,
         legacyDaemon: legacyDaemon);
--- a/packages/sshnoports/lib/sshnp/sshnp.dart
+++ b/packages/sshnoports/lib/sshnp/sshnp.dart
@@ -142,7 +142,7 @@ abstract class SSHNP {
       required String localPort,
       String? remoteUsername,
       bool verbose = false,
-      required String localSshdPort,
+      String localSshdPort = '22',
       required bool legacyDaemon}) {
     return SSHNPImpl(
         atClient: atClient,
--- a/packages/sshnoports/lib/sshnp/sshnp_impl.dart
+++ b/packages/sshnoports/lib/sshnp/sshnp_impl.dart
@@ -56,6 +56,9 @@ class SSHNPImpl implements SSHNP {
   @override
   final List<String> localSshOptions;
 
+  @override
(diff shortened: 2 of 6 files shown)


cconstab:
@XavierChanth I am not sure if I got the optional parameter code correct to pull in from config file could you take a very close look for me ?

@JeremyTubongbanua Would love help to create a test for this option please!

xavierchanth (approved):

JeremyTubongbanua (dismissed):

packages/sshnoports/lib/sshnp/sshnp_impl.dart
```diff
@@ -179,7 +182,8 @@ class SSHNPImpl implements SSHNP {
       required this.localPort,
       this.remoteUsername,
       this.verbose = false,
-      required this.legacyDaemon}) {
+      required this.legacyDaemon,
+      required this.localSshdPort}) {
```
xavierchanth: probably best to have this not required, instead `this.localSshdPort = 22`

packages/sshnoports/lib/sshnp/sshnp.dart
```diff
@@ -137,6 +142,7 @@ abstract class SSHNP {
       required String localPort,
       String? remoteUsername,
       bool verbose = false,
+      required String localSshdPort,
```
xavierchanth: Same thing here `this.localSshdPort = 22`

packages/sshnoports/lib/sshnp/sshnp.dart
```diff
@@ -63,6 +63,11 @@ abstract class SSHNP {
   /// is set to '0' then
   abstract String localPort;
 
+  /// Port that local sshd is listening on localhost interface
+  /// Default set to 22
+
+  abstract String localSshdPort;
```
xavierchanth, line 69: Would prefer this to be an int, since int.parse could be a non-numeric String, and that is extra validation that needs to be done by the sshnp class

xavierchanth (changes requested):
Added a few more recommended changes

packages/sshnoports/lib/sshnp/sshnp.dart
cconstab: I set as a String as this is what the Params give but thought the same as I was writing the code... Prefer int too.. Where is the best place the do that conversion ?

packages/sshnoports/lib/sshnp/sshnp.dart
cconstab: Not possible in a constructer

packages/sshnoports/lib/sshnp/sshnp.dart
cconstab: I think this would need a larger refactor as many other params are currently Strings too.. 
e.g. LocalPort etc
Also not sure how/where to do the conversion to int without loosing fucntionality in the conf file stuff as ints cannot be set as Null

packages/sshnoports/lib/sshnp/sshnp.dart
cconstab: used ' String localSshdPort = '22' instead

packages/sshnoports/lib/sshnp/sshnp.dart
cconstab: In the meantime added checks for int's with int.tryParse and port range checks

xavierchanth (approved):

xavierchanth:
FYI #367 is likely to have merge conflicts with this PR, I will address them when/if they come up

xavierchanth:
Feel free to merge this first now @cconstab there should be no issues given what is already on trunk

cconstab:
Thanks for your attention to detail..

PR #386, opened 2023-08-23
ci: e2e tests timeout max 6 minutes

JeremyTubongbanua:
**- What I did**

- docker services will be max up 6 minutes
- some tests will run forever and never fail
**- How I did it**

**- How to verify it**

### DIFF
--- a/.github/workflows/end2end_tests.yaml
+++ b/.github/workflows/end2end_tests.yaml
@@ -16,7 +16,7 @@ on:
 env:
   SSHNP_ATSIGN: "@8incanteater"
   SSHNPD_ATSIGN: "@8052simple"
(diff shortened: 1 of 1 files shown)


cconstab:
Does this actually work ? I thought that --timeout only worked in interactive mode ?

JeremyTubongbanua:
> Does this actually work ? I thought that --timeout only worked in interactive mode ?

Possibly, I will double check

xavierchanth:
6 mins is not enough, the sync based ones need ~10 mins

xavierchanth:
I also have a way to improve e2e test time, docker compose has a dependency and healthcheck feature which we can use to have sshnp wait until sshnpd is ready to receive connections (rather than some arbitrary time frame)

PR #388, opened 2023-08-24
feat: More Installer requirements checks

xavierchanth:
**- What I did**

Still doesn't check that sshd is running, but checks that every program needed is available
- which is hard to do
- I think sshnpd / sshnp should check for that instead (since it is a program requirement, not installer requirement)

**- How I did it**

**- How to verify it**

feat: More Installer requirements checks

### DIFF
--- a/scripts/install_sshnp
+++ b/scripts/install_sshnp
@@ -204,7 +204,7 @@ parse_requirements() {
 
 # Check the script's requirements
 check_requirements() {
-  if [ -z "$REQ_DART" ] && [ -z "$SSHNP_DEV_MODE" ]; then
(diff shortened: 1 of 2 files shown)


gkc (dismissed):

gkc (commented):
The script changes LGTM but the e2e test rigs need to be updated to ensure all required dependencies are present on the docker images

gkc (changes requested):
Why is dart a required dependency?

xavierchanth:
Dart is a required dependency when you build from the repo's source code, it should only check for dart when building from source, but I had it only checking for dart for the opposite. It should be fixed now

gkc (approved):

PR #389, opened 2023-08-24
ci: E2e test improvements

xavierchanth:
**- What I did**

1. Removed wait time for sshnp
  - instead the sshnpd emits a healthcheck (returns healthy once monitor is started)
  - sshnp waits for sshnpd to be healthy before starting
2. Removed some of the unused/hardcoded bits of the test suit
  - Moved the manual-tool from tests/ to tools/
3. Added the ability to provide and specify different atSigns and their keys more easily (needed to run the local environement)
  - The environment automatically expects the secret for a set of .atKeys to be named after it
  - i.e. @XavierChanth has its .atKeys secret stored as ATKEYS_XAVIERCHANTH
4. Added a [nektos/act](https://github.com/nektos/act) wrapper for the e2e tests so that they can be run locally
  - Pass your env vars to .env and secrets (.atKeys) to .secrets
  - Starts up a docker container simulating the Github runner, which in turns connects to the local docker host to run the docker compose test suite
5. Put an explicit check in place to see if "Test Passed" are actually written to logs since sometimes they would pass when they shouldn't be passing
  - Caught a bug in trunk... I spent way too long wondering why my local runs weren't working (I was testing local-local) only to realize it wasn't the tests, but that sshnp is writing a forward ssh command not a loopback ssh command to stdout... 
6. Changed the way the tests mount the entrypoints and keys folder (uses a /mount before) since the way that nektos/act starts the docker environment would cause a "mount-inception" with the current setup. So I had to separate the mounts into a host mount, and a docker build mount, then copy files from the host mount to where they need to be. (Yay debugging this issue)

**- How I did it**

**- How to verify it**

ci: E2e test improvements


xavierchanth:
These tests that are failing are due to the fact that sshnp is not defaulting to legacyDaemon=true on trunk

xavierchanth:
Sorry about some of these diffs, most are changing file locations + updating file paths

cpswan (approved):

PR #406, opened 2023-08-31
ci: Split e2e tests

xavierchanth:
**- What I did**

Major change:
- Separated the e2e tests into 3 groups, and simplified them where possible.
  - Group 1 - local against trunk - runs pretty much always
  - Group 2 - local against installer - runs pretty much always, since the installer's prod is trunk branch
  - Group 3 - local against old-releases - runs only on a manual workflow call

These changes were done to reduce the number of times we trigger backwards tests, but also to simplify things so that @JeremyTubongbanua can implement tests against prod rvds more easily. (He will implement them similar to how Group 3 is implemented)

In addition, I made the following minor changes requested by @cpswan :
- [x] README for act wrapper
- [x] Filename more explicit (i.e. act run)
- [x] Note where some of the ../ paths are pointing to
- [x] Comment why shellchecks are disabled

**- How I did it**

**- How to verify it**

ci: Split e2e tests


xavierchanth:
As you can see, when the PR is created the backward (release) tests are skipped. Before reviewing this, I'd like to manually run against this branch to confirm that those tests are working as intended. I will link that workflow run to this PR.

xavierchanth:
I broke the atKey loading from secrets, by creating and trying to use the env var in the same step... confirming locally before I push the fix

xavierchanth:
Here are the additional backward tests which I've triggered: https://github.com/atsign-foundation/sshnoports/actions/runs/6034676624

cpswan (approved):

PR #410, opened 2023-08-31
fix: Fixed runtime error if running sshnp with invalid arguments (e.g. no arguments)

gkc:
**- What I did**
- Fixed runtime error if running sshnp with invalid arguments (e.g. no arguments)

**- How I did it**
- fix: sshnp_params.dart : because ArgParser.addOption's `defaultsToParameter` is a String?, the default value must also be presented to the parser as a String (or null, if there is no default value)
- feat: added a `toString()` to SSHNPArg

**- How to verify it**
- `dart bin/sshnp.dart` should successfully print usage information and exit


cconstab (approved):
tested and works just fine TY

PR #413, opened 2023-09-01
ci: Verify tags before building

### LINKED ISSUE #412: Check version.dart and pubspec.yaml for the correct version when building from a tag
A common mistake is building a new release but with the wrong version # in version.dart, so we should put in a check to fail the build if version.dart is wrong

xavierchanth:
closes #412 
**- What I did**

Guard against a mismatch between version.dart and the GitHub tag for:

- dockerhub sshnpd
https://github.com/atsign-foundation/sshnoports/actions/runs/6045156860

- multibuild
https://github.com/atsign-foundation/sshnoports/actions/runs/6045158188

Note that I used a different git tag than the version in version.dart on purpose to ensure that these workflows fail when they are supposed to.

CleanShot 2023-09-01 at 11 45 51@2x

**- How I did it**

**- How to verify it**

ci: Verify tags before building

### DIFF
--- a/.github/workflows/dockerhub_sshnpd.yml
+++ b/.github/workflows/dockerhub_sshnpd.yml
@@ -22,6 +22,20 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
+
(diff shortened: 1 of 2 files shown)


cconstab:
Grest check ty

cconstab (approved):

PR #426, opened 2023-09-05
refactor: fix volume paths

JeremyTubongbanua:
**- What I did**

- fixed volume paths after the refactor that was done when moving the manual tool from end-to-end tests to root `tools/`
- reordered so sshnp container service is at the top

**- How I did it**

**- How to verify it**


xavierchanth (approved):
Sorry for breaking these!

PR #439, opened 2023-09-08
fix: send ssh public key

xavierchanth:
**- What I did**

- Added the namespace to the sshpublickey, since the daemon wasn't receiving it
- Switched to using `Platform.pathSeperator` for generating the authorized_keys file path

**- How I did it**

**- How to verify it**

fix: send ssh public key

### DIFF
--- a/packages/sshnoports/lib/sshnp/sshnp_impl.dart
+++ b/packages/sshnoports/lib/sshnp/sshnp_impl.dart
@@ -638,6 +638,7 @@ class SSHNPImpl implements SSHNP {
         ..key = 'sshpublickey'
         ..sharedBy = clientAtSign
         ..sharedWith = sshnpdAtSign
+        ..namespace = namespace
         ..metadata = (Metadata()
           ..ttr = -1
           ..ttl = 10000);
--- a/packages/sshnoports/lib/sshnpd/sshnpd_impl.dart
+++ b/packages/sshnoports/lib/sshnpd/sshnpd_impl.dart
@@ -295,10 +295,6 @@ class SSHNPDImpl implements SSHNPD {
 
     late final String sshPublicKey;
     try {
-      var sshHomeDirectory = '$homeDirectory/.ssh/';
-      if (Platform.isWindows) {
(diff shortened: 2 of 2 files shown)


packages/sshnoports/lib/sshnpd/sshnpd_impl.dart
```diff
@@ -309,7 +305,9 @@ class SSHNPDImpl implements SSHNPD {
       }
 
       // Check to see if the ssh Publickey is already in the file if not append to the ~/.ssh/authorized_keys file
-      var authKeys = File('${sshHomeDirectory}authorized_keys');
+      var authKeysFilePath = [homeDirectory, '.ssh', 'authorized_keys']
+          .join(Platform.pathSeparator);
+      var authKeys = File(authKeysFilePath);
```
gkc, line 310: While you're in this code could you add the check to see if it is already in the file? Currently it duplicates entries.

xavierchanth, line 310: It does check, lines 316/314 - 318/316 (just after this diff)

packages/sshnoports/lib/sshnp/sshnp_impl.dart
```diff
@@ -638,6 +638,7 @@ class SSHNPImpl implements SSHNP {
         ..key = 'sshpublickey'
         ..sharedBy = clientAtSign
         ..sharedWith = sshnpdAtSign
+        ..namespace = namespace
```
gkc, line 641: In 4.0.0-rc2 I see this in daemon logs
```
INFO|2023-09-08 14:05:31.280427| sshnpd |ssh Public Key received from @garycasey notification id : d5e36215-aee7-4682-8a5a-c77b8e83e6f2 
```
and the key is successfully added to the authorized_keys file

so I don't believe this is currently a bug, because the namespace is picked up from the AtClient's namespace.

However it is indeed better to be explicit and set the namespace

gkc (approved):

PR #443, opened 2023-09-10
feat: sshnp pure dart for direct ssh

gkc:
**- What I did**
- support the pure-dart SSHClient for forward ssh
- use ephemeral ssh keys for forward ssh tunnel. these are generated on the daemon side which obviates the requirement to have ssh-keygen available on the client side when using forward ssh (which will be the default behaviour once sshnoports v4.0.0 is released)
- configurable `--idle-timeout` for tunnel connections
- optionally apply local port forwarding from localSshOptions to the initial (tunnel) forward ssh connection (new `--add-forwards-to-tunnel` flag in sshnp)
- configurable `--ephemeral-permissions` option on daemon to control what hosts and ports can be forwarded to on the forward tunnel connections (because the forward ssh tunnel connections use ephemeral keys and thus are constrained to be able to literally nothing other than connect, by default)
- added `--local-sshd-port` option to sshnpd for installations which are using non-standard sshd ports, and accordingly also added `--remote-sshd-port` option to sshnp
- fixed a couple of bugs which had crept into trunk for forward ssh. Next we need to [add to the e2e test pack](#446) in order to prevent any more regressions in the future.

**- How I did it**
See messages on the individual commits in the PR

**- How to verify it**
- [x] Existing e2e tests pass
- [x] Manual tests of forward ssh pass

### DIFF
--- a/packages/sshnoports/lib/sshnp/sshnp.dart
+++ b/packages/sshnoports/lib/sshnp/sshnp.dart
@@ -7,6 +7,7 @@ import 'package:at_client/at_client.dart' hide StringBuffer;
 import 'package:at_commons/at_builders.dart';
 import 'package:at_utils/at_logger.dart';
 import 'package:at_utils/at_utils.dart';
+import 'package:dartssh2/dartssh2.dart';
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
 import 'package:path/path.dart' as path;
@@ -20,6 +21,8 @@ import 'package:sshnoports/sshrv/sshrv.dart';
 import 'package:sshnoports/sshrvd/sshrvd.dart';
 import 'package:sshnoports/version.dart';
 import 'package:uuid/uuid.dart';
+
+import '../common/defaults.dart' as defaults;
 
 part 'sshnp_impl.dart';
 part 'sshnp_params.dart';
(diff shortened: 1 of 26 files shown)


packages/sshnoports/lib/sshnp/sshnp.dart
```diff
@@ -146,19 +152,27 @@ abstract class SSHNP {
...
   static const defaultSendSshPublicKey = '';
   static const defaultLocalSshOptions = <String>[];
-  static const defaultVerbose = false;
-  static const defaultRsa = false;
-  static const defaultRootDomain = 'root.atsign.org';
-  static const defaultSshrvGenerator = SSHRV.localBinary;
-  static const defaultLocalSshdPort = 22;
   static const defaultLegacyDaemon = true;
   static const defaultListDevices = false;
+  static const defaultSshClient = SupportedSshClient.hostSsh;
```
xavierchanth, line 179: Wouldn't it be best to move the rest of these to `common/defaults.dart`

That way we keep all of the defaults in one place for those who are less familiar with the code base

xavierchanth (dismissed):

packages/sshnoports/lib/sshnp/sshnp.dart
xavierchanth: Not a blocker, but can be done in a later PR

gkc: I didn't move these few as I felt like they are specific to SSHNP - legacyDaemon and listDevices for sure, and we may in future have different default sshClients in daemon / client

xavierchanth: > I didn't move these few as I felt like they are specific to SSHNP - legacyDaemon and listDevices for sure, and we may in future have different default sshClients in daemon / client

That makes sense

xavierchanth (approved):

PR #448, opened 2023-09-11
ci: alternate_port_test improvements and remove testing for v3.2.0

JeremyTubongbanua:
#449 should be merged first

**- What I did**

- alternate_port_test would not run on `workflow_dispatch`, this pr fixes that
- alternate_port_test builds the docker-compose.yaml instead
- alternate_port_test now uses the default sshnpd_entrypoint.sh, while using its own unique sshnp_entrypoint.sh
- removed backwards compatibility test v3.2.0 because our current health check does not work on it

**- How I did it**

**- How to verify it**

- alternate port test passing (see latest CI tests in this PR)
- running end2end_tests through a workflow_dispatch will now pass, since the v3.2.0 test that was always failing would be no longer there. Proof: https://github.com/atsign-foundation/sshnoports/actions/runs/6167204444


JeremyTubongbanua:
This PR should wait for #449

xavierchanth (approved):

PR #474, opened 2023-09-14
ci: prod tests fail-fast

JeremyTubongbanua:
**- What I did**

- prod tests has strategy - fail-fast: true

**- How I did it**

**- How to verify it**

### DIFF
--- a/.github/workflows/prod_tests.yaml
+++ b/.github/workflows/prod_tests.yaml
@@ -39,7 +39,7 @@ jobs:
         run: |
           echo "rvd: ${{ matrix.rvd }}"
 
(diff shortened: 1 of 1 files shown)


cpswan:
@JeremyTubongbanua it's obvious **what** you're doing here, and I can guess **why**, but maybe you can give your version?

JeremyTubongbanua:
> @JeremyTubongbanua it's obvious **what** you're doing here, and I can guess **why**, but maybe you can give your version?

I wish that other jobs do not get cancelled when one of the jobs fail. For example, if @rv_am fails, I would still want the @rv_eu and @rv_ap tests to run.

cpswan (approved):

PR #502, opened 2023-09-26
refactor: add noports_core package

xavierchanth:
**- What I did**

- split sshnoports into the current package and noports_core which will contain the decoupled core code.
- Removed cli dependent code from it by doing the following:
  - No longer generating atClient in the library instead accepting either an AtClient or AtClient generator function
  - Added an option usageCallback function which gets called when the usage should be called for the CLI

(Looks like a lot of changes, but it's mostly just import renames)

FYI this package needs a README to be written still, but it is enough to progress with the demo applications that need to be created using the code. The README will be written before publishing the package (since pub.dev won't let you publish without it)

**- How I did it**

**- How to verify it**

refactor: add noports_core package


gkc (approved):
Thank you @XavierChanth

PR #516, opened 2023-10-04
refactor: Remove file system bindings

xavierchanth:
**- What I did**

- Updated key arguments:
  - Added `-i` just like ssh, `-s` is now a bool which sends the public key associated with `-i`
- Renamed `SSHNPImpl` to `SSHNPCore` since it is abstract, it isn't really an implementation
- A major overhaul to local file and ssh key management
  - Added `AtSSHKeyPair` which holds a set of ssh keys
  - Added `LocalSSHKeyUtil` and `DartSSHKeyUtil` to manage keys depending on what is available to store / create SSH Keys
  - Added `SSHNPDLocalSSHKeyHandler` (a mixin on `SSHNPCore`) which exposes `LocalSSHKeyUtil` safely
  - Added `SSHNPDDartSSHKeyHandler` (a mixin on `SSHNPForwardDart`) which exposes `DartSSHKeyUtil` safely
- Split the ForwardDart implementation into two:
  - one which expects that the local file system is still available, so it can output an ssh command
  - another which expects that there is no local file system available and just passes a handler to the dartssh2 `SSHClient`
  - The reason for this design decision is that I expect we will want to run the first in a separate isolate so that we can exit the main program safely, but then we can no longer access the `SSHClient` so it is not suitable for use in sshnp_gui
  - I have not implemented these design decisions as I am leaving them open ended for an architecture discussion
- Deferred payload handling from sshnpd to mixins, since it is implementation specific
  - The default payload handler mixin exposes the ephemeralPrivateKey safely
- Added type-safety to
  - DefaultArgs
  - String names in SSHNPArg/SSHNPParams
- Replaced the rsa boolean with an enum of sshAlgorithms
  - currently just ed25519 (default) and rsa

Several Action items come out of these changes:
- Write new unit tests for arg parsing which validate against the new argument changes to:
  - `-i`, `-s`, `-r` 
- Modify the GUI to be able to input the new arguments
  - Need to write UI for handling SSH key management
- Architecture discussion
  - How to handle the ForwardDart Implementations
  - 
**- How to verify it**

**NB - Ignore the failed client-trunk test, it is being provided the `-i` arg and doesn't know how to handle it.**

refactor: Remove file system bindings


packages/noports_core/lib/src/common/default_args.dart
```diff
@@ -1,30 +1,39 @@
-import 'package:noports_core/src/common/supported_ssh_clients.dart';
+import 'dart:io';
```
xavierchanth, line 1: Added explicit types to this file

packages/noports_core/lib/src/common/file_system_utils.dart
```diff
@@ -0,0 +1,51 @@
+import 'dart:io';
```
xavierchanth, line 1: This was utils.dart, but I extracted the minimum required shared file utilities.

packages/noports_core/lib/src/common/ssh_key_utils.dart
```diff
@@ -0,0 +1,217 @@
+import 'dart:async';
```
xavierchanth, line 1: All SSH Key Management is deferred to these classes. The AtSSHKeyPair takes a few extra ms (5ms in my tests) to parse a privateKey, but the benefit is that we have a common interface.

xavierchanth: Public Key encoding typically takes 3ms

packages/noports_core/lib/src/common/supported_ssh_clients.dart
```diff
@@ -1,18 +0,0 @@
-enum SupportedSshClient {
```
xavierchanth, line 1: Moved to types.dart

packages/noports_core/lib/src/common/types.dart
```diff
@@ -0,0 +1,39 @@
...
+      (arg) => arg._cliArg == cliArg,
+      orElse: () => throw ArgumentError('Unsupported SSH client: $cliArg'),
+    );
+  }
+
+  @override
+  String toString() => _cliArg;
+}
+
+enum SupportedSSHAlgorithm {
```
xavierchanth, line 23: Changed the bool rsa to an enum

packages/noports_core/lib/src/sshnp/forward_direction/sshnp_forward_exec_impl.dart
```diff
@@ -29,27 +39,27 @@ class SSHNPForwardExecImpl extends SSHNPImpl with SSHNPForwardDirection {
     }
 
     logger.info(
-        'Starting direct ssh session for ${params.username} to $host on port $sshrvdPort with forwardLocal of $localPort');
+        'Starting direct ssh session to $host on port $sshrvdPort with forwardLocal of $localPort');
 
     try {
       String? errorMessage;
       Process? process;
```
xavierchanth, line 57: Replaced the hardcoded ephemeral key writer with SSH Key Util calls

packages/noports_core/lib/src/sshnp/mixins/sshnp_local_file_handler.dart
```diff
@@ -0,0 +1,25 @@
...
+
+import 'package:noports_core/src/sshnp/sshnp_impl.dart';
+import 'package:noports_core/src/sshnp/sshnp_result.dart';
+import 'package:noports_core/utils.dart';
+
+mixin SSHNPLocalFileHandler on SSHNPImpl {
+  final LocalSSHKeyUtil _sshKeyUtil = LocalSSHKeyUtil();
+  @override
+  LocalSSHKeyUtil get keyUtil => _sshKeyUtil;
+
```
xavierchanth: A mixin on SSHNPImpl which makes the local file system safely available. At init time, it checks whether the platform is valid to run the local filesystem (i.e. mac/linux/windows). Most importantly, it exposes the LocalSSHKeyUtil

packages/noports_core/lib/src/sshnp/mixins/sshnpd_payload_handler.dart
```diff
@@ -0,0 +1,58 @@
...
+import 'package:noports_core/sshnp_impl.dart';
+import 'package:noports_core/utils.dart';
+
+mixin DefaultSSHNPDPayloadHandler on SSHNPImpl {
+  @protected
+  late String ephemeralPrivateKey;
+
+  @override
+  FutureOr<bool> handleSshnpdPayload(AtNotification notification) async {
+    if (notification.value?.startsWith('{') ?? false) {
```
xavierchanth, line 19: Deferred the payload to a mixin, since implementation details were leaking into SSHNPImpl, this is the default (AKA non-legacy) payload handler

packages/noports_core/lib/src/sshnp/mixins/sshnpd_payload_handler.dart
```diff
@@ -0,0 +1,58 @@
...
+    }
+    return false;
+  }
+}
+
+mixin LegacySSHNPDPayloadHandler on SSHNPImpl {
+  @override
+  bool handleSshnpdPayload(AtNotification notification) {
+    return (notification.value == 'connected');
+  }
```
xavierchanth, line 64: And here's the legacy payload handler

packages/noports_core/lib/src/sshnp/reverse_direction/sshnp_reverse_direction.dart
```diff
@@ -0,0 +1,81 @@
...
+import 'package:noports_core/src/sshnp/mixins/sshnp_local_file_handler.dart';
+import 'package:noports_core/sshnp.dart';
+import 'package:noports_core/sshrv.dart';
+import 'package:noports_core/utils.dart';
+
+abstract class SSHNPReverseDirection extends SSHNPImpl
+    with SSHNPLocalFileHandler {
+  SSHNPReverseDirection({
+    required super.atClient,
+    required super.params,
```
xavierchanth: This was a mixin, it is now a proper subclass of SSHNPImpl

packages/noports_core/lib/src/sshnp/sshnp_impl.dart
```diff
@@ -114,6 +110,13 @@ abstract class SSHNPImpl implements SSHNP {
   static String getNamespace(String device) => '$device.sshnp';
   String get namespace => getNamespace(params.device);
 
+  // ====================================================================
+  // Auxiliary
+  // ====================================================================
+
+  @protected
+  AtSSHKeyUtil get keyUtil;
```
xavierchanth: Only the base class for SSHKeyUtil is exposed, with the intention that it is overridden with the specific type needed based on the implementation.

packages/noports_core/lib/src/sshnp/sshnp_impl.dart
```diff
@@ -246,44 +241,7 @@ abstract class SSHNPImpl implements SSHNP {
         .toLowerCase();
     logger.info('Received $notificationKey notification');
 
-    bool connected = false;
-
-    if (notification.value == 'connected') {
-      connected = true;
-    } else if (notification.value?.startsWith('{') ?? false) {
-      late final Map envelope;
```
xavierchanth: This payload code was leaking implementation details, so it was replaced with the abstract handleSshnpdPayload which is implemented by the Payload Handler mixins

packages/noports_core/lib/src/sshnp/sshnp_params/sshnp_arg.dart
```diff
@@ -14,6 +14,21 @@ enum ArgType {
...
+  always,
+  never,
+}
+
+enum ParserType {
+  normal,
+  commandLine,
+  configFile,
+}
+
```
xavierchanth, line 48: Added ParseWhen enum and ParserType to give better (and clearer) control over what the parser should parse / ignore

xavierchanth:
Trunk test is failing because of the breaking change caused by adding the -i option, this makes me wonder if we even need the trunk test, or if it should be disabled by default like the backward tests.

packages/noports_core/lib/src/sshnpd/sshnpd_params.dart
```diff
@@ -1,68 +1,82 @@
...
+  final String atKeysFilePath;
+  final String deviceAtsign;
+  final bool verbose;
+  final bool makeDeviceInfoVisible;
+  final bool addSshPublicKeys;
+  final SupportedSshClient sshClient;
+  final String rootDomain;
+  final int localSshdPort;
+  final String ephemeralPermissions;
+  final SupportedSSHAlgorithm sshAlgorithm;
```
xavierchanth, line 21: Remove late from all of these, which means I had to change how they get set below

packages/noports_core/lib/src/sshrv/sshrv.dart
```diff
@@ -4,8 +4,6 @@ import 'package:noports_core/src/sshrv/sshrv_impl.dart';
 import 'package:socket_connector/socket_connector.dart';
 import 'package:noports_core/src/common/default_args.dart';
 
-typedef SSHRVGenerator = SSHRV Function(String, int, {int localSshdPort});
```
xavierchanth, line 7: Moved this to src/common/types.dart

packages/noports_core/lib/sshnp_core.dart
```diff
@@ -0,0 +1,3 @@
+library noports_core_sshnp_core;
+
+export 'src/sshnp/sshnp_core.dart';
```
xavierchanth, line 3: Exported this so it can implemented in a flutter library later

xavierchanth: If we want to optimize down the road

packages/noports_core/pubspec.yaml
```diff
@@ -11,22 +11,19 @@ environment:
...
 dependencies:
-  archive: ^3.3.8
   args: ^2.4.2
   at_chops: ^1.0.4
   at_client: ^3.0.65
-  at_commons: ^3.0.55
-  at_lookup: ^3.0.40
   at_utils: ^3.0.15
-  crypton: ^2.1.0
+  cryptography: ^2.7.0
```
xavierchanth, line 18: Cryptography is used for generating ed25519 keys in pure dart

packages/noports_core/pubspec.yaml
```diff
@@ -11,22 +11,19 @@ environment:
...
   dartssh2: ^2.8.2
-  logging: ^1.2.0
   meta: ^1.9.1
+  openssh_ed25519: ^1.1.0
```
xavierchanth, line 21: openssh_ed25519 is used for converting the generated ed25519 keys to openssh format

packages/noports_core/pubspec.yaml
```diff
@@ -11,22 +11,19 @@ environment:
...
   path: ^1.8.3
+  posix: ^5.0.0
```
xavierchanth, line 23: posix is used for calling chmod to modify the written ephemeral keys

xavierchanth: We will have to assess if this method works on windows, I have yet to test it

gkc (approved):

PR #539, opened 2023-10-18
refactor: SSHNPParams unit tests preparation

xavierchanth:
**- What I did**

- Started writing unit tests (which some are also included in this PR)
- Realized that I needed to replace most of the inheritance with composition of classes so I removed the hierarchy and instead:
  - Split most of the logic in the SSHNP implementations to mixins and classes:
    - SshrvdChannel - responsible for all communication to sshrvd
    - SshnpdChannel - responsible for all communication to sshnpd (there are subclasses because the legacy client uses unsigned payloads)
    - SshnpInitialTunnelHandler - a mixin which handles starting the initial ssh tunnel for forward clients
    - SshnpSshKeyHandler - a wrapper around SshKeyUtil which helps to safely use the correct ssh key management approach depending on implementation type (important for CLI vs Flutter implementations)

Other changes:
- Separated lifecycle methods to mixins
  - AsyncInitialization and AsyncDisposal
    - Note: I think AsyncDisposal might not be written to correctly handle lifecycle, but I will change that when/if we need to.
- Renamed any uppercase class name segments such as SSHKeyUtil or SSHNPBlahBlahImpl to SshKeyUtil or SshnpBlahBlahImpl
- Removed the e2e installer tests since they are failing and we are going to be ditching that particular installer in the 4.0 release

**- How I did it**

**- How to verify it**

refactor: SSHNPParams unit tests preparation


packages/noports_core/lib/src/sshnp/sshnp_params/config_key_repository.dart
```diff
@@ -1,29 +1,33 @@
 import 'package:at_client/at_client.dart';
+import 'package:meta/meta.dart';
 import 'package:noports_core/src/common/default_args.dart';
 import 'package:noports_core/src/sshnp/sshnp_params/sshnp_params.dart';
 
 class ConfigKeyRepository {
-  static const String _keyPrefix = 'profile_';
-  static const String _configNamespace = 'profiles.${DefaultArgs.namespace}';
+  @visibleForTesting
```
xavierchanth, line 7: Made visible for testing in the unit tests

packages/noports_core/lib/src/sshnp/sshnp_params/sshnp_arg.dart
```diff
@@ -325,7 +325,7 @@ class SSHNPArg {
     allowed: SupportedSshClient.values.map((c) => c.toString()).toList(),
     parseWhen: ParseWhen.commandLine,
   );
-  static final ssHAlgorithmArg = SSHNPArg(
+  static final sshAlgorithmArg = SSHNPArg(
```
xavierchanth: Renamed ssH -> ssh

packages/noports_core/test/sshnp_params/sshnp_core_test.dart
```diff
@@ -0,0 +1 @@
+void main() {}
```
xavierchanth: These are to be written in a later PR

.github/workflows/unit_tests.yaml
```diff
@@ -19,3 +19,17 @@ jobs:
...
+      - uses: dart-lang/setup-dart@b64355ae6ca0b5d484f0106a033dd1388965d06d # v1.6.0
+      - name: dart pub get
+        working-directory: packages/noports_core
+        run: dart pub get
+      - name: dart analyze
+        working-directory: packages/noports_core
+        run: dart analyze
+      - name: dart test
+        working-directory: packages/noports_core
+        run: dart test
```
xavierchanth, line 35: setup dart, dart analyze, dart test noports_core

xavierchanth:
I just have to call it out... I was prepared to rip my hair out over a billion merge conflicts with [4ca6441](https://github.com/atsign-foundation/sshnoports/pull/539/commits/4ca644124ffdffabb7645106ca7aaf77aabee607). The git renames have saved this PR!

.github/workflows/end2end_tests.yaml
```diff
@@ -192,172 +192,6 @@ jobs:
           docker compose down
 
   # Test suite 2
-  # Installer tests (local vs installer)
-  e2e_installer_test:
-    # Don't run on forks (cause no secrets), don't run if dependebot (cause no secrets)
```
xavierchanth, line 197: Removing these tests, since we don't need them for the next release

xavierchanth:
@gkc the trunk tests are not building (as expected) but local-local is now building, so I think it is safe for us to review.

packages/noports_core/lib/src/sshnp/impl/sshnp_dart_pure_impl.dart
```diff
@@ -0,0 +1,101 @@
+import 'dart:async';
+
+import 'package:at_client/at_client.dart';
+import 'package:dartssh2/dartssh2.dart';
+import 'package:noports_core/sshnp_foundation.dart';
+
+class SshnpDartPureImpl extends SshnpCore
+    with SshnpDartSshKeyHandler, SshnpDartInitialTunnelHandler {
```
gkc, line 8: Lots of code duplication between this and sshnp_dart_local_impl.dart - can it be refactored?

packages/noports_core/lib/src/sshnp/impl/sshnp_exec_local_impl.dart
```diff
@@ -0,0 +1,102 @@
+import 'dart:async';
+import 'dart:io';
+
+import 'package:at_client/at_client.dart';
+import 'package:noports_core/sshnp_foundation.dart';
+
+class SshnpExecLocalImpl extends SshnpCore
+    with SshnpLocalSshKeyHandler, SshnpExecInitialTunnelHandler {
```
gkc, line 8: same refactoring question as above

gkc (approved):
Looks good @XavierChanth ... I think there's an opportunity to refactor some duplicated code in the various SshnpCore subclasses but that can come in a follow-up PR

packages/noports_core/lib/src/sshnp/impl/sshnp_dart_pure_impl.dart
xavierchanth: So I thought the same thing at first, but I decided not to (at least for now). In the current abstraction we defer to the run() function to be an explicit orchestration definition. As I went to try and reduce repetitiveness here, I realized that it might be better to be explicit. For one, this way it's a little more clear what is going on under the hood. and I think these explicit steps here are very interchangeable when it comes to differing implementation needs.

gkc: 👍

PR #546, opened 2023-10-24
fix: Uptake notify ephemeral changes

sitaram-kalluri:
**- What I did**
- Uptake the notify ephemeral changes which enables having the value in the notification without having to cache in the keyStore on the recipient's atServer
- Removed code which sets "TTR" to -1 in the metadata
  - Except in SSHNPImpl for the `username.$device` and `device_info.$device` keys which need to be cacheable


gkc (approved):
LGTM

@XavierChanth I am presuming the change to pubspec.lock for noports_core is correct based on the pubspec_overrides.yaml (the change came from when I ran `dart run melos bootstrap` after checking out this branch) ... let me know if it is not

PR #565, opened 2023-11-13
test: Mock tests for SshnpCore

xavierchanth:
**- What I did**

- Stubbed and mocked the constructor + initialization of SshnpCore

**- How I did it**

**- How to verify it**

test: Mock tests for SshnpCore


packages/noports_core/test/sshnp/sshnp_core_constants.dart
```diff
@@ -0,0 +1,14 @@
+class TestingKeyPair {
```
gkc, line 1: Why not just generate at test time?

gkc (dismissed):
Generally I prefer to generate keypairs etc in test code rather than hard-coding but otherwise LGTM, nice work

xavierchanth:
Can I get another review @gkc? Had to address a merge conflict.

packages/noports_core/test/sshnp/sshnp_core_constants.dart
xavierchanth: I can open another PR which generates the temp keys later on, it was quicker for me to hard code.

gkc (approved):

PR #568, opened 2023-11-15
refactor: wrap all dart:io calls

xavierchanth:
**- What I did**

Wrapped all io calls:
- Processes have typedefs which can be replaced with a mocked function
- File system calls are now using the file package which allows for the use of LocalFileSystem by default, but a MemoryFileSystem (i.e. an in memory file system) for testing purposes
- Platform is exported from the io_types file to ensure that it's easy to catch any leakage of dart:io into the rest of the library

Note: I left the dart:io call in `SshnpDartInitialTunnelHandler` since that code needs some discussion before it can be properly addressed. (I'd like to remove all io calls from that tunnel handler altogether, which means changing some of the overall functionality)

- Moved findLocalPortIfRequired from SshnpCore to SshnpOpensshLocalImpl since it depends on SecureSockets, if we find a need to expose this to other implementations later, we can move it to a mixin or something later.

**- How I did it**

**- How to verify it**

refactor: wrap all dart:io calls

### DIFF
--- a/packages/noports_core/lib/src/common/at_ssh_key_util/local_ssh_key_util.dart
+++ b/packages/noports_core/lib/src/common/at_ssh_key_util/local_ssh_key_util.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
-import 'dart:io';
 
+import 'package:meta/meta.dart';
+import 'package:noports_core/src/common/io_types.dart';
 import 'package:noports_core/sshnp.dart';
 import 'package:noports_core/utils.dart';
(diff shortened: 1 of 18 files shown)


gkc (approved):
LGTM; please can you also update the PR description with some info about the refactoring of findLocalPortIfRequired

xavierchanth:
> LGTM; please can you also update the PR description with some info about the refactoring of findLocalPortIfRequired

Forgot I moved that, good catch

PR #602, opened 2023-12-02
feat: New bundles to replace templates in releases

xavierchanth:
**- What I did**

- Added a new bundles folder which splits things up into 3 groups:
  - core (for all releases)
  - shell (mac / linux releases)
  - windows (windows releases)
- Updated the build scripts to use these new bundle paths
- Wrote a new install.sh script in the shell folder which will replace the old installers
  - This script is much smaller and easier to maintain, in exchange it forces the user to configure somethings on their own.
  - I've done this on purpose, because the user should know about how their service was configured, and what the installer did for them.
  - The new script should now be fully POSIX sh compliant
- Move some items in the templates folder to the tools folder (they don't really belong in the release, since they only make sense when you have cloned the full repo)

**- How I did it**

**- How to verify it**

feat: New bundles to replace templates in releases


packages/sshnoports/bundles/shell/headless/sshnpd.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+sleep 10; # allow machine to bring up network
+while true; do
+  # disable "var is referenced but not assigned" warning for template
+  # shellcheck disable=SC2154
+  "$HOME"/.local/bin/sshnpd -a "$device_atsign" -m "$manager_atsign" -d "$device_name" -v
+  sleep 10
+done
```
xavierchanth, line 8: Simplified version of the old headless script we were using, this one intentionally requires manually configuration so that users actually know and learn how to configure their own instances of sshnpd

packages/sshnoports/bundles/shell/headless/sshrvd.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+sleep 10; # allow machine to bring up network
+while true; do
+  # disable "var is referenced but not assigned" warning for template
+  # shellcheck disable=SC2154
+  "$HOME"/.local/bin/sshrvd -a "$atsign" -i "$internet_address"
+  sleep 10
```
xavierchanth, line 7: Same story but for sshrvd

packages/sshnoports/bundles/shell/install.sh
```diff
@@ -0,0 +1,296 @@
+#!/bin/sh
+
```
xavierchanth, line 2: New POSIX sh compliant install script, should also work on ash!

- I make some assumptions, remove some smoke and mirrors, which has resulted in 4x less shell script to maintain, and the shell script is also much less complicated as well

packages/sshnoports/bundles/shell/install.sh
```diff
@@ -0,0 +1,296 @@
...
+  unit_name="$1"
+  no_mac
+  dest="$systemd_dir/$unit_name"
+  cp "$script_dir/systemd/$unit_name" "$dest"
+  post_systemd_message
+}
+
+install_systemd_sshnpd() {
+  root_only
+  install_single_binary "sshnpd"
```
xavierchanth, line 127: while systemd units can be installed as a --user unit, it adds a lot of complexity not worth trying to automate. hence `root_only`

packages/sshnoports/bundles/shell/install.sh
```diff
@@ -0,0 +1,296 @@
...
+  echo "Headless job installed, make sure to configure the job by editing $dest"
+  echo "Learn more in $script_dir/headless/README.md"
+  echo ""
+  echo "The job will start upon next system boot"
+  echo "To start the job immediately:"
+  echo "  $command &"
+  echo ""
+  echo "Warning: logs stored at $log_file and $err_file have unlimited potential size"
+  echo "In a production environment, it's recommended that you install via systemd"
+  echo "or use a cron job to call logrotate on these log files if systemd is not available"
```
xavierchanth, line 162: logrotate isn't available on every system, so I'm not trying to automate it or have to call it... this option should be a last resort anyways.

packages/sshnoports/bundles/shell/install.sh
```diff
@@ -0,0 +1,296 @@
...
+  # Provide initial startup command
+  echo "tmux service installed, make sure to configure the service by editing $dest"
+  echo "Learn more in $script_dir/headless/README.md"
+  echo ""
+  echo "To start the service immediately:"
+  echo "  $command &"
+  echo ""
+  echo "Warning: sometimes tmux is not set to linger which will kill the tmux session on logout"
+  echo "Make sure your system is configured so that tmux sessions can linger, or disown the tmux process before logging out"
+}
```
xavierchanth, line 225: There are ways to automate this, but each solution has trade-offs, leave it to the user to pick and choose.

packages/sshnoports/bundles/windows/README.md
```diff
@@ -0,0 +1,7 @@
+# SSH No Ports Suite
+
+## Installation
+
+At this time, windows doesn't have an automated installer. To use the sshnp.exe
+client, copy the binary to a convenient location for you. You may also add the
+binary to a folder in the PATH so that it can be accessed from anywhere.
```
xavierchanth, line 7: I plan to update this in a further PR, but just so we have something for now.

packages/sshnoports/tools/Dockerfile
```diff
@@ -1,3 +1,6 @@
+# Dockerfile
+# Build image for a containerized instance of sshnpd
+
```
xavierchanth, line 3: Labelling all of our dockerfiles so we know what they do and why they need to exist

scripts/install_sshnp
```diff
@@ -409,6 +409,19 @@ do_update() {
   echo; echo "Update complete!";
 }
 
+deprecation_msg() {
+  case "$SSHNP_VERSION" in
+    # TODO remove latest to signify deprecation on the launch of v4.0.0
+    latest|1.*|2.*|3.*|4.0.0-test.*|4.0.0-rc.*)
+      # no message, this group is expected to use this script
```
xavierchanth, line 416: Added an automated deprecation message which will automatically trigger on 4.0.0 and later, we just have to remove latest from the safe-list once we release 4.0.0 (this is the old script)

cpswan (approved):

PR #604, opened 2023-12-03
feat: Add support for pure dart interactive shell

gkc:
**- What we did**
feat: Add support for pure dart interactive shell

**- How I did it**
Feature
- added startUserSession method to InitialTunnelHandler
- renamed *InitialTunnelHandler to *SshSessionHandler
- added userKeyPairIdentifier to SshnpCore constructor
- substantial refactoring of SshnpDartSshSessionHandler to enable code reuse when implementing startUserSession
- added `bool get canRunShell` and `Future<SshnpRemoteProcess> runShell()` to Sshnp and all subclasses and implemented for the SshnpDartPureImpl
- make sshnoports/bin/sshnp.dart main call `runShell` if `canRunShell` is true, and hook up stdio to the remote shell

Other
- fixed a few typos
- improved logging a bit
- ran dart format

**- How to verify it**
Manually test - e.g.
```
dart bin/sshnp.dart -f @client_at_sign -t @daemon_at_sign -d mbp -h @rv_eu \
  -i ~/.ssh/noports -s ~/.ssh/noports.pub \
  -u gary --ssh-client dart -v 
```


packages/noports_core/lib/src/sshnp/util/sshnp_initial_tunnel_handler/sshnp_dart_initial_tunnel_handler.dart
```diff
@@ -1,56 +1,124 @@
...
+      keyPair: keyPair,
+    );
+
+    // Start local forwarding to the remote sshd
+    logger.info('Starting port forwarding'
+        ' from localhost:$localPort on local side'
+        ' to localhost:${params.remoteSshdPort} on remote side');
+
+    await helper.startForwarding(
+      fLocalPort: localPort,
```
xavierchanth: I think we will have to rethink how we do this later, ideally we can control whether we forward the sshd port to a physical socket, or forward it via Dart Streams to the user session.

The first is preferred on Desktop, but not possible on mobile, so we will need the option of falling back to the second somehow

xavierchanth: I think the current behaviour is fine for the current release though, it shouldn't cause any breaking changes.

gkc: Yes, it's an implementation detail within the PureDartImpl so should be fine to change; my objective for 4.0.0 is to get the pure dart CLI working in such a way that we can follow with 4.1.0 which works in mobile

gkc:
@XavierChanth I think we need to make `-i` mandatory for the openssh implementation as well; when it's not supplied we get `ssh -p 52777 gary@localhost` as output which means almost certain auth failure if you've got lots of identity files, vs `ssh -p 53257 -o StrictHostKeyChecking=accept-new -o IdentitiesOnly=yes gary@localhost -i /Users/gary/.ssh/noports` when `-i` is supplied

wdyt?

xavierchanth (approved):

xavierchanth:
> @XavierChanth I think we need to make `-i` mandatory for the openssh implementation as well; when it's not supplied we get `ssh -p 52777 gary@localhost` as output which means almost certain auth failure if you've got lots of identity files, vs `ssh -p 53257 -o StrictHostKeyChecking=accept-new -o IdentitiesOnly=yes gary@localhost -i /Users/gary/.ssh/noports` when `-i` is supplied
> 
> wdyt?

openssh has a list of fallback keys, hence why I didn't enforce it:

from the man pages:
```
-i identity_file
             Selects a file from which the identity (private key) for public key authentication is read.  You can also
             specify a public key file to use the corresponding private key that is loaded in ssh-agent(1) when the
             private key file is not present locally.  The default is ~/.ssh/id_rsa, ~/.ssh/id_ecdsa,
             ~/.ssh/id_ecdsa_sk, ~/.ssh/id_ed25519, ~/.ssh/id_ed25519_sk and ~/.ssh/id_dsa.  Identity files may also
             be specified on a per-host basis in the configuration file.  It is possible to have multiple -i options
             (and multiple identities specified in configuration files).  If no certificates have been explicitly
             specified by the CertificateFile directive, ssh will also try to load certificate information from the
             filename obtained by appending -cert.pub to identity filenames.
             ```
             
"The default is ~/.ssh/id_rsa, ~/.ssh/id_ecdsa, ~/.ssh/id_ecdsa_sk, ~/.ssh/id_ed25519, ~/.ssh/id_ed25519_sk and ~/.ssh/id_dsa."

gkc:
Ack 👍

PR #605, opened 2023-12-03
feat: pure dart interactive ssh -> gkc

xavierchanth:
**- What I did**
- Renamed some files to follow the new conventions
- Removed userKeyPairIdentity which is redundant with params.identityFile
- Started an early draft of the pure dart runShell()

**- How I did it**

**- How to verify it**

feat: pure dart interactive ssh -> gkc

### DIFF
--- a/packages/noports_core/lib/src/sshnp/sshnp.dart
+++ b/packages/noports_core/lib/src/sshnp/sshnp.dart
@@ -18,7 +18,7 @@ abstract interface class Sshnp {
     required SshnpParams params,
   }) {
     return SshnpUnsignedImpl(
-        atClient: atClient, params: params, userKeyPairIdentifier: null);
+        atClient: atClient, params: params);
   }
 
   /// Think of this as the "default" client - calls openssh
@@ -27,7 +27,7 @@ abstract interface class Sshnp {
     required SshnpParams params,
   }) {
     return SshnpOpensshLocalImpl(
-        atClient: atClient, params: params, userKeyPairIdentifier: null);
+        atClient: atClient, params: params);
   }
 
   /// Uses a dartssh2 ssh client - requires that you pass in the identity keypair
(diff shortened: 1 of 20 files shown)


packages/noports_core/lib/src/sshnp/util/ssh_session_handler/dart_ssh_session_handler.dart
```diff
@@ -146,7 +166,7 @@ class SshClientHelper {
     }
   }
 
-  Future<void> startForwarding(
+  Future<int> startForwarding(
```
xavierchanth, line 169: if fLocalPort is 0, then an ephemeral port is generated and we need to return it

packages/noports_core/lib/src/sshnp/util/ssh_session_handler/dart_ssh_session_handler.dart
```diff
@@ -69,13 +69,33 @@ mixin SshnpDartSshSessionHandler on SshnpCore
...
   @override
-  Future<SSHClient> startUserSession(
-      {required SSHClient tunnelSession}) async {
-    throw UnimplementedError();
-    // TODO v similar to startInitialTunnelSession
+  Future<SSHClient> startUserSession({
+    required SSHClient tunnelSession,
+  }) async {
+    if (identityKeyPair == null) {
+      throw SshnpError('Identity Key pair is mandatory with the dart client.');
```
xavierchanth, line 76: Wrote an early implementation, but the client is not authenticating, I think the tunnel is closing before authentication completes on the userSession:

```
INFO|2023-12-03 18:04:52.430560|Sshnp|Started port forwarding from localhost:54053 on local side to localhost:22 on remote side 
INFO|2023-12-03 18:04:52.430675|Sshnp|ssh session will terminate after 15 seconds if it is not being used 
Failed to authenticate as root@localhost:54053 : SSHAuthAbortError(Connection closed before authentication)
```

packages/noports_core/lib/src/sshnp/sshnp.dart
```diff
@@ -56,8 +52,6 @@ abstract interface class Sshnp {
   /// The parameters used to configure this Sshnp instance
   SshnpParams get params;
 
-  String? get userKeyPairIdentifier;
```
xavierchanth, line 59: I removed this variable entirely, since params.identityFile stores the exact same information.

PR #606, opened 2023-12-05
chore: Release v4.0.0

xavierchanth:
**- What I did**

Cleaned up all remaining odds and ends:
- Fixed path to dockerfiles which were moved in a previous PR
- Updated the release build so that the copied files are flattened into the release zip (i.e. contents of core/shell/windows folders are in the root of the release zip)
- Fixed up the install script
- Added the debug sshrvd to the notarization script

**- How I did it**

**- How to verify it**

chore: Release v4.0.0


.github/workflows/multibuild.yaml
```diff
@@ -2,6 +2,12 @@ name: Multibuild
 
 on:
   workflow_dispatch:
+    inputs:
+      x64_only:
+        description: "Run x64 builds only"
+        required: true
+        default: false
+        type: boolean
```
xavierchanth, line 10: Added this to speed up the ability to confirm/test x64 releases, since non-x64 releases can take up to 10x longer

packages/sshnoports/bundles/shell/install.sh
```diff
@@ -71,6 +84,10 @@ install_single_binary() {
   mkdir -p "$dest"
   cp "$script_dir/$1" "$dest/$1"
   echo "Installed $1 to $dest"
+  if is_root & ! [ -f "$user_bin_dir/$1" ] ; then
+    ln -sf "$dest/$1" "$user_bin_dir/$1"
+    echo "Linked $user_bin_dir/$1 to $dest"
+  fi
```
xavierchanth, line 90: If the binaries were installed as root, we need to do a symbolic to where the binaries were installed, since we install the binaries to /usr/local/bin if we are root

packages/sshnoports/bundles/shell/install.sh
```diff
@@ -164,8 +183,12 @@ post_headless_message() {
...
 install_headless_job() {
   job_name=$1
-  dest="$user_bin_dir/$job_name"
-  cp "$script_dir/headless/$job_name" "$dest"
+  mkdir -p "$user_bin_dir"
+  mkdir -p "$user_log_dir"
+  dest="$user_bin_dir/$job_name.sh"
+  if ! [ -f "$dest" ]; then
+    cp "$script_dir/headless/$job_name.sh" "$dest"
+  fi
```
xavierchanth, line 191: Make necessary folders, and don't overwrite existing job scripts when reinstalling

packages/sshnoports/bundles/shell/install.sh
```diff
@@ -174,12 +197,14 @@ install_headless_job() {
...
-  has_existing_entry=$(echo "$crontab_contents" | grep -Fxq "$cron_entry")
-  if $has_existing_entry; then
-    echo "cron job already installed, killing old $job_name process"
-    pgrep "$command" | xargs kill
+  if echo "$crontab_contents" | grep -Fxq "$cron_entry"; then
+    echo "cron job already installed, killing any old $job_name.sh processes"
+    pids=$(pgrep "$command")
+    if [ -n "$pids" ]; then
+     echo "$pids" | xargs kill
+    fi
```
xavierchanth, line 205: Only call the kill command if there are processes to kill

gkc (approved):

PR #607, opened 2023-12-05
feat: add new package-windows script

xavierchanth:
**- What I did**

- This only works on macos because I opted to use ditto instead of zip, but we required an apple silicon device to do the release anyways.

**- How I did it**

**- How to verify it**

feat: add new package-windows script

### DIFF
--- a/tools/package-windows-x64.sh
+++ b/tools/package-windows-x64.sh
@@ -1,8 +1,11 @@
 #!/bin/bash
 
-# This script is used to package the macOS x64 binaries for sshnoports.
-# It first unpacks the prebuilt binaries, then it calls the notarize-macos.sh
-# script to notarize the binaries
+# This script is used to package the windows x64 binaries for sshnoports.
+# It first unpacks the tgz archive of binaries, then it repacks them as a .zip
+# This script requires that you do this packaging on a mac, since it is expected
+# that you do it while notarizing the macos binaries to save time
+
+# This script expects the path to the tgz archive as an argument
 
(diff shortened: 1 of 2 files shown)


tools/package-windows-x64.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+# This script is used to package the macOS x64 binaries for sshnoports.
```
cpswan: The purpose of the script needs to be clarified

tools/package-windows-x64.sh
```diff
@@ -0,0 +1,31 @@
...
+FULL_PATH_TO_SCRIPT="$(realpath "${BASH_SOURCE[0]}")"
+SCRIPT_DIRECTORY="$(dirname "$FULL_PATH_TO_SCRIPT")"
+ROOT_DIRECTORY="$SCRIPT_DIRECTORY/.."
+
+if [ "$#" -ne 1 ]; then
+  echo "Usage: $0 <tgz file>"
+  exit 1
+fi
+
+if [ "$(uname)" != "Darwin" ]; then
```
cpswan, line 19: Why we're running a Windows related script on macOS deserves some explanation

Should we have something that runs on Windows for anybody doing this for themselves?

cpswan (changes requested):

tools/package-windows-x64.sh
xavierchanth: This script is only for packaging the release, if they want to build from source, they don't need this step.

cpswan (approved):

PR #659, opened 2024-01-02
feat: rv authentication and extra e2ee

### LINKED ISSUE #519: Implement challenge response on initial connections to rvd

- rvd's job is to allow two processes running on separated networks to establish a socket connection between them, via ports which the rvd assigns and then listens on
- Currently, some other process can in theory connect to a newly-opened rvd port in the brief window before the intended process has done so
- We would like to prevent this in a cryptographically secure way

- When requesting a port pair from the rvd via atProtocol, the requesting atClient should provide a unique one-time session ID as well as the atSigns of the two processes which are attempting to communicate. (Note that the same atSign could be being used by both processes)
- When a process tries to create a client socket to a port that the rvd has made available, then
  - the rvd will know what atSign should be connecting to that port
  - and the rvd will know the session ID
  - the rvd can therefore require the client to provide a base64-encoded cryptographic signature where the data being signed is the session ID, and the signing key is the client atSign's private encryption key
  - the rvd can then verify the signature using the client atSign's public encryption key
- If the client process does not supply a valid signature then the rvd should drop the connection
- If the client process has supplied a valid signature then the rvd can proceed as it currently does, and join this socket's i/o to the other process's socket's i/o (once the other socket client has also presented a valid signature)

### LINKED ISSUE #626: Implement encryption between client and the device

- rvd's facilitates processes running on separated networks to establish a socket connection between them, via ports which the rvd assigns and then listens on
- processes connecting to the ports should be able to encrypt and decrypt the data the way they wish it to be

- When requested, rvd opens two ports, let's say PortA and PortB for a given session
- Client connecting to PortA makes the information related to PortB available to the other side (Device)
- Client creates a symmetric encryption key and shares this information with the device
- Now information shared between client and the device will be encrypted and decrypted using the symmetric key

_No response_

### Additional context

_No response_

gkc:
Closes #626
Closes #519 

**- What I did**
- Added ability to authenticate to the socket rendezvous, and made this the 
  default behaviour.
- Added ability to end-to-end encrypt all traffic via the socket rendezvous (
  SR), and made this the default behaviour. This provides a good general defense
  against compromise by man-in-the-middle attacks if the two ends are
  communicating via the SR. The encryption is implemented by exchanging a 
  symmetric key via an ephemeral encryption keypair generated by the client 
  for every session.
- Added ability to "ping" a daemon for info about it, including which 
  features the daemon supports.
- By default, sshnp now immediately drops into a prompt for clients that don't
  support sshnp.canRunShell(). `-x` flag allows the ssh command to be output 
  instead of executing ssh immediately.
- Renamed everything sshrvd and sshrv to srvd and srv respectively

**- How to verify it**
- Tests pass


packages/noports_core/lib/src/common/at_ssh_key_util/local_ssh_key_util.dart
```diff
@@ -129,7 +129,8 @@ class LocalSshKeyUtil implements AtSshKeyUtil {
   }) async {
     // Check to see if the ssh public key is
     // supported keys by the dartssh2 package
-    if (!sshPublicKey.startsWith(RegExp(r'^(ecdsa-sha2-nistp)|(rsa-sha2-)|(ssh-rsa)|(ssh-ed25519)|(ecdsa-sha2-nistp)'))) {
+    if (!sshPublicKey.startsWith(RegExp(
+        r'^(ecdsa-sha2-nistp)|(rsa-sha2-)|(ssh-rsa)|(ssh-ed25519)|(ecdsa-sha2-nistp)'))) {
```
gkc: this was just dart format

packages/noports_core/lib/src/common/default_args.dart
```diff
@@ -18,6 +18,10 @@ class DefaultArgs {
   static const bool addForwardsToTunnel = false;
   static final bool allowLocalFileSystem =
       Platform.isLinux || Platform.isMacOS || Platform.isWindows;
+  static const bool authenticateClientToRvd = true;
+  static const bool authenticateDeviceToRvd = true;
+  static const bool encryptRvdTraffic = true;
+  static const bool discoverDaemonFeatures = true;
```
gkc: extra command line flags

packages/noports_core/lib/src/common/mixins/at_client_bindings.dart
```diff
@@ -3,19 +3,26 @@ import 'package:at_utils/at_utils.dart';
...
-        .notify(NotificationParams.forUpdate(atKey, value: value),
-            onSuccess: (NotificationResult notification) {
-      logger.info('SUCCESS:$notification with key: ${atKey.toString()}');
-    }, onError: (notification) {
-      logger.info('ERROR:$notification');
-    });
+    String value, {
+    required bool checkForFinalDeliveryStatus,
+    required bool waitForFinalDeliveryStatus,
+  }) async {
```
gkc: I added these params as required and dealt with the ripple effects as it is better in general to have these parameters explicitly exposed

packages/noports_core/lib/src/common/types.dart
```diff
@@ -1,12 +1,21 @@
 import 'package:noports_core/sshrv.dart';
 
-typedef SshrvGenerator<T> = Sshrv<T> Function(String, int, {int localSshdPort});
+typedef SshrvGenerator<T> = Sshrv<T> Function(
+  String,
+  int, {
+  required int localPort,
```
gkc: renamed localSshdPort to localPort in anticipation of future daemon versions which connect to ports other than the sshd port.

Also made it `required` to make sure all callers are being explicit

packages/noports_core/lib/src/common/types.dart
```diff
@@ -1,12 +1,21 @@
...
+  required bool bindLocalPort,
```
gkc: Client-side we now run an rv which binds to a local port on one side, and talks to the rvd on the other; the ssh tunnel is then created by sshing to the local port that the local sshrv is listening on

packages/noports_core/lib/src/common/types.dart
```diff
@@ -1,12 +1,21 @@
...
+  String? rvdAuthString,
+  String? sessionAESKeyString,
+  String? sessionIVString,
```
gkc: Parameters for auth and e2ee, for connections to rvd

packages/noports_core/lib/src/sshnp/impl/sshnp_dart_pure_impl.dart
```diff
@@ -2,14 +2,35 @@ import 'dart:async';
...
+      params.encryptRvdTraffic = false;
+    }
+    if (params.authenticateDeviceToRvd) {
+      logger.shout('$runtimeType: disabling authenticateDeviceToRvd flag');
+      params.authenticateDeviceToRvd = false;
+    }
+    if (params.authenticateClientToRvd) {
+      logger.shout('$runtimeType: disabling authenticateClientToRvd flag');
+      params.authenticateClientToRvd = false;
+    }
```
gkc: As mentioned above - the major work outstanding for these features is to be able to an RV in-memory, which we need to be able to do for pure dart apps e.g. the noports gui etc

packages/noports_core/lib/src/sshnp/impl/sshnp_openssh_local_impl.dart
```diff
@@ -66,6 +67,10 @@ class SshnpOpensshLocalImpl extends SshnpCore
 
     logger.info('Sending request to sshnpd');
 
+    final server = await ServerSocket.bind(InternetAddress.anyIPv4, 0);
+    int localRvPort = server.port;
+    await server.close();
```
gkc: Above code is to identify a port that is available for the rv process to bind to

packages/noports_core/lib/src/sshnp/impl/sshnp_openssh_local_impl.dart
```diff
@@ -94,6 +109,16 @@ class SshnpOpensshLocalImpl extends SshnpCore
       );
     }
 
+    /// Start sshrv
+    await sshrvdChannel.runSshrv(
+      directSsh: true,
+      localRvPort: localRvPort,
+      sessionAESKeyString: sshnpdChannel.sessionAESKeyString,
+      sessionIVString: sshnpdChannel.sessionIVString,
+    );
```
gkc: We now run an rv on the client side

packages/noports_core/lib/src/sshnp/impl/sshnp_openssh_local_impl.dart
```diff
@@ -94,6 +109,16 @@ class SshnpOpensshLocalImpl extends SshnpCore
...
+
+    await Future.delayed(Duration(seconds: 1));
```
gkc: Had to add this in order to give the sshrv process time to start up and bind to its port

packages/noports_core/lib/src/sshnp/impl/sshnp_unsigned_impl.dart
```diff
@@ -94,6 +95,9 @@ class SshnpUnsignedImpl extends SshnpCore with SshnpLocalSshKeyHandler {
       throw SshnpError('sshnpd did not acknowledge the request');
     }
 
+    /// Start sshrv
+    var bean = await sshrvdChannel.runSshrv(directSsh: false);
```
gkc: uhmmm this might have been added in error, will need to check it

packages/noports_core/lib/src/sshnp/models/sshnp_params.dart
```diff
@@ -319,8 +379,12 @@ class SshnpPartialParams {
...
     return SshnpPartialParams(
       profileName: args[SshnpArg.profileNameArg.name],
-      clientAtSign: args[SshnpArg.fromArg.name],
-      sshnpdAtSign: args[SshnpArg.toArg.name],
+      clientAtSign: args[SshnpArg.fromArg.name] == null
+          ? null
+          : AtUtils.fixAtSign(args[SshnpArg.fromArg.name]),
+      sshnpdAtSign: args[SshnpArg.toArg.name] == null
+          ? null
+          : AtUtils.fixAtSign(args[SshnpArg.toArg.name]),
```
gkc: Need atSigns to consistently have the `@` prefixed

packages/noports_core/lib/src/sshnp/util/sshrvd_channel/sshrvd_channel.dart
```diff
@@ -31,27 +33,34 @@ abstract class SshrvdChannel<T> with AsyncInitialization, AtClientBindings {
...
   final SshnpParams params;
   final String sessionId;
+  final String clientNonce = DateTime.now().toIso8601String();
 
   // * Volatile fields which are set in [params] but may be overridden with
   // * values provided by sshrvd
 
   String? _host;
-  int? _port;
+  int? _portA;
```
gkc: Renamed to help readability

packages/noports_core/lib/src/sshnp/util/sshrvd_channel/sshrvd_channel.dart
```diff
@@ -31,27 +33,34 @@ abstract class SshrvdChannel<T> with AsyncInitialization, AtClientBindings {
...
+  String? sessionAESKeyString;
+  String? sessionIVString;
+
   /// Whether sshrvd acknowledged our request
   @visibleForTesting
   SshrvdAck sshrvdAck = SshrvdAck.notAcknowledged;
 
   /// The port sshrvd is listening on
-  int? _sshrvdPort;
+  int? _portB;
```
gkc: Renamed to help readability

packages/noports_core/lib/src/sshnpd/sshnpd_impl.dart
```diff
@@ -329,9 +341,10 @@ class SshnpdImpl implements Sshnpd {
           'ssh Public Key received from ${notification.from} notification id : ${notification.id}');
       sshPublicKey = notification.value!;
 
-    // Check to see if the ssh public key is
-    // supported keys by the dartssh2 package
-    if (!sshPublicKey.startsWith(RegExp(r'^(ecdsa-sha2-nistp)|(rsa-sha2-)|(ssh-rsa)|(ssh-ed25519)|(ecdsa-sha2-nistp)'))) {
+      // Check to see if the ssh public key is
+      // supported keys by the dartssh2 package
+      if (!sshPublicKey.startsWith(RegExp(
+          r'^(ecdsa-sha2-nistp)|(rsa-sha2-)|(ssh-rsa)|(ssh-ed25519)|(ecdsa-sha2-nistp)'))) {
```
gkc: no change, just dart format

packages/noports_core/lib/src/sshrv/sshrv_impl.dart
```diff
@@ -34,10 +58,36 @@ class SshrvImplExec implements Sshrv<Process> {
...
+      '--local-port',
+      localPort.toString(),
+    ];
+    if (bindLocalPort) {
+      rvArgs.add('--bind-local-port');
+    }
+    Map<String, String> environment = {};
+    if (rvdAuthString != null) {
+      rvArgs.addAll(['--rv-auth']);
+      environment['RV_AUTH'] = rvdAuthString!;
```
gkc: Adding to environment rather than command line to prevent any potentially interesting info from being unnecessarily visible

packages/noports_core/lib/src/sshrv/sshrv_impl.dart
```diff
@@ -34,10 +58,36 @@ class SshrvImplExec implements Sshrv<Process> {
...
+    }
+    if (sessionAESKeyString != null && sessionIVString != null) {
+      rvArgs.addAll(['--rv-e2ee']);
+      environment['RV_AES'] = sessionAESKeyString!;
+      environment['RV_IV'] = sessionIVString!;
```
gkc: Adding to environment rather than command line to prevent any potentially interesting info from being unnecessarily visible

packages/noports_core/lib/src/sshrvd/build_env.dart
```diff
@@ -1,4 +1,5 @@
+import 'dart:io';
+
 class BuildEnv {
-  static final bool enableSnoop =
-      bool.fromEnvironment('ENABLE_SNOOP', defaultValue: false);
+  static final bool enableSnoop = (Platform.environment['ENABLE_SNOOP'] ?? "false").toLowerCase() == 'true';
```
gkc: I found that `bool.fromEnvironment` was not working properly on MacOS

packages/noports_core/lib/src/sshrvd/signature_verifying_socket_authenticator.dart
```diff
@@ -0,0 +1,118 @@
...
+    Completer<(bool, Stream<Uint8List>?)> completer = Completer();
+    bool authenticated = false;
+    StreamController<Uint8List> sc = StreamController();
+    logger.info('SignatureAuthVerifier $tag: starting listen');
+    socket.listen((Uint8List data) {
+      if (authenticated) {
+        sc.add(data);
+      } else {
+        try {
+          final message = String.fromCharCodes(data);
```
gkc: TODO: This code currently assumes that all of the authentication challenge response comes in a single flow. Ideally should parse the ints until we get a newline, then
- use everything up to the newline for auth
- discard the newline
- add everything after the newline to the output stream via `sc.add`

packages/noports_core/lib/src/sshrvd/socket_connector.dart
```diff
@@ -1,48 +1,123 @@
...
 
   /// Create the socket connector
-  SocketConnector socketStream = await SocketConnector.serverToServer(
-    serverAddressA: InternetAddress.anyIPv4,
-    serverAddressB: InternetAddress.anyIPv4,
-    serverPortA: portA,
-    serverPortB: portB,
-    verbose: snoop,
-  );
+  SocketConnector connector = await SocketConnector.serverToServer(
```
gkc: Renamed here (and other places) to 'connector'; the name `socketStream` was confusing me since SocketConnector is not a stream of Sockets

packages/noports_core/lib/src/sshrvd/socket_connector.dart
```diff
@@ -1,48 +1,123 @@
...
 
   /// Return the assigned ports to the main isolate
   sendPort.send((portA, portB));
 
   /// Shut myself down once the socket connector closes
   bool closed = false;
   while (closed == false) {
-    closed = await socketStream.closed();
+    logger.info('Waiting for connector to close');
+    closed = await connector.closed();
```
gkc: TODO don't need a `while` loop here

packages/noports_core/pubspec.yaml
```diff
@@ -24,6 +24,12 @@ dependencies:
   socket_connector: ^1.0.11
```
gkc: Set to 1.1.0 once socket_connector PR has been merged

packages/noports_core/pubspec.yaml
```diff
@@ -24,6 +24,12 @@ dependencies:
...
   uuid: ^3.0.7
 
+dependency_overrides:
+  socket_connector:
+    git:
+      url: https://github.com/gkc/socket_connector.git
+      ref: socket-authenticator-option
```
gkc: Remove once socket_connector PR has been merged and the package has been published

packages/noports_core/lib/src/sshnp/util/sshnpd_channel/sshnpd_channel.dart
```diff
@@ -87,16 +87,16 @@ abstract class SshnpdChannel with AsyncInitialization, AtClientBindings {
   Future<SshnpdAck> handleSshnpdPayload(AtNotification notification);
 
   /// Wait until we've received an acknowledgement from the daemon.
-  /// Returns true if the deamon acknowledged our request.
+  /// Returns true if the daemon acknowledged our request.
   /// Returns false if a timeout occurred.
-  Future<SshnpdAck> waitForDaemonResponse() async {
+  Future<SshnpdAck> waitForDaemonResponse({int maxWaitMillis = 15000}) async {
     // Timer to timeout after 10 Secs or after the Ack of connected/Errors
```
xavierchanth: This comment is out of sync with the changes, maybe change it to say
```dart
/// Timer to timeout after [maxWaitMillis] milliseconds or after the Ack of connected/Errors
```

packages/noports_core/lib/src/sshrvd/build_env.dart
xavierchanth: We should probably file a bug with the dart team

packages/noports_core/lib/src/sshrv/sshrv_impl.dart
xavierchanth: We should have good documentation explaining how the flag + environment variable works in tandem

xavierchanth: I think including this in the CLI printed usage is good enough, but it needs to be well written.

packages/noports_core/lib/src/sshrvd/signature_verifying_socket_authenticator.dart
```diff
@@ -40,60 +42,75 @@ class SignatureAuthVerifier implements SocketAuthVerifier {
...
+          logger.info('SignatureAuthVerifier $tag received data: $message');
+          // Expected message to be the JSON format with the below structure:
+          // {
+          // "signature":"<signature>",
+          // "hashingAlgo":"<algo>",
+          // "signingAlgo":"<algo>",
+          // "payload":{<the data which was signed>}
+          // }
```
xavierchanth: We should dartdoc (`///`) this comment block with message

packages/noports_core/lib/src/sshnp/impl/sshnp_openssh_local_impl.dart
xavierchanth: Can't we observe the process in which we start up sshrv and wait for that?

xavierchanth: nvm... we would have to make that process fork and return an exit code, which we currently don't do. (like ssh with ForkAfterAuthentication enabled)

packages/noports_core/lib/src/sshnp/impl/sshnp_unsigned_impl.dart
xavierchanth: Unsigned client always ran sshrv, this is the legacy v3 client. I see that it has been moved to start up after the notification for an sshd request is sent, this could lead to the daemon attempting to ssh connect back to the client before sshrv is open (but in the real world, I don't think this is likely to happen)

packages/noports_core/lib/src/sshnp/models/sshnp_params.dart
xavierchanth: We dropped support for non-atSign hosts, so I think we can enforce the check for those too.

packages/noports_core/lib/src/sshnp/util/sshnpd_channel/sshnpd_default_channel.dart
```diff
@@ -66,8 +68,16 @@ mixin SshnpdDefaultPayloadHandler on SshnpdChannel {
...
+
       ephemeralPrivateKey = daemonResponse['ephemeralPrivateKey'];
+      logger.info('Received ephemeralPrivateKey: $ephemeralPrivateKey');
+
+      sessionAESKeyString = daemonResponse['sessionAESKey'];
+      logger.info('Received sessionAESKey: $sessionAESKeyString');
+
+      sessionIVString = daemonResponse['sessionIV'];
+      logger.info('Received sessionIV: $sessionIVString');
+
```
xavierchanth: Should we be logging these values?

xavierchanth: Could pose a risk to someone who is screen sharing, or sharing logs for debugging...

packages/noports_core/lib/src/common/features.dart
```diff
@@ -0,0 +1,8 @@
+/// Features which can be supported by the NoPorts Daemon
+enum DaemonFeatures {
+  /// authenticate when connecting to the Socket Rendezvous (sr)
+  srAuth,
+
+  /// End-to-end encrypt traffic sent via the SocketRendezvous (sr)
+  srE2ee,
+}
```
xavierchanth: Should we add share public key to this list of features? (i.e. whether -s flag is on or off)

packages/noports_core/lib/src/common/validation_utils.dart
```diff
@@ -1,4 +1,5 @@
 import 'dart:convert';
+import 'dart:io';
```
xavierchanth: I was very meticulous to wrap all io operations into src/common/io_types.dart and src/common/file_system_utils.dart so that we can mock it in the unit tests and ensure that we have virtualized operations for the pure dart client.

Ideally, we don't import dart:io outside of these files I mentioned above.

packages/noports_core/lib/src/sshnp/models/sshnp_arg.dart
```diff
@@ -350,18 +351,28 @@ class SshnpArg {
   );
   static const authenticateClientToRvdArg = SshnpArg(
```
xavierchanth: Can we add an abbreviation to these authentication flags?

Maybe -a for client auth, and -A for daemon auth?

packages/noports_core/lib/src/sshrvd/socket_connector.dart
```diff
@@ -103,10 +106,9 @@ void socketConnector(ConnectorParams connectorParams) async {
...
 
-  logger.warning(
-      'Finished session ${sshrvdSessionParams.sessionId}'
-          ' for ${sshrvdSessionParams.atSignA} to ${sshrvdSessionParams.atSignB}'
-          ' using ports [$portA, $portB]');
+  logger.info('Finished session ${sshrvdSessionParams.sessionId}'
+      ' for ${sshrvdSessionParams.atSignA} to ${sshrvdSessionParams.atSignB}'
+      ' using ports [$portA, $portB]');
 
   Isolate.current.kill();
```
xavierchanth: Question about isolates as we scale this:

What happens to the memory of the current isolate when we kill it, and how does the GC handle it?

packages/noports_core/lib/src/sshnp/impl/notification_request_message.dart
```diff
@@ -1,33 +1,29 @@
...
-  }
+  Map<String, dynamic> toJson() => {
+        'direct': direct,
+        'sessionId': sessionId,
+        'host': host,
+        'port': port,
+        'authenticateToRvd': authenticateToRvd,
+        'clientNonce': clientNonce,
+        'rvdNonce': rvdNonce,
+      };
```
xavierchanth: Going forward, we might want to consider a lightweight JSON spec for these various payloads - considering we have a Dart & Python impl, with C on the way.

xavierchanth: This might cause more tech debt, without any real benefit though...

packages/noports_core/lib/src/sshrvd/build_env.dart
```diff
@@ -1,4 +1,4 @@
 class BuildEnv {
-  static final bool enableSnoop =
-      bool.fromEnvironment('ENABLE_SNOOP', defaultValue: false);
+  static final bool enableSnoop = true;
+      // bool.fromEnvironment('ENABLE_SNOOP', defaultValue: false);
```
xavierchanth: Are we eliminating the debug build?

packages/noports_core/lib/src/sshrv/sshrv_impl.dart
```diff
@@ -73,13 +73,18 @@ class SshrvImplDart implements Sshrv<SocketConnector> {
...
         socketPortA: localSshdPort,
         socketAddressB: hosts[0],
         socketPortB: streamingPort,
         verbose: true,
       );
+
+      // TBD - Gary/Xavier/Jagan
+      // Should we expose the sockets from socketConnector? whats the best way to do this.
+      //authenticationProvider?.authenticate(socketConnector.socketB);
+      return socketConnector;
```
xavierchanth: I am mostly unopinionated on this, but we should have some way to kill the socket connector externally, I have thought about some use-cases where this is useful.

packages/noports_core/lib/src/sshnp/impl/sshnp_dart_pure_impl.dart
xavierchanth: As mentioned in one of my other comments, it would be useful to have an in-memory RV with the openssh client as well, then we can call the ssh command as a child process using Process.run()

packages/noports_core/lib/src/sshnp/models/sshnp_arg.dart
```diff
@@ -346,4 +350,44 @@ class SshnpArg {
...
+    name: 'encrypt-rvd-traffic',
+    help: 'When true, traffic via the socket rendezvous is encrypted,'
+        ' in addition to whatever encryption the traffic already has'
+        ' (e.g. an ssh session)',
+    defaultsTo: DefaultArgs.encryptRvdTraffic,
+    format: ArgFormat.flag,
+    parseWhen: ParseWhen.commandLine,
+    mandatory: false,
+  );
+  static const discoverDaemonFeaturesArg = SshnpArg(
```
xavierchanth: would be nice to have an abbreviation for this as well.

xavierchanth:
@gkc @VJag general feedback:

- Make sure to wrap all dart:io calls as I had previously done with io_types.dart and file_system_utils.dart, we need everything wrapped for two reasons:
  - unit testing with mock io calls
  - the pure-dart client must not depend on any io calls, so we must avoid leaking io calls into the base classes
- end2end tests need to be fixed, we currently build in a docker image from a branch and then create run images from that, instead we should download from the releases, extract and call the binary from there. We also need to define testing matrices which understand the differences between version and test against older versions / limited features sets correctly.

packages/noports_core/lib/src/sshnp/util/sshnpd_channel/sshnpd_default_channel.dart
gkc: Now logging the encrypted values

packages/noports_core/lib/src/sshnp/util/sshnpd_channel/sshnpd_channel.dart
gkc: Done

packages/noports_core/lib/src/sshrvd/signature_verifying_socket_authenticator.dart
gkc: Done

packages/noports_core/lib/src/common/features.dart
gkc: Good idea!

packages/noports_core/lib/src/common/validation_utils.dart
gkc: was an unnecessary import, fixed

packages/noports_core/lib/src/sshnp/models/sshnp_arg.dart
gkc: done

packages/noports_core/lib/src/sshrvd/socket_connector.dart
gkc: Maybe something that @cconstab has looked at on the production SR daemons?

packages/noports_core/lib/src/sshnp/impl/notification_request_message.dart
gkc: I've created [a ticket](https://github.com/atsign-foundation/noports/issues/692) for now

packages/noports_core/lib/src/sshrvd/build_env.dart
gkc: this was just for my convenience while actively developing this PR's features

packages/noports_core/lib/src/sshnp/models/sshnp_arg.dart
gkc: Have added as 'F'
Have also added abbreviation 'E' for 'encrypt-rvd-traffic'

packages/noports_core/lib/src/sshrv/sshrv_impl.dart
gkc: one can call close() on a socket connector

xavierchanth (approved):

PR #663, opened 2024-01-04
feat: remove paramiko and use openssh binary

Xlin123:
**- What I did**
 Removed paramiko which was used for reverse ssh in order to address this [code scanning alert](https://github.com/atsign-foundation/sshnoports/security/code-scanning/16)

**- How I did it**
Removed paramiko and used the binary.

**- How to verify it**
 Use a sshnp command like the following 
 `sshnp -f {atsign} -t {atsign} -h {host} -d {device} --legacy-daemon`

feat: remove paramiko and use openssh binary

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -2,7 +2,7 @@
   // Use IntelliSense to learn about possible attributes.
   // Hover to view descriptions of existing attributes.
   // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
-  "version": "0.3.0",
+  "version": "0.4.2",
   "configurations": [
     {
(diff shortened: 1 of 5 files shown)


Xlin123:
@cpswan @XavierChanth I'm not sure what best practices for versioning and removing paramiko from poetry / requirements.txt are. If you could give me some help there would be greatly appreciated.

cpswan:
I've set this to draft as I think we should have an arch discussion before pushing ahead with this. I'd like to better understand the pros and cons of removing python native capability in exchange for other dependencies.

Xlin123:
After discussion in arch, decided to opt for python native solution.

PR #698, opened 2024-01-19
chore: added msix config

sachins-geekyants:
**- What I did**
-  msix config setup

**- How I did it**

**- How to verify it**

### DIFF
--- a/packages/dart/sshnp_flutter/pubspec.yaml
+++ b/packages/dart/sshnp_flutter/pubspec.yaml
@@ -81,7 +81,7 @@ flutter_launcher_icons:
 
 msix_config:
   display_name: "SSH No Ports Desktop"
-  publisher_display_name: "The @ Company"
+  publisher_display_name: "Atsign"
   identity_name: "TheCompany.SSHNoPortsDesktop"
   publisher: CN=BBFE1D0B-F713-4C7F-B375-5EA851CBB1FF
   msix_version: 1.0.0.0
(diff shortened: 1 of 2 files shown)


packages/dart/sshnp_flutter/pubspec.yaml
```diff
@@ -77,3 +78,13 @@ flutter_launcher_icons:
   macos:
     generate: true
     image_path: "assets/images/app_logo.png"
+
+msix_config:
+  display_name: "SSH No Ports Desktop"
+  publisher_display_name: "The @ Company"
```
xavierchanth: Are we able to update the publisher name to Atsign?

xavierchanth (approved):

packages/dart/sshnp_flutter/pubspec.yaml
```diff
@@ -77,3 +78,13 @@ flutter_launcher_icons:
   macos:
     generate: true
     image_path: "assets/images/app_logo.png"
+
+msix_config:
+  display_name: "SSH No Ports Desktop"
+  publisher_display_name: "Atsign"
+  identity_name: "TheCompany.SSHNoPortsDesktop"
```
gkc, line 85: Should we change the identity name from "TheCompany.SSHNoPortsDesktop" to "Atsign.SSHNoPortsDesktop" ? Or is it not relevant / important?

sachins-geekyants: It's for internal use and is not shown on windows stores.
We can keep it as it is.

gkc: > It's for internal use and is not shown on windows stores. We can keep it as it is.

I think it would be better to change it unless that will cause a problem

sachins-geekyants: It can not be changed even if we update organisation name, identity_name will remain same.

PR #699, opened 2024-01-19
chore: update dependencies for at commons 4.0.0

xavierchanth:
**- What I did**

- Updated locally dependency versions
- Put in place dependency overrides for at_client,at_onboarding_cli,at_auth,at_server_status on both noports_core and sshnoports
- Fixed lint errors related to value assertions:
  - Stopped checking whether AtKey.key or AtKey.metadata are null
  - Fixed a potentially null notification.value in the sshrvd_impl notification handler

**- How I did it**

**- How to verify it**

- noports_core Unit tests for lint errors and type errors
- e2e tests for any functional errors

chore: update dependencies for at commons 4.0.0


xavierchanth:
Thanks @purnimavenkatasubbu for investigating this further

@gkc we will also have to update the dependencies for sshnp_flutter, it should be as simple as updating the same dependencies as for sshnoports.

gkc:
All looks fine but I presume this should not be merged until the packages have been published and we can remove the dependency overrides?

gkc (commented):
All looks fine but I presume this should not be merged until the packages have been published and we can remove the dependency overrides?

xavierchanth:
@purnimavenkatasubbu can you follow up on this PR when you get a chance?

purnimavenkatasubbu:
@XavierChanth I have raised a new PR #767, I think this PR has some missing changes.

PR #710, opened 2024-01-22
ci: allow local multi build

xavierchanth:
**- What I did**

- Enabled multibuild to work against the local noports_core and pubspec.lock dependencies
- Should be more consistent builds since we can commit pubspec.lock without error
- We can also do test builds without needing to release noports_core first

**- How I did it**

**- How to verify it**

ci: allow local multi build


xavierchanth:
Oops PR to wrong base

PR #711, opened 2024-01-22
ci: allow local multi build

xavierchanth:
**- What I did**

- Enabled multibuild to work against the local noports_core and pubspec.lock dependencies
- Should be more consistent builds since we can commit pubspec.lock without error
- We can also do test builds without needing to release noports_core first

**- How I did it**

**- How to verify it**

ci: allow local multi build


tools/package-macos-arm64.sh
```diff
@@ -11,33 +11,37 @@ ROOT_DIRECTORY="$SCRIPT_DIRECTORY/.."
...
-  echo "Usage: $0"
-  exit 1
+	echo "Usage: $0"
+	exit 1
 fi
 
 if [ "$(uname)" != "Darwin" ]; then
-  echo "This script is only for macOS";
-  exit 1;
+	echo "This script is only for macOS"
```
cpswan, line 19: Is the switch from spaces to tabs intentional?

cpswan (approved):

PR #718, opened 2024-01-23
fix: empty session and tunnel username are treated as null.

### LINKED ISSUE #712: Empty Session and Tunnel Username prevents a successful ssh in the desktop app
### Describe the bug

SSH Fails if the session or tunnel username is empty. Empty strings should be treated as null.

### Steps to reproduce

1. First I created a profile without that does not require a session username
2. Then I tried to ssh and it failed

### Expected behavior

A profile that does not require a session username should not fail to ssh when a session username is not provided.

### Screenshots

_No response_

### Smartphones

- Device: M1 Mac
- OS: MacOS

- Version: 14.1.1

SSH No Ports Desktop

### Additional context

_No response_

CurtlyCritchlow:
**- What I did**

- Replaces empty string as null when using the session and tunnel username.
- Added additional validation to SR Address, Device address, remote sshd port and local forwarding port.

**- How I did it**

**- How to verify it**

1. Run the app
2.  edit or create a profile. 
3. enter an invalid value in the form field

Improve validation when creating profiles and empty session and tunnel username are treated as null.

### DIFF
--- a/packages/dart/sshnp_flutter/lib/src/utility/form_validator.dart
+++ b/packages/dart/sshnp_flutter/lib/src/utility/form_validator.dart
@@ -10,12 +10,14 @@ class FormValidator {
     return null;
   }
 
-  static String? validateRequiredIntField(String? value) {
+  static String? validateRequiredPortField(String? value) {
     String valid = r'^[0-9]+$';
     if (value?.isEmpty ?? true) {
       return kEmptyFieldValidationError;
     } else if (!RegExp(valid).hasMatch(value!)) {
-      return kIntFieldValidationError;
+      return kPortFieldValidationError;
+    } else if (!(int.parse(value) >= 0 && int.parse(value) <= 65535)) {
+      return kPortFieldValidationError;
     }
     return null;
   }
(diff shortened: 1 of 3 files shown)


packages/dart/sshnp_flutter/lib/src/utility/form_validator.dart
```diff
@@ -10,12 +10,23 @@ class FormValidator {
...
+  static String? validateRequiredIntField(String? value) {
+    String valid = r'^[0-9]+$';
+    if (value?.isEmpty ?? true) {
+      return kEmptyFieldValidationError;
+    } else if (!RegExp(valid).hasMatch(value!)) {
+      return kIntFieldValidationError;
+    }
+    return null;
+  }
+
```
xavierchanth, line 24: Specifically for ports, we can do one better, and say the range is 0-65535

CurtlyCritchlow: done

xavierchanth (approved):

xavierchanth:
Nice work @CurtlyCritchlow

PR #720, opened 2024-01-24
feat: v5 args tweaking

gkc:
**- What I did**
- feat: Sshnp: default discoverDaemonFeatures to false
- fix: SshnpArg: when creating ArgParser, add aliases (if any) for flags as well as for (currently) options
- feat: SshnpArg: Remove abbreviations for the new flags, add suitabe short aliases instead
- feat: supply usageLineLength to sshnpd's ArgParser
- feat: use fork of `args` package, so that usage output includes alias(es)
- chore: style: dart format
- refactor: better readability in print_version.dart

**- How I did it**
- See commits

**- How to verify it**
- Tests pass
- binaries' usage output should include alias(es) - e.g. for sshnp
  ```
    --[no-]authenticate-client-to-rvd,--[no-]ac 
    --[no-]authenticate-device-to-rvd,--[no-]ad
    --[no-]encrypt-rvd-traffic,--[no-]et
    --[no-]discover-daemon-features,--[no-]ddf
  ```


packages/dart/sshnoports/lib/src/print_version.dart
```diff
@@ -1,9 +1,10 @@
...
-import 'package:sshnoports/src/version.dart' as release;
-import 'package:noports_core/version.dart' as package;
+import 'package:sshnoports/src/version.dart' as binaries;
+import 'package:noports_core/version.dart' as core;
 
 /// Print version number
 void printVersion() {
-  stderr.writeln('Version : ${release.packageVersion} (core: ${package.packageVersion}');
+  stderr.writeln('Version : ${binaries.packageVersion}'
+      ' (core: ${core.packageVersion}');
```
xavierchanth, line 9: Missing the closing bracket for (core: x.x.x)

xavierchanth (approved):

packages/dart/sshnoports/lib/src/print_version.dart
gkc: Thanks - will fix in next PR

PR #765, opened 2024-01-29
fix: Handle problems caused by the tmux function in install.sh masking the tmux binary itself

### LINKED ISSUE #761: v5.0.1 install.sh does not work to upgrade sshnp/sshnpd
### Describe the bug

various problems with upgrade process that need to be addressed

If sshnp/sshnpd are in use then they cannot be copied over (I have sucessfully `dodged this by mv sshnpd sshnpd.old` then using `cp`

The tmux install fails badly see bel;ow

### Steps to reproduce

1. First I install sshnp/sshnpd - TMUX / bin installs tested
2. Then I try to install over them to upgrade
3. Things fail on Mac/Linux

### Expected behavior

just works/upgrades with single command idealy so you could use sshnp to upgrade in a single command to make things easier if you had 100 or 1000 machines!

Perhaps auto upgrade solution ?

### Screenshots

```
constab@tarial sshnp % ./install.sh tmux sshnpd                                                                             
Installed sshnpd to /Users/cconstab/.local/bin
Installed srv to /Users/cconstab/.local/bin
cron job already installed, killing old tmux session
Unknown tmux service: kill-session
./install.sh [command]
Available commands:
at_activate     - install at_activate
sshnp           - install sshnp
sshnpd          - install sshnpd
srv           - install srv
srvd          - install srvd
binaries        - install all base binaries

debug_srvd    - install srvd with debugging enabled
debug           - install all debug binaries

all             - install all binaries (base and debug)

headless <job>  - install a headless cron job
                  available jobs: [sshnpd, srvd]

tmux <service>  - install a service in a tmux session
                  available services: [sshnpd, srvd]
cconstab@tarial
```

### Smartphones

- sshnp v5.0.1

_No response_

### Additional context

_No response_

gkc:
Fixes #761 

**- What I did**
- fix: install.sh: Handle problems caused by the tmux function masking the tmux binary itself
- fix: install.sh:
  - When copying binaries during installation, rename existing binaries as they may be in use (which is a problem in, for example, Windows)
  - Exit the install script if either the rename of existing or the copying of new binaries fails
- fix: install.sh: for installs using tmux, separate the handling of crontab installation from the handling of tmux process killing
- feat: install.sh: If the installation killed a process running as a tmux session, then also restart it

**- How I did it**
- See file diff
- Note that I didn't rename the `tmux` function but instead used `command tmux` whenever needing to use the tmux command itself
- While mulling over all of this, I came to the opinion that just killing an existing tmux session might be unhelpful and that it would be more helpful to restart the tmux session if there was one there to kill. However I preserve the current behaviour of not starting the tmux session if it is not already running.

### DIFF
--- a/packages/dart/sshnoports/bundles/shell/install.sh
+++ b/packages/dart/sshnoports/bundles/shell/install.sh
@@ -91,12 +91,12 @@ install_single_binary() {
     dest="$user_bin_dir"
   fi
   mkdir -p "$dest"
-  cp "$script_dir/$1" "$dest/$1"
-  echo "Installed $1 to $dest"
+  cp -f "$script_dir/$1" "$dest/$1"
+  echo "=> Installed $1 to $dest"
   if is_root & ! [ -f "$user_bin_dir/$1" ] ; then
     mkdir -p "$user_bin_dir"
     ln -sf "$dest/$1" "$user_bin_dir/$1"
-    echo "Linked $user_bin_dir/$1 to $dest"
+    echo "=> Linked $user_bin_dir/$1 to $dest"
   fi
 }
 
@@ -116,7 +116,7 @@ install_debug_binary() {
   fi
   mkdir -p "$dest"
   cp "$script_dir/debug/$1" "$bin_dir/debug_$1"
-  echo "Installed debug_$1 to $dest"
+  echo "=> Installed debug_$1 to $dest"
 }
 
 install_debug_binaries() {
@@ -218,13 +218,13 @@ install_headless_job() {
   crontab_contents=$(crontab -l 2>/dev/null)
 
   if echo "$crontab_contents" | grep -Fxq "$cron_entry"; then
-    echo "cron job already installed, killing any old $job_name.sh processes"
+    echo "=> cron job already installed, killing any old $job_name.sh processes"
     pids=$(pgrep "$command")
     if [ -n "$pids" ]; then
      echo "$pids" | xargs kill
     fi
   else
-     echo "Installing cron job: $cron_entry"
+     echo "=> Installing cron job: $cron_entry"
     (echo "$crontab_contents"; echo "$cron_entry") | crontab -;
   fi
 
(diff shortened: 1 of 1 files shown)


xavierchanth (approved):

cconstab:
So I tried this in today and it failed to copy over sshnpd the first time it would seem. The second time worked fine see log below..

```
cconstab@orac:~/sshnp$ vi install.sh
cconstab@orac:~/sshnp$ ./install.sh tmux sshnpd
cconstab@orac:~/sshnp$
cconstab@orac:~/sshnp$
cconstab@orac:~/sshnp$
cconstab@orac:~/sshnp$ ~/.local/bin/sshnpd
-bash: /home/cconstab/.local/bin/sshnpd: No such file or directory
cconstab@orac:~/sshnp$ ./install.sh tmux sshnpd
=> Installed sshnpd to /home/cconstab/.local/bin
cconstab@orac:~/sshnp$ ~/.local/bin/sshnpd
Version : 5.0.1 (core: 6.0.1)
-k, --key-file,--keyFile              Sending atSign's keyFile if not in ~/.atsign/keys/
-a, --atsign (mandatory)              atSign of this device
-m, --manager (mandatory)             Managers atSign, that this device will accept triggers from
-d, --device                          Send a trigger to this device, allows multiple devices share an atSign
                                      (defaults to "default")
-s, --[no-]sshpublickey               When set, will update authorized_keys to include public key sent by manager
-u, --[no-]un-hide,--[no-]username    When set, makes various information visible to the manager atSign - e.g. username, version, etc
-v, --[no-]verbose                    More logging
    --ssh-client                      What to use for outbound ssh connections.
                                      [openssh (default), dart]
    --root-domain                     atDirectory domain
                                      (defaults to "root.atsign.org")
    --local-sshd-port                 port on which sshd is listening locally on localhost
                                      (defaults to "22")
    --ephemeral-permissions           The permissions which will be added to the authorized_keys file for the ephemeral public keys which are generated when a client is
                                      connecting via forward ssh e.g. PermitOpen="host-1:3389",PermitOpen="localhost:80"
                                      (defaults to "")
    --ssh-algorithm                   Use RSA 4096 keys rather than the default ED25519 keys
                                      [ssh-ed25519 (default), ssh-rsa]
    --storage-path                    Directory for local storage. Defaults to $HOME/.sshnp/${atSign}/storage

Invalid argument(s): Option atsign is mandatory.
cconstab@orac:~/sshnp$
```

PR #774, opened 2024-01-31
fix: profiles are refreshed when first sync after onboarding is completed

### LINKED ISSUE #754: Make sure the dashboard page reloads to show the profiles after initial sync

CurtlyCritchlow:
…eted

**- What I did**
    Fixes #754 
    Clear KeyChain the first time the app is run
**- How I did it**
   Add a progress listener that refreshes the profiles controller when sync is completed, this is only done after onboarding when the user is on the Home Screen.
**- How to verify it**

1. Run the app, 
2. Onboard 
3. A snackbar will be show stating if sync failed, in progress or completed

Profiles are refreshed when the first sync after onboarding is completed

### DIFF
--- a/packages/dart/sshnp_flutter/lib/src/utility/my_sync_progress_listener.dart
+++ b/packages/dart/sshnp_flutter/lib/src/utility/my_sync_progress_listener.dart
@@ -1,0 +1,42 @@
+import 'dart:developer';
+
+import 'package:at_onboarding_flutter/at_onboarding_flutter.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:go_router/go_router.dart';
+import 'package:sshnp_flutter/src/presentation/widgets/utility/custom_snack_bar.dart';
+import 'package:sshnp_flutter/src/utility/extensions/go_router_extensions.dart';
+
+import '../controllers/config_controller.dart';
+import '../controllers/navigation_controller.dart';
+import '../repository/authentication_repository.dart';
+import '../repository/navigation_repository.dart';
+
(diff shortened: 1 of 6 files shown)


packages/dart/sshnp_flutter/lib/src/utility/my_sync_progress_listener.dart
```diff
@@ -0,0 +1,42 @@
...
+    if (isSyncStatusSuccess && isCurrentRouteHome && isFirstRun) {
+      await ref.read(configListController.notifier).refresh();
+      CustomSnackBar.notification(content: 'Sync Completed');
+      await AuthenticationRepository().setFirstRun(false);
+      log('Initial Sync completed');
+    } else if (isSyncStatusFailure && isCurrentRouteHome && isFirstRun) {
+      await ref.read(configListController.notifier).refresh();
+      CustomSnackBar.notification(content: 'Sync failed... Retrying');
+
+      log('Initial Sync failed');
```
xavierchanth, line 39: @CurtlyCritchlow maybe we should implement a counter to stop trying after 3 tries or something

xavierchanth (approved):

packages/dart/sshnp_flutter/lib/src/utility/my_sync_progress_listener.dart
CurtlyCritchlow: We want it to keep retrying but the snackbar should only be show once. I'll make the update.

xavierchanth: Okay, if you think it's fine as is

PR #804, opened 2024-02-07
feat: have the daemon support multiple manager atSigns

gkc:
**- What I did**
- feat: have daemon support multiple manager atSigns
  - made `managerAtsign` into a list and renamed to `managerAtsigns` in Sshnpd, SshnpdImpl and SshnpdParams
  - changed the long arg name for the managerAtsigns to `managers` but preserved `manager` as an alias
  - modified SshnpdImpl.isFromAuthorizedAtsign accordingly
  - extracted the username sharing code into its own function called from SshnpdImpl.run
  - modified the username and device_info sharing functions to iterate through the manager atSigns and share (or un-share) with each one
- dependency management
  - build: sshnoports: updated dependencies at_onboarding_cli, at_client, at_lookup to latest so no dependency conflicts with latest noports_core which uses at_chops ^2.0.0 and at_commons ^4.0.0
  - build: noports_core: updated at_chops dependency to ^2.0.0
  - build: sshnp_flutter: override at_client_mobile and at_commons dependency versions. Temporary until https://github.com/atsign-foundation/at_client_sdk/issues/1232 and https://github.com/atsign-foundation/at_widgets/issues/794 have been resolved
- bug fixes
  - fix: set ccd (cached cascade delete) for the username and device_info so that if they are un-shared then their cached copies are deleted from the manager atSigns
  - fix: sshnp client: sshnpd_channel: resolveRemoteUsername was correctly detecting if remote username had not been shared, but was actually returning null rather than (as intended) throwing an SshnpError

### DIFF
--- a/packages/dart/noports_core/CHANGELOG.md
+++ b/packages/dart/noports_core/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 6.0.4
+- build: upgrade at_chops dependency to ^2.0.0
+
 # 6.0.3
 - chore: downgrade at_commons dependency from ^4.0.1 to ^4.0.0
 - chore: downgrade at_client dependency from ^3.0.75 to ^3.0.73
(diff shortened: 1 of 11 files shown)


gkc:
@XavierChanth fyi for the c daemon

cconstab (approved):
Gary works as expected under manual testing.. TY

PR #912, opened 2024-03-28
fix: Improvements to universal installer script

cpswan:
fixes #906

**- What I did**

Numerous fixes to get script working as expected on Linux

**- How I did it**

Test iterations on clean Ubuntu 22.04 and Amazon Linux 2023 VMs

**- How to verify it**

Run against a test VM of your choice.

**- Description for the changelog**

fix: Improvements to universal installer script

PR #913, opened 2024-03-28
fix: the rest of universal installer

### LINKED ISSUE #906: universal.sh / multibuild changes required for 5.1.0 release
What I have done:

- laid the foundation for universal.sh script
- updated the multibuild action to update the bound release version for universal.sh and export as an artifact when doing a tagged build
- Added npt to install.sh (the more manual installer)
- Modified the install.sh and bundles to prepare for universal.sh 

What needs doing:
- [x] create a new branch off of trunk
- [x] change the sshnp_version metadata in universal.sh to match the prerelease containing the new bundle format ("5.1.0-rc.5")
- [x] run universal.sh to test and iterate; known issues:
  - [ ] some variables are used when unset in certain flows, you may need to change them from `unset foo` to `foo=''`
    - This is done to ensure we are intentional about the state of variables in the script
  - [x] unarchiving on linux is running into blockers
  - [x] can't assume ~/.local/bin exists
  - [x] can't assume ~/.local/bin is on PATH
  - [x] need to copy binaries to where they'll be run from
  - [x] need to copy template for systemd unit
  - [x] need to copy template for tmux/headless script
  - [x] on linux we assume systemd if we have root access, we should also make sure systemd is installed before trying to use it
- [ ] Add npt to the codesign list in multibuild

xavierchanth:
fixed #906 
closes #912 

**- What I did**

- Fixed the launchd installation
- Changed the artifact name in multibuild so it is unique across builds
- Added an extra print statement for clarity to the magic script
- cleaned up remaining bugs and input stuff in universal.sh

**- How I did it**

**- How to verify it**

If you get stuck please run with -v and send me the logs.
I will sort out the rest from there.
 
Places tested:

- [x] @xavierchanth's m1 macbook (client, device, both)
- [x] Ubuntu 22.04 VM (client, device, both) - using interactive prompts
- [x] Ubuntu 22.04 VM (client, device, both) - using command line flags
- [x] AL2023 VM (device) - using interactive prompts
- [x] AL2023 VM (device) - using command line flags

v5.1.0-rc.11 tests (with macbook fixes & new multi atkeys selector for @gkc):
- [x] m1 macbook - (device - many atkeys files available)
- [x] m1 macbook - (device - single atkeys file available)
- [x] Ubuntu 22.04 - (device - 2 atkeys files)
- [x] Ubuntu 22.04 - (device - 1 atkeys files)

fix: the rest of universal installer

### DIFF
--- a/packages/dart/sshnoports/bundles/universal.sh
+++ b/packages/dart/sshnoports/bundles/universal.sh
@@ -452,6 +452,7 @@ get_device_atsign() {
 }
 
 get_installed_atsigns() {
+  clientOrDevice="$1"
   atkeycount=0
   atkeys=""
   selectedatsign=""
@@ -476,13 +477,11 @@ get_installed_atsigns() {
           ;;
       esac
     else
-      echo "${atkeycount} atKeys found: ${atkeys}"
-      echo "Which one should be used:"
       echo "0) None"
       for i in $(seq 1 $atkeycount); do
         echo "$i) @$(echo "$atkeys" | cut -d' ' -f"$i")"
       done
-      printf '$ '
+      printf '=> Found .atKeys for %s atSigns. Choose %s atSign : $ ' "$atkeycount" "$clientOrDevice"
       read -r selectedinput
 
       selectedindex=$((selectedinput))
@@ -542,7 +541,7 @@ client() {
 
   # get the inputs for the magic script
   if [ -z "$client_atsign" ]; then
-    get_installed_atsigns
+    get_installed_atsigns "client"
     client_atsign="$selectedatsign"
     get_client_atsign
   fi
@@ -602,7 +601,7 @@ client() {
   write_metadata "$magic_script" "host_atsign" "$(norm_atsign "$host_atsign")"
   write_metadata_array "$magic_script" "devices" "$devices"
 
-  echo "Run np.sh to get your quick pick of devices to connect to."
+  echo "Run ${bin_path}/np.sh to get your quick pick of devices to connect to."
 }
 
 # DEVICE INSTALLATION #
@@ -628,7 +627,7 @@ device() {
   get_client_atsign
 
   if [ -z "$device_atsign" ]; then
-    get_installed_atsigns
+    get_installed_atsigns "device"
     device_atsign="$selectedatsign"
     get_device_atsign
   fi
--- a/packages/dart/sshnoports/bundles/universal.sh
+++ b/packages/dart/sshnoports/bundles/universal.sh
@@ -3,7 +3,7 @@
 # SCRIPT METADATA
 # DO NOT MODIFY/DELETE THIS BLOCK
 script_version="3.0.0"
-sshnp_version="5.1.0"
+sshnp_version="5.1.0-rc.5"
 repo_url="https://github.com/atsign-foundation/sshnoports"
 # END METADATA
 
--- a/packages/dart/sshnoports/bundles/universal.sh
+++ b/packages/dart/sshnoports/bundles/universal.sh
@@ -9,6 +9,11 @@ repo_url="https://github.com/atsign-foundation/sshnoports"
 
 # N.B. Other than the variable definitions, and the call to the main function,
 # nothing else should be writen outside the main function to avoid side effects
+### Environment config
+
+# GREP_COLOR not used directly so ignore the shellcheck warning for it
+# shellcheck disable=SC2034
+GREP_COLOR=never
 
 ### Environment based variables
 arg_zero="$0"
--- a/packages/dart/sshnoports/bundles/universal.sh
+++ b/packages/dart/sshnoports/bundles/universal.sh
@@ -448,34 +448,51 @@ get_device_atsign() {
 }
 
 get_installed_atsigns() {
-    atkeycount=0
-    atkeys=""
-    selectedatsign=""
-    if [ -d "${user_home}/.atsign/keys" ]; then
-        for file in $(find "$user_home"/.atsign/keys/*.atKeys -type f); do
-            atkeys="$atkeys $(echo "$file" | sed s/^.*@// | sed s/_key.atKeys//)"
-            atkeycount=$((atkeycount+1))
-        done
-        if [ $atkeycount -eq 0 ]; then
-            echo '$HOME/.atsign/keys directory found but there are no keys there yet'
-        else
-            echo "${atkeycount} atKeys found: ${atkeys}"
-            for file in $(find "$user_home"/.atsign/keys/*.atKeys -type f); do
-                atkey=$(echo "$file" | sed s/^.*@// | sed s/_key.atKeys//)
-                printf "Would you like to use @%s? " "$atkey"
-                read -r use_atkey
-                case $use_atkey in
-                [Yy]*)
-                    selectedatsign=$atkey
-                    break 2
-                    ;;
-                esac
-            done
-        fi
+  atkeycount=0
+  atkeys=""
+  selectedatsign=""
+  if [ -d "${user_home}/.atsign/keys" ]; then
+    # Disable unsafe find looping (will break if atkey file name contains a space, which it shouldn't)
+    # shellcheck disable=SC2044
+    for file in $(find "$user_home"/.atsign/keys/*.atKeys -type f); do
+      atkeys="$atkeys $(echo "$file" | sed s/^.*@// | sed s/_key.atKeys//)"
+      atkeycount=$((atkeycount + 1))
+    done
+    atkeys="$(echo "$atkeys" | sed s/\ \ /\ / | sed s/^\ //)" # remove double & leading space since it will interfere with cut
+    if [ $atkeycount -eq 0 ]; then
+      echo "$HOME/.atsign/keys directory found but there are no keys there yet"
+    elif [ $atkeycount -eq 1 ]; then
+      atkey=$(echo "$atkeys" | sed "s/\ //g")
+      echo "1 atKeys file found: ${atkey}"
+      printf "Would you like to use @%s? " "$atkey"
+      read -r use_atkey
+      case $use_atkey in
+        [Yy]*)
+          selectedatsign=$atkey
+          ;;
+      esac
     else
-        mkdir -p "$user_home"/.atsign/keys
-        echo '$HOME/.atsign/keys directory created'
+      echo "${atkeycount} atKeys found: ${atkeys}"
+      echo "Which one should be used:"
+      echo "0) None"
+      for i in $(seq 1 $atkeycount); do
+        echo "$i) @$(echo "$atkeys" | cut -d' ' -f"$i")"
+      done
+      printf '$ '
+      read -r selectedinput
+
+      selectedindex=$((selectedinput))
+      if [ "$selectedindex" -gt 0 ]; then
+        selectedatsign="$(echo "$atkeys" | cut -d' ' -f"$selectedindex")"
+        echo "Selected: @$selectedatsign"
+      else
+        echo "No existing atkeys were selected"
+      fi
     fi
+  else
+    mkdir -p "$user_home"/.atsign/keys
+    echo "$HOME/.atsign/keys directory created"
+  fi
 }
 
 suggest_sudo() {
--- a/packages/dart/sshnoports/bundles/universal.sh
+++ b/packages/dart/sshnoports/bundles/universal.sh
@@ -3,7 +3,7 @@
 # SCRIPT METADATA
 # DO NOT MODIFY/DELETE THIS BLOCK
 script_version="3.0.0"
-sshnp_version="5.1.0-rc.8"
+sshnp_version="5.1.0-rc.9"
 repo_url="https://github.com/atsign-foundation/sshnoports"
 # END METADATA
 
--- a/packages/dart/sshnoports/bundles/universal.sh
+++ b/packages/dart/sshnoports/bundles/universal.sh
@@ -45,570 +45,593 @@ device_name=""
 device_type=""
 
 norm_atsign() {
-	# Prepend an @ to the front of the atsign if missing
-	atsign="@$(echo "$1" | sed -e 's/"//g' -e 's/^@//g')"
-	echo "$atsign"
+    # Prepend an @ to the front of the atsign if missing
+    atsign="@$(echo "$1" | sed -e 's/"//g' -e 's/^@//g')"
+    echo "$atsign"
 }
 
 norm_version() {
-	# Ensure the version is in the format "tags/vX.Y.Z" (github version tag)
-	version="tags/v$(echo "$1" | sed -e 's/"//g' -e 's/^tags\///g' -e 's/^v//g')"
-	echo "$version"
+    # Ensure the version is in the format "tags/vX.Y.Z" (github version tag)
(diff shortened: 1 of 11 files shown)


gkc:
@XavierChanth 

Fails immediately for me on MacOS
```
gary@gkc2019-2 ~ % ./universal.sh -v
+ shift
+ '[' 0 -gt 0 ']'
+ '[' true = true ']'
+ print_env
+ echo Environment:
Environment:
+ echo 'Platform Name: macos'
Platform Name: macos
+ echo 'System Arch: arm64'
System Arch: arm64
+ echo 'Temporary Path: /tmp'
Temporary Path: /tmp
+ echo 'As Root: false'
As Root: false
+ echo 'Binary Path: /Users/gary/.local/bin'
Binary Path: /Users/gary/.local/bin
+ echo 'User: gary'
User: gary
+ '[' -n '' ']'
++ get_download_url
++ unset download_urls
+++ grep browser_download_url
++++ norm_version 5.1.0-rc.6
+++ cut '-d"' -f4
+++++ echo 5.1.0-rc.6
+++++ sed -e 's/"//g' -e 's/^tags\///g' -e 's/^v//g'
++++ version=tags/v5.1.0-rc.6
++++ echo tags/v5.1.0-rc.6
+++ curl -fsSL https://api.github.com/repos/atsign-foundation/noports/releases/tags/v5.1.0-rc.6
++ download_urls='https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-linux-arm.tgz
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-linux-arm64.tgz
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-linux-riscv64.tgz
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-linux-x64.tgz
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-macos-arm64.zip
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-macos-x64.zip
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-windows-x64.zip
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/universal.sh'
++ '[' -z 'https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-linux-arm.tgz
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-linux-arm64.tgz
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-linux-riscv64.tgz
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-linux-x64.tgz
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-macos-arm64.zip
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-macos-x64.zip
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-windows-x64.zip
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/universal.sh' ']'
++ echo 'https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-linux-arm.tgz
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-linux-arm64.tgz
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-linux-riscv64.tgz
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-linux-x64.tgz
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-macos-arm64.zip
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-macos-x64.zip
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-windows-x64.zip
https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/universal.sh'
++ grep macos
++ grep arm64
++ cut '-d"' -f4
+ download_url='https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-macos-arm64.zip'
+ echo 'https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-macos-arm64.zip'
+ download_archive
+ read -r download_url
+ echo 'Downloading archive from https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-macos-arm64.zip'
Downloading archive from https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-macos-arm64.zip
+ curl -sL 'https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-macos-arm64.zip' -o /tmp/sshnp-1711716526.zip
+ cleanup
+ rm -f /tmp/sshnp-1711716526.zip
+ rm -rf /tmp/sshnp-1711716526
gary@gkc2019-2 ~ % echo $?
3
```

curl exit code 3 is malformed URL which is weird because if I copy and paste that precise command as echoed above into a terminal it works fine
```
gary@gkc2019-2 ~ % curl -sL 'https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-macos-arm64.zip' -o /tmp/sshnp-1711716526.zip

gary@gkc2019-2 ~ % echo $?
0
```

I suspect that the single quotes are the problem here. So for example
```
gary@gkc2019-2 ~ % curl -sL "'https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-macos-arm64.zip'" -o /tmp/sshnp-1711716526.zip

gary@gkc2019-2 ~ % echo $?
3

```

gkc:
Testing on Ubuntu, client installation. Looks good except that the binaries aren't available where they should be

```
ubuntu@gkcrvd:~$ ./universal.sh
Downloading archive from https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-linux-x64.tgz
Install type (device, client, both): 
$ client
Enter client atSign:
$ alice
Enter device atSign:
$ bob
Pick your default region:
  am   : Americas
  ap   : Asia Pacific
  eu   : Europe
  @___ : Specify a custom region atSign
$ eu
Installing a quick picker script to make it easy to connect to devices...
Enter the device names you would like to include in the quick picker script
/done to finish
$ test_universal
$ foo
$ /done
ubuntu@gkcrvd:~$ cat .local/bin/@sshnp
#!/bin/bash
# SCRIPT METADATA
binary_path="$HOME/.local/bin"
client_atsign="@alice"
device_atsign="@bob"
host_atsign="@rv_eu"
devices=( test_universal foo)
additional_args=()
# END METADATA

unset d
echo "Select a device (enter the number):"
select d in "${devices[@]}"; do
	break
done

if [ -z "$d" ]; then
	echo "No device selected"
	exit 1
fi

"$binary_path"/sshnp -f "$client_atsign" -t "$device_atsign" -h "$host_atsign" -d "$d" "${additional_args[@]}" "$@"
ubuntu@gkcrvd:~$ .local/bin/@sshnp
Select a device (enter the number):
1) test_universal
2) foo
#? 2
.local/bin/@sshnp: line 22: /home/ubuntu/.local/bin/sshnp: No such file or directory
ubuntu@gkcrvd:~$ ls -trl .local/bin
total 4
-rwxr-xr-x 1 ubuntu ubuntu 463 Mar 29 13:03 @sshnp
ubuntu@gkcrvd:~$ 
```

gkc:
Testing on Ubuntu, device installation. Same problem as with client installation - while the binaries have been downloaded and unpacked, they haven't been copied to ~/.local/bin

```
ubuntu@gkcrvd:~$ ./universal.sh
Downloading archive from https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.6/sshnp-linux-x64.tgz
Install type (device, client, both): 
$ device
Enter client atSign:
$ alice
Enter device atSign:
$ bob
Enter device name:
$ test_universal
sed: can't read /home/ubuntu/.local/bin/sshnpd.sh: No such file or directory
ubuntu@gkcrvd:~$ 
```

packages/dart/sshnoports/bundles/universal.sh
```diff
@@ -79,12 +85,28 @@ usage() {
 	fi
```
cpswan: @XavierChanth should this be `universal.sh` rather than `install.sh`?

cpswan:
Testing a device install on Ubuntu. `install.sh` is being called, but doesn't seem to be doing anything.

e.g. this can be seen in verbose logging:

```
/tmp/sshnp-1712051112/sshnp/install.sh -b /home/ubuntu/.local/bin -u ubuntu tmux sshnpd
```

but ~/.local/bin remains empty

It's the same if I run:

```
/tmp/sshnp-1712051112/sshnp/install.sh -b /home/ubuntu/.local/bin sshnpd
```

I've already done `mkdir -p ~/.local/bin` to ensure the target directory is there (it isn't by default on a cloud-init VM).

cpswan:
Running just:

```
/tmp/sshnp-1712051112/sshnp/install.sh sshnpd
```

Works just fine :)

```
=> Installed sshnpd to /home/ubuntu/.local/bin
```

Looks like the problem is with option handing:

```
	case "$1" in
	--help | '')
		usage
		exit 0
		;;
	-b)
		binary_dir="$1"
		shift
		;;
```

1. This will set `binary_dir` to `-b` as that's what $1 is that we just tested for 🤦 
2. If we use $2 then at least `binary_dir` gets set correctly
3. But... we then pass through the rest of the `case` statement and exit with nothing happening, because the `shift` doesn't cause things to be re-evaluated.

packages/dart/sshnoports/bundles/universal.sh
xavierchanth: Good catch, yes it should

xavierchanth:
Thanks @cpswan, I'll do an rc.7 release
edit: rc.8

cpswan:
rc9 worked on my Ubuntu test VM for a device install :)

But... there's no mention of keys, so I think we should be testing for the presence of keys and alerting the user to put them in place.

Also the installer should say (at least a little bit) about what it's done (along similar lines to what install.sh says) e.g.:

```
tumx sshnp has been installed into /home/ubuntu/.local/bin
an entry has been added to the crontab which will take effect on reboot
run now with:
tmux new-session -d -s sshnpd && tmux send-keys -t sshnpd /home/ubuntu/.local/bin/sshnpd.sh C-m
you can connect to the daemon script with:
tmux attach -t sshnpd
and disconnect with:
Crtl-b d
```

cpswan:
rc9 failed on my AL2023 test VM:

```
/tmp/sshnp-1712076095/sshnp/install.sh: line 326: crontab: command not found
```

xavierchanth:
> rc9 failed on my AL2023 test VM:
> 
> ```
> /tmp/sshnp-1712076095/sshnp/install.sh: line 326: crontab: command not found
> ```

I literally can't trust anything to be consistently available across the industry anymore...

xavierchanth:
Nice... industry moving to systemd as the one size fits all solution, when some things can't/ don't run systemd... cron is the one thing that is usually consistently available to schedule tasks directly under init.d
CleanShot 2024-04-02 at 12 00 18

Nothing wrong with systemd, I think it's great, but especially in recent times, this is frustrating. (Also amazon linux just doesn't play nice in general, I have had to do some much extra work just to support them)

cpswan:
@XavierChanth I'm using AL2023 to test as it's probably the thorniest of ELs and so if we can get it working there, we can reasonably expect things to work in other EL flavours

If we're going to lean on systemd (which I think is wise given its prevalence in mainstream distros) then the installer needs to be nudging folk in that direction.

Where there's no systemd (or no sudo to install to it) we need to be checking that other tools are available and suggesting how to install them. This sadly descends into a rats nest of conditionals pretty quickly :(

We're somewhat back to the tribulations of the dess installer :/

More fun... no tmux either in default AL2023

cpswan:
Just tested a systemd install on my AL2023 VM. It's not working, and if I look in /etc/systemd/system/sshnpd.service the User= field hasn't been replaced:

```
# MANDATORY: User to run the daemon as
User=<username>
```

If I manually update it things work OK. So need to patch up the script to do the replacement.

xavierchanth:
> More fun... no tmux either in default AL2023

I never expect tmux to be available by default. but I do expect cron and nohup, which is what our headless one depends on. Guess we will need to check for that... thankfully AL2023 has systemd which means we should use that. if you run universal.sh as root (i.e. with sudo) it should automatically use systemd... maybe we should check if systemd is installed and nudge them down the systemd route with a few y/n prompts?

To be clear, I also only kept the headless version around because it was the one true fallback that would just work on any system. I actually think it is a terrible way to install sshnpd, so if it can't be the one true fallback then I think we should get rid of it to simplify things.

cpswan:
@XavierChanth I think this is ready to be merged. I've done a bunch more testing on both Ubuntu and AL2023 for both client and device installs using the guided prompts and command line options.

xavierchanth:
> @XavierChanth I think this is ready to be merged. I've done a bunch more testing on both Ubuntu and AL2023 for both client and device installs using the guided prompts and command line options.

I wasn't able to reproduce @gkc 's install issues, let me confirm with him first.

gkc:
We figured out what the problems were (1) my grep options which included --color=always and (2) missing function mac_only

packages/dart/sshnoports/bundles/universal.sh
```diff
@@ -35,435 +42,667 @@ device_atsign=""
...
+        echo "Selected: @$selectedatsign"
+      else
+        echo "No existing atkeys were selected"
+      fi
+    fi
+  else
+    mkdir -p "$user_home"/.atsign/keys
+    echo "$HOME/.atsign/keys directory created"
+  fi
+}
```
xavierchanth: I modified this function so that it only asks y/n if a single key was found.
Otherwise it presents a numerical options list to choose from with 0/empty being don't select anything, and numbers 1+ being the various atkeys to choose from

xavierchanth:
@gkc can I get one last sanity test against: https://github.com/atsign-foundation/noports/releases/download/v5.1.0-rc.11/universal.sh

I will prepare the v5.1.0 release

xavierchanth:
https://github.com/atsign-foundation/noports/releases/tag/v5.1.0

xavierchanth:
I cheated and am rebuilding universal.sh for v5.1.0 on commit https://github.com/atsign-foundation/noports/pull/913/commits/d3498627d9d07998ee5bb5c1bf7b6460a7e68c36 (also installs at_activate now)

The rest of v5.1.0 is built on https://github.com/atsign-foundation/noports/pull/913/commits/4cac6783d8b27b9aeae99ba6ebce9d0cd92db441

cpswan (dismissed):

gkc (approved):

cpswan (approved):

xavierchanth:
Okay, I will move tag v5.1.0 to https://github.com/atsign-foundation/noports/pull/913/commits/546c59dea5218cf3b4d427d3c9422f2049431735 and rebuild the release.

xavierchanth:
There is a straggler dockerhub build, but that should be unchanged, since we haven't changed anything other than shell scripts since it's last v5.1.0 tagged build

PR #940, opened 2024-04-11
fix: 902 noports desktop enhancements v4

### LINKED ISSUE #902: NoPorts Desktop Enhancements V4.0

All changes are made - I have a couple new notes:

- [x] I think font size might be too small in some areas (might just be me)
- [x] Could we get these two button menus to have the same layout (whichevers easiest for you)
> Image
> Image

- [x]  This was intended for when we had more than one option
 >  Home Screen - can we do the following
- Can we do something like this
Image

- [x] Replace app assets with the new noports logo
- [x] Use new logo as launcher icon

.

_No response_

### Additional context

_No response_

CurtlyCritchlow:
**- What I did**

- Updated the app so the font size and some widgets size will increase up to a maximum point based the device size.
- Updates app assets with the new no ports logo
- Update app launcher icon on macOS

closes #902

**- How I did it**

**- How to verify it**
Run the app and increase the windows size, you should observe the app font size increasing up to a certain point.

App Scales based on device Size
App uses new no ports logos


xavierchanth (approved):
Looks good!

PR #941, opened 2024-04-11
fix: sudo suggest

xavierchanth:
**- What I did**

- Added a few guard rails to make it clearer
- Fixed some flows where device_type_input could be left unset

**- How I did it**

**- How to verify it**

fix: sudo suggest

### DIFF
--- a/packages/dart/sshnoports/bundles/universal.sh
+++ b/packages/dart/sshnoports/bundles/universal.sh
@@ -36,6 +36,7 @@ unset tmp_path
 install_type=""
 unset download_url
 local_archive=""
+no_sudo=false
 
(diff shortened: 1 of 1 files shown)


cconstab (approved):

cpswan:
@XavierChanth too late for this one, but you might want to check over the logic map for command availability. E.g. can a tmux install succeed without cron? Should we check for `cron` or `crontab`?

xavierchanth:
> @XavierChanth too late for this one, but you might want to check over the logic map for command availability. E.g. can a tmux install succeed without cron? Should we check for `cron` or `crontab`?

We should also check there, I was focused on making sure that our fallback option is covered.

PR #945, opened 2024-04-12
chore: bump version v5.1.1-rc.1

xavierchanth:
**- What I did**

@cpswan not sure why we are protecting `release-` branches now... all noports releases are tied to tags.

There are always last minute things that need to be added to before we release, hence the workflow has been: create release- branch, push changes to that branch, tag it, and then release.

Unit tests are also not running, yet required...
CleanShot 2024-04-12 at 11 48 32

**- How I did it**

**- How to verify it**

chore: bump version v5.1.1-rc.1

### DIFF
--- a/packages/dart/sshnoports/bundles/universal.sh
+++ b/packages/dart/sshnoports/bundles/universal.sh
@@ -3,7 +3,7 @@
 # SCRIPT METADATA
 # DO NOT MODIFY/DELETE THIS BLOCK
 script_version="3.0.0"
-sshnp_version="5.1.0"
+sshnp_version="5.1.1.rc.1"
 repo_url="https://github.com/atsign-foundation/sshnoports"
 # END METADATA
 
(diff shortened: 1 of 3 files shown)


Xlin123 (approved):

cpswan:
@XavierChanth those branches are protected because otherwise we get a ding on the scorecard (quite rightly, because you're suggesting using unprotected branches as a way to circumvent review).

The whole point is we shouldn't be having flaps around releases. We need to figure out how to keep things orderly. Perhaps we should get a testing path through the process?

cpswan (approved):

xavierchanth:
> @XavierChanth those branches are protected because otherwise we get a ding on the scorecard (quite rightly, because you're suggesting using unprotected branches as a way to circumvent review).
> 
> The whole point is we shouldn't be having flaps around releases. We need to figure out how to keep things orderly. Perhaps we should get a testing path through the process?

Makes sense. the main reason for cutting these branches is to be able to pre-release without needing to commit the pre-release version tags to trunk.

I guess we could add a build step which automatically updates the tag to match release-(v*.*.*) or refs/tag/(v*.*.*)

PR #961, opened 2024-04-18
feat: Include allowed services in ping

xavierchanth:
**- What I did**

- Added 'allowedServices' as a `List<String>` to the device info map which is published as a part of the ping / list devices, this maps directly to `--permit-open`.
- Updated the output of `sshnp --list-devices` to include availableServices on a line below it's associated device name & version.

**- How I did it**

**- How to verify it**

Output from tests:

```
Searching for devices...
Active Devices:
  charon - v5.1.0 (core v6.0.6)
  hermes - v5.1.0 (core v6.0.6)
  test1 - v5.1.0 (core v6.0.6)
  - allowedServices: localhost:22 localhost:3389
  test2 - v5.1.0 (core v6.0.6)
  - allowedServices: localhost:22 localhost:3389
  test3 - v5.1.0 (core v6.0.6)
  - allowedServices: localhost:22 localhost:3389
  test4 - v5.1.0 (core v6.0.6)
  - allowedServices: some.thing.different.com:7878
Inactive Devices:
  asdf - v5.0.2 (core v6.0.2)
  foo1 - v5.1.0-rc.11 (core v6.0.6)
  foo2 - v5.1.0-rc.11 (core v6.0.6)
  hermeslaunchd - v5.0.2 (core v6.0.2)
  mar28 - v5.0.2 (core v6.0.2)
  mar282 - v5.0.2 (core v6.0.2)
  mar2828 - v5.0.2 (core v6.0.2)
```

Daemon commands:
```
parallel -j 3 -i sh -c "echo starting {}; sshnpd -a @xchan -m @xavierchanth -d {} --storage-path $HOME/.sshnp/@xchan-{}/storage -u" -- test1 test2 test3
```

```
sshnpd -a @xchan -m @xavierchanth -d test4 --storage-path $HOME/.sshnp/@xchan-test
4/storage -u
```

Client command:
```
dart run packages/dart/sshnoports/bin/sshnp.dart -f @xavierchanth -t @xchan -h @xchan_8286 --list-devices
```

feat: Include allowed services in ping

### DIFF
--- a/packages/dart/sshnoports/bin/sshnp.dart
+++ b/packages/dart/sshnoports/bin/sshnp.dart
@@ -168,7 +168,7 @@ void main(List<String> args) async {
       // Run List Devices Operation
       if (params.listDevices) {
         stderr.writeln('Searching for devices...');
-        var deviceList = await sshnp.listDevices();
+        SshnpDeviceList deviceList = await sshnp.listDevices();
         printDevices(deviceList);
         exitProgram();
       }
--- a/packages/dart/sshnoports/lib/src/print_devices.dart
+++ b/packages/dart/sshnoports/lib/src/print_devices.dart
@@ -22,12 +22,16 @@ void printDeviceList(Iterable<String> devices, Map<String, dynamic> info) {
   for (var device in devices) {
     stderr.writeln('  $device - v${info[device]?['version']}'
         ' (core v${info[device]?['corePackageVersion']})');
+
     if (info[device]['allowedServices'] != null &&
-        (info[device]['allowedServices'] is List<String>) &&
-        (info[device]['allowedServices'] as List<String>).isNotEmpty) {
+        (info[device]['allowedServices'] is List) &&
+        (info[device]['allowedServices'] as List).isNotEmpty) {
+      // allowedServices should be a List<String> but casting from json means it's actually a List<dynamic>
+      // It's an unnecessary pain to bother going through casting hell for,
+      // since all json types should have a .toString() which is reasonable to print as output
       stderr.write("  - allowedServices:");
       for (String service in info[device]['allowedServices']) {
-        stderr.write(' $service');
+        stderr.write(' ${service.toString()}');
       }
       stderr.writeln();
     }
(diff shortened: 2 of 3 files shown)


packages/dart/sshnoports/bin/sshnp.dart
```diff
@@ -168,7 +168,7 @@ void main(List<String> args) async {
       // Run List Devices Operation
       if (params.listDevices) {
         stderr.writeln('Searching for devices...');
-        var deviceList = await sshnp.listDevices();
+        SshnpDeviceList deviceList = await sshnp.listDevices();
```
xavierchanth, line 171: Made this type explicit, because I wrote this code, and I was sure it was a custom object, but the name I chose suggested otherwise... oops...

packages/dart/sshnoports/lib/src/print_devices.dart
```diff
@@ -26,5 +22,14 @@ void printDeviceList(Iterable<String> devices, Map<String, dynamic> info) {
...
         ' (core v${info[device]?['corePackageVersion']})');
+    if (info[device]['allowedServices'] != null &&
+        (info[device]['allowedServices'] is List<String>) &&
+        (info[device]['allowedServices'] as List<String>).isNotEmpty) {
+      stderr.write("  - allowedServices:");
+      for (String service in info[device]['allowedServices']) {
+        stderr.write(' $service');
+      }
+      stderr.writeln();
+    }
```
xavierchanth, line 37: The format should be something like
```
Available devices:
  mydevice - v5.1.0 (Core v6.x.x)
  - allowedServices: localhost:22 localhost:3389
```

xavierchanth: Should be easy to parse, if the first non-whitespace character on the following line is a `-`, then this is an attribute associated with the device. `-` is not allowed in the device name, so this should be an acceptable approach.

I could change the `:` after 'allowedServices' since it conflicts with the `:` in the host:port scheme, but this can still even be parsed in the shell with cut, awk or sed.

gkc (dismissed):
LGTM

xavierchanth:
Will need another review @gkc , the tests listed in the initial PR body are tested against commit https://github.com/atsign-foundation/noports/pull/961/commits/90be843f6a4088e3e8363ebe3af7788eec7938dc

gkc (approved):

PR #972, opened 2024-04-23
ci: Release automation

### LINKED ISSUE #948: Improve release workflows & automation
```[tasklist]
### Tasks
- [x] Try to make it so that multibuild.yaml automatically uploads to the releases tab
- [ ] multibuild.yaml outputs a hash for releases which we can include in a table as part of the release notes
- [ ] Releases/builds created from tags should automatically update the version in pubspec.yaml & version.dart to match the git tag & then build [multibuild & docker images]
```

cpswan:
Fixes #948

**- What I did**

Refactored multibuild.yaml after much experimentation with https://github.com/cpswan/release_automation

**- How I did it**

* Create a temporary branch which is used to carry version changes from a release tag
  * That branch is deleted by a cleanup job
* other_build workflow reworked to make steps conditional rather than the whole workflow
* github-release job added that downloads artifacts and signs them with Sigstore

**- How to verify it**

Workflow can be run manually as before (with option to skip Docker builds).

When a (pre) release is tagged the workflow will add artifacts and signatures to it.

ci: Release automation


xavierchanth (approved):
Yay!

PR #973, opened 2024-04-23
chore: deprecate -h, only -f needed for list-devices

xavierchanth:
**- What I did**

- deprecated -h, --host (also hide both of these)
- made it so -t, and -h are not needed to --list-devices
- with -t only list-devices for that daemon, without -t, list for all devices.

**- How I did it**

**- How to verify it**

- Tested -h and --host for breaking changes, still negotiates with rv correctly.
- List devices works with -t, only listing devices for -t
- List devices work without -t, listing devices as devicename@deviceatsign
  - This approach maintains compatibility with the notes from #968 

chore: deprecate -h, only -f needed for list-devices


packages/dart/noports_core/lib/src/sshnp/models/sshnp_device_list.dart
```diff
@@ -12,4 +12,9 @@ class SshnpDeviceList {
 
   Set<String> get inactiveDevices =>
       info.keys.toSet().difference(activeDevices);
+
+  void add(SshnpDeviceList o) {
+    info.addEntries(o.info.entries);
+    activeDevices.addAll(o.activeDevices);
+  }
```
xavierchanth, line 19: Helper function to join the entries of another SshnpDeviceList into this one

packages/dart/noports_core/lib/src/sshnp/models/sshnp_params.dart
```diff
@@ -262,16 +263,23 @@ class SshnpParams extends ClientParamsBase
...
-        (throw ArgumentError('to (npdAtSign) is mandatory'));
-    partial.srvdAtSign ?? (throw ArgumentError('srvdAtSign is mandatory'));
+
+    if (!(partial.listDevices ?? DefaultSshnpArgs.listDevices)) {
+      // if list-devices is not set, then ensure sshnpdAtSign and srvdAtSign are set
+      partial.sshnpdAtSign ??
+          (throw ArgumentError(
+              'Option to is mandatory, unless list-devices is passed.'));
+      partial.srvdAtSign ?? (throw ArgumentError('srvdAtSign is mandatory'));
+    }
```
xavierchanth, line 278: Only throw argument errors if list-devices is also false.

packages/dart/noports_core/lib/src/sshnp/models/sshnp_params.dart
```diff
@@ -262,16 +263,25 @@ class SshnpParams extends ClientParamsBase
...
+              'srvdAtSign is mandatory, unless list-devices is passed.'));
+    }
+
     return SshnpParams(
       profileName: partial.profileName,
       clientAtSign: partial.clientAtSign!,
-      sshnpdAtSign: partial.sshnpdAtSign!,
-      srvdAtSign: partial.srvdAtSign!,
+      sshnpdAtSign: partial.sshnpdAtSign ?? "",
+      srvdAtSign: partial.srvdAtSign ?? "",
```
xavierchanth, line 284: Easier to pass these as empty strings than to refactor everything to null check.

packages/dart/noports_core/lib/src/sshnp/models/sshnp_params.dart
```diff
@@ -479,7 +489,8 @@ class SshnpPartialParams {
       sshnpdAtSign: args[SshnpArg.toArg.name] == null
           ? null
           : AtUtils.fixAtSign(args[SshnpArg.toArg.name]),
-      srvdAtSign: args[SshnpArg.srvdArg.name],
+      srvdAtSign:
+          args[SshnpArg.srvdArg.name] ?? args[SshnpArg.legacySrvdArg.name],
```
xavierchanth, line 493: map -h and --host as a fallback to -r and --srvd

packages/dart/noports_core/lib/src/sshnp/util/sshnpd_channel/sshnpd_channel.dart
```diff
@@ -255,39 +255,86 @@ abstract class SshnpdChannel with AsyncInitialization, AtClientBindings {
...
+    for (var list in deviceLists) {
+      consolidatedList.add(list);
+    }
 
-    var atKeys =
-        await getAtKeysRemote(regex: scanRegex, sharedBy: params.sshnpdAtSign);
+    return consolidatedList;
+  }
 
+  Future<SshnpDeviceList> _listDevices(String sharedBy,
```
xavierchanth, line 290: This is the old logic for listDevices

packages/dart/noports_core/lib/src/sshnp/util/sshnpd_channel/sshnpd_channel.dart
```diff
@@ -255,39 +255,86 @@ abstract class SshnpdChannel with AsyncInitialization, AtClientBindings {
   /// and corresponding info, and a list of active devices (devices which also
   /// responded to our real-time ping).
   Future<SshnpDeviceList> listDevices() async {
```
xavierchanth, line 257: This function wraps the previous logic, deferring to the old logic if -t is set, or aggregating across all devices if -t is not set.

packages/dart/noports_core/lib/src/sshnp/util/sshnpd_channel/sshnpd_channel.dart
```diff
@@ -325,7 +372,8 @@ abstract class SshnpdChannel with AsyncInitialization, AtClientBindings {
       bool showHiddenKeys = false}) async {
     var builder = ScanVerbBuilder()
       ..sharedWith = sharedWith
-      ..sharedBy = sharedBy
+      ..sharedBy =
+          sharedBy // for some reason, if this is null it only returns cached keys
```
xavierchanth, line 376: Noted a behavioral bug, which explains why I went an round-about way to implement listDevices without -t

cpswan (commented):
LGTM, though can there be a new test for this path through the usage?

xavierchanth:
> LGTM, though can there be a new test for this path through the usage?

Sure, I will add it today

xavierchanth:
#### Additional changes
1. `minus_r_flag` e2e_all test:
- base `sshnp` command still uses `-h` otherwise we have to do a bunch of refactoring of the tests to change the base command based on the client version
- Test 1 : `--host` regression test
- Test 2 : `-r` overrides `-h` value (made `-h` value a non-activated atSign, so it will fail if the `-h` value is used)

2. Updated noports core to 6.0.7 and updated changelog
3. Updated sshnoports to 5.2.0 (since I need a cut-off version for when this test is allowed to run)

tests/e2e_all/scripts/tests/minus_r_flag
```diff
@@ -0,0 +1,125 @@
+#!/bin/bash
+
+# 1. Run with --host
+#  - expect this test to pass
+# 2. Run with -h && -r (-h set to an invalid rvd atSign)
+#  - expect this test to pass
+
+# TODO: Would like to add a test for, but this causes the test to timeout waiting for the bad srvd
+# 3. Run with -h && -r (-r set to an invalid rvd atSign)
+#  - expect this test to fail
```
xavierchanth, line 10: It would be nice to implement this test, but test case 2 asserts the inverse case is validated.

cpswan (approved):

PR #1047, opened 2024-05-09
fix: sshnp: various arg parsing issues

### LINKED ISSUE #1046: sshnp: arg parsing issues
### Describe the bug

1. Integer arg parsing does not fail when a non-integer parameter is supplied. As a result, some command-line arg problems are not detected and can cause end-user confusion
2. We expect there to be nothing in the 'rest' of the parsedArgs because everything is captured by options or flags, but we do not assert that this is the case
3. Device name is not mandatory and defaults to `default`; this may cause some confusion

### Steps to reproduce

For example:
`bin/sshnp --from @alice --to @bob -r @chuck -l --device my_device -u root -v --no-ad --no-et --root-domain vip.ve.atsign.zone` currently results in
1. SshnpPartialParams.fromArgList trying to parse a local port from `--device` and evaluating to null because we are using tryParse rather than parse
2. argParser.rest (what's left over after parsing) containing "my_device"
3. device being "default" because the intended `--device` option was swallowed as an argument to the `-l` option

### Expected behavior

1. Integer parsing should fail if a non-parseable value is found
2. `--device` should be mandatory and not have a default value
3. we should assert that argParser.rest is empty

### Screenshots

_No response_

### Smartphones

_No response_

_No response_

### Additional context

_No response_

gkc:
**- What I did**
Fixes #1046 

**- How I did it**
- make parsing of integer options fail if a non-parseable value is supplied
- if `--device` is not supplied, print a warning message stating that, and mentioning that it will be mandatory in future 
- assert that `argParser.rest` is empty after parsing, since we are capturing all parameters via options or flags
- test: fix a test which had extraneous arguments in its command-line and was now failing due to the `argParser.rest` assertion we added

**- How to verify it**
Tests pass


cpswan (dismissed):
If we're not changing `--device` to mandatory then LGTM.

gkc:
> If we're not changing `--device` to mandatory then LGTM.

This PR does change it to mandatory see sshnp_arg.dart file diff

xavierchanth (changes requested):
We cannot make device mandatory, it will break the ability to use list-devices, it instead should be validated similar to how --from and --to are.


```dart
  // models/sshnp_params.dart:274

  factory SshnpParams.fromPartial(SshnpPartialParams partial) {
    // Always need the clientAtSign
    partial.clientAtSign ??
        (throw ArgumentError('from (clientAtSign) is mandatory'));

    if (!(partial.listDevices ?? DefaultSshnpArgs.listDevices)) {
      // if list-devices is not set, then ensure sshnpdAtSign and srvdAtSign are set
      partial.sshnpdAtSign ??
          (throw ArgumentError(
              'Option to is mandatory, unless list-devices is passed.'));
      partial.srvdAtSign ??
          (throw ArgumentError(
              'srvdAtSign is mandatory, unless list-devices is passed.'));
    }

    return SshnpParams(
      profileName: partial.profileName,
      clientAtSign: partial.clientAtSign!,
      sshnpdAtSign: partial.sshnpdAtSign ?? "",
      srvdAtSign: partial.srvdAtSign ?? "",
      device: partial.device ?? DefaultSshnpArgs.device,
      ...
);
```

gkc:
> We cannot make device mandatory, it will break the ability to use list-devices, it instead should be validated similar to how --from and --to are.
> 
> ```dart
>   // models/sshnp_params.dart:274
> 
>   factory SshnpParams.fromPartial(SshnpPartialParams partial) {
>     // Always need the clientAtSign
>     partial.clientAtSign ??
>         (throw ArgumentError('from (clientAtSign) is mandatory'));
> 
>     if (!(partial.listDevices ?? DefaultSshnpArgs.listDevices)) {
>       // if list-devices is not set, then ensure sshnpdAtSign and srvdAtSign are set
>       partial.sshnpdAtSign ??
>           (throw ArgumentError(
>               'Option to is mandatory, unless list-devices is passed.'));
>       partial.srvdAtSign ??
>           (throw ArgumentError(
>               'srvdAtSign is mandatory, unless list-devices is passed.'));
>     }
> 
>     return SshnpParams(
>       profileName: partial.profileName,
>       clientAtSign: partial.clientAtSign!,
>       sshnpdAtSign: partial.sshnpdAtSign ?? "",
>       srvdAtSign: partial.srvdAtSign ?? "",
>       device: partial.device ?? DefaultSshnpArgs.device,
>       ...
> );
> ```

@XavierChanth  ack

gkc:
@XavierChanth as per your comment on #1046 I've reverted the change to make `--device` mandatory and instead am outputting a warning from the sshnp CLI if device is "default"

xavierchanth:
What I should have probably done from the start is make list-devices a subcommand instead of a flag. We can change it in v6, which will then allow us to enforce the arguments (I think... args is not the most flexible package)

gkc:
@XavierChanth There's a flurry of PRs right now so I'll need to rerun the e2e_all check before I merge, but those tests are working fine

xavierchanth (approved):

PR #1077, opened 2024-05-24
feat: add --quiet flag to the sshnp and npt CLIs

gkc:
**- What I did**
- feat: add --quiet flag to the sshnp and npt CLIs

Other stuff, not directly related to the feature
- build: updated pubspec.lock files via dart run melos bootstrap
- fix: Modify buildBinaries script so it produces a binary called at_activate, as our actual build processes do. And add it to .gitignore
- feat: In order to make local testing of the installer easier, add a buildArchive script which creates a zip or tgz in the same way as our gh actions do for releases. The script creates a build/ directory in the packages/dart/sshnoports directory; add that build/ directory to .gitignore

**- How I did it**
See commits

**- How to verify it**
Tests pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ pubspec_overrides.yaml
 
 # Binaries build from dart/sshnoports
 packages/dart/sshnoports/bin/activate_cli
+packages/dart/sshnoports/bin/at_activate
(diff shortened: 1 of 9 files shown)


xavierchanth (approved):

cconstab (approved):
Yay

PR #1080, opened 2024-05-27
feat: make daemon storage location map to its device name

### LINKED ISSUE #1038: Default daemon storage location should map to device name
So you can run multiple devices without needing to specify storage path.

Not a high priority, but maybe:

`/default-storage-path/devicename`

@gkc thoughts?

gkc:
**- What I did**
- feat: make daemon storage location map to its device name
- Closes #1038 

**- How I did it**
Modified a bunch of code related to generating a canonical AtClient storage path / creating an AtClient using that storage path

**- How to verify it**
Tests pass


gkc:
@XavierChanth please review when you get a minute

xavierchanth (approved):

PR #1090, opened 2024-05-31
feat: No double ssh when sockets encrypted

### LINKED ISSUE #990: Remove / disable the tunnel layer from sshnp
@gkc would like your thoughts on this.

Needed for #954

gkc:
**- What I did**
Closes #990 

**- How I did it**
See commits

**- How to verify it**
Tests pass

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -50,7 +50,7 @@ jobs:
 
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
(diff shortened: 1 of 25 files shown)


xavierchanth:
@gkc pushed a minor change, the C daemon will return null for the ephemeral ssh key in the payload, so I made sure to only assert that the ephemeral key is not null when encryptTraffic = false.

gkc:
@XavierChanth Shall we merge this to trunk?

xavierchanth:
> @XavierChanth Shall we merge this to trunk?

Yes, I think we can, I've not encountered any further issues.

xavierchanth (approved):

cconstab (approved):

PR #1112, opened 2024-06-11
test: test against trunk version of at_client as of 240611

gkc:
**- What I did**
test: test against trunk version of at_client as of 240611

**- How I did it**
dependency override for at_client

**- How to verify it**
tests pass


xavierchanth:
Can this be closed @gkc

gkc:
Yep thanks for the nudge

PR #1120, opened 2024-06-13
feat: [c daemon] verify_envelope_signature

realvarx:
**- What I did**
- verification of sshnp-to-sshnpd request notification payloads

**- How I did it**
- `verify_envelope_signature` function

**- How to verify it**
Run sshnpd normally with verbose flag. A new `DEBG` message should indicate wether the signature verification was successful.

image

If the original signature is modified before it is verified:
```
value[100] = 'A';
```
We will observe how the verification fails:

image


xavierchanth (dismissed):
Thanks! I will likely move this to another file, since I need this for npt_request too.

realvarx:
> Thanks! I will likely move this to another file, since I need this for npt_request too.

So would you like to merge this branch, or create a different one with these commits and other new features?

xavierchanth (approved):

PR #1135, opened 2024-06-17
fix: C SSHNPD memory leaks before SSH connection established

JeremyTubongbanua:
**- What I did**
- Pinned atsdk commit which is from https://github.com/atsign-foundation/at_c/pull/325
- Removed atlogger_free() since that is not the most recent feature

**- How to verify it**

valgrind leak summary after starting daemon, pkam authenticating, starting main loop, then Ctrl-C

```
==194277== LEAK SUMMARY:
==194277==    definitely lost: 0 bytes in 0 blocks
==194277==    indirectly lost: 0 bytes in 0 blocks
==194277==      possibly lost: 0 bytes in 0 blocks
==194277==    still reachable: 0 bytes in 0 blocks
==194277==         suppressed: 223 bytes in 1 blocks
```

### DIFF
--- a/packages/c/sshnpd/sshnpd.supp
+++ b/packages/c/sshnpd/sshnpd.supp
@@ -1,4 +1,6 @@
 # key allocation & free is a known false positive
+# TODO: I couldn't see anything wrong with how we are doing it, but need to get 
+# a second set of eyes on this issue - @XavierChanth
 {
   atchops_rsakey_privatekey_init
     Memcheck:Leak
--- a/packages/c/sshnpd/sshnpd.supp
+++ b/packages/c/sshnpd/sshnpd.supp
@@ -1,0 +1,148 @@
+# key allocation & free is a known false positive
+{
+  atchops_rsakey_privatekey_init
+    Memcheck:Leak
+    match-leak-kinds: definite
+    fun:malloc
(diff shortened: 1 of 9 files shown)


JeremyTubongbanua:
As of most recent commit of sending this message,

C SSHNPD works with at_c commit hash: 977cf8f51bb93e494b14fedfe3a577f440e45f0e

JeremyTubongbanua:
Will wait for https://github.com/atsign-foundation/at_c/pull/325 to be pushed to trunk before we push this to trunk

xavierchanth (approved):

packages/c/sshnpd/sshnpd.supp
```diff
@@ -0,0 +1,442 @@
...
+    fun:print
+    fun:cJSON_PrintUnformatted
+    fun:main
+}
+{
+  FIXME atclient_atsign_init
+    Memcheck:Leak
+    fun:malloc
+    fun:atclient_atsign_init
+}
```
xavierchanth, line 16: These also need to be unsuppressed & fixed to get rid of all pre-connection leaks

JeremyTubongbanua:
Will do 1 more pass before I merge @XavierChanth

PR #1149, opened 2024-06-20
chore: set default parameters in sshnpd _notify

realvarx:
**- What I did**
- All calls to _notify in `sshnpd_impl.dart` were taking the same values for some of its parameters. So I removed the required keyword and set those values as default.

**- How I did it**
https://github.com/atsign-foundation/noports/blob/33af62b7fc8b44a1bfb5352a711134cc92403a4a/packages/dart/noports_core/lib/src/sshnpd/sshnpd_impl.dart#L1131-L1133

**- How to verify it**


xavierchanth (approved):
Looks good as is, I know you mentioned another change to make, ping me if you need another review

PR #1153, opened 2024-06-20
fix: ensure directories we create as root are chown-ered to user

xavierchanth:
**- What I did**

Still encountering directory ownership issues with the installer, so I implemented a cleanup step which will call chown if the directory didn't exist prior to installation.

Also through a step in to print_env which will tell us the ssh status, useful for future debugging.

**- How I did it**

**- How to verify it**

fix: ensure directories we create as root are chown-ered to user

### DIFF
--- a/packages/dart/sshnoports/bundles/universal.sh
+++ b/packages/dart/sshnoports/bundles/universal.sh
@@ -112,8 +112,10 @@ sedi() {
 }
 
 chown_dir() {
-  echo "$1 was created by this installer, ensuring that it is owned by $user"
-  chown -R $user:$user "$1" 2>/dev/null || chown -R $user "$1" 2>/dev/null
+  if [ -d $1 ]; then
+    echo "$1 was created by this installer, ensuring that it is owned by $user"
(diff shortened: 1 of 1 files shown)


packages/dart/sshnoports/bundles/universal.sh
```diff
@@ -804,6 +843,22 @@ done
   fi
 }
 
+chown_dir() {
+  echo "$1 was created by this installer, ensuring that it is owned by $user"
+  chown -R $user:$user "$1" || chown -R $user "$1"
+}
+
+cleanup() {
+  if as_root; then
```
xavierchanth: Only run this step if we were running as root,

cpswan (dismissed):

xavierchanth:
Doing a sanity test before I merge

xavierchanth:
Okay so here are the cases mapped to behaviour during `cleanup`:

- dir didn't exist, but was created   : calls `chown_dir` which gives ownership to the user.
- dir didn't exist, and still doesn't : calls `chown_dir` which silently does nothing.
- dir already existed                 : doesn't call `chown_dir` which obviously does nothing.

Xlin123 (approved):
lgtm

PR #1199, opened 2024-07-15
fix: (i) fix race condition in daemon-side multi srv; (ii) add `\n` separator to npt control messages

gkc:
**- What I did**
- fix: fix race condition occurring in the daemon-side multi srv when connections are being made within the same very short time period (microS to low milliS)
- fix: append `\n` to the connect message sent on the npt control socket

**- How I did it**
1. wrap the control event handling in a mutex
2. handle situation where multiple connect requests are arriving in the same tcp/ip flow
3. fix: append \n to the connect message sent on the npt control socket

**- How to verify it**
(i) e2e tests pass
(ii) manually verified backwards compatibility with the `\n` separator by running multiple ssh sessions over an npt tunnel to port 22

### DIFF
--- a/packages/dart/noports_core/lib/src/srv/srv_impl.dart
+++ b/packages/dart/noports_core/lib/src/srv/srv_impl.dart
@@ -538,10 +538,6 @@ class SrvImplDart implements Srv<SocketConnector> {
       multi: multi,
       timeout: timeout,
       beforeJoining: (Side sideA, Side sideB) {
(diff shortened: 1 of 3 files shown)


gkc:
@XavierChanth note that this resolves only the race condition in the srv code. There will be follow-up PRs which address other aspects of the issues which occur when client machines are creating many socket connections within the same few milliseconds

@realvarx please ensure the same message handling is in place in the C daemon

xavierchanth (approved):

PR #1222, opened 2024-07-25
fix: use wrapped chown method in all places

xavierchanth:
**- What I did**

Used the `chown_dir` wrapper function instead of calling `chown` directly.

**- How I did it**

**- How to verify it**

fix: use wrapped chown method in all places

### DIFF
--- a/packages/dart/sshnoports/bundles/universal.sh
+++ b/packages/dart/sshnoports/bundles/universal.sh
@@ -617,7 +617,7 @@ get_atsign() {
     fi
   else
     mkdir -p "$user_home"/.atsign/keys
-    chown -R $user:$user "$user_home"/.atsign
+    chown_dir "$user_home"/.atsign
     echo "$HOME/.atsign/keys directory created"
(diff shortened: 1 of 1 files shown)


packages/dart/sshnoports/bundles/universal.sh
```diff
@@ -617,7 +617,7 @@ get_atsign() {
     fi
   else
     mkdir -p "$user_home"/.atsign/keys
-    chown -R $user:$user "$user_home"/.atsign
+    chown_dir "$user_home"/.atsign
```
xavierchanth, line 620: The function is on line 114:
```sh
chown_dir() {
  if [ -d $1 ]; then
    echo "$1 was created by this installer, ensuring that it is owned by $user"
    chown -R $user:$user "$1" 2>/dev/null || chown -R $user "$1" 2>/dev/null
  fi
}
```

cpswan (approved):

JeremyTubongbanua (approved):

PR #1233, opened 2024-07-31
chore: add multi flag to srv

xavierchanth:
**- What I did**

- `--multi` was not being set explicitly, so on occasion, it would be allocated as true even if it shouldn't of been.

**- How I did it**

**- How to verify it**

chore: add multi flag to srv

### DIFF
--- a/packages/c/sshnpd/src/handle_ssh_request.c
+++ b/packages/c/sshnpd/src/handle_ssh_request.c
@@ -468,7 +468,7 @@ void handle_ssh_request(atclient *atclient, pthread_mutex_t *atclient_lock, sshn
       free(session_iv_base64);
     }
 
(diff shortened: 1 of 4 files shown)


packages/c/sshnpd/src/handle_ssh_request.c
```diff
@@ -468,7 +468,7 @@ void handle_ssh_request(atclient *atclient, pthread_mutex_t *atclient_lock, sshn
       free(session_iv_base64);
     }
 
-    int res = run_srv_process(params, host, port, authenticate_to_rvd, rvd_auth_string, encrypt_rvd_traffic,
+    int res = run_srv_process(params, host, port, authenticate_to_rvd, rvd_auth_string, encrypt_rvd_traffic, false,
```
xavierchanth, line 471: in ssh_request code, multi is always false.

realvarx (approved):

PR #1234, opened 2024-07-31
chore: apply c sdk refactors

JeremyTubongbanua:
**- What I did**

**- How I did it**

**- How to verify it**


xavierchanth:
The one bad thing about the new tests in parallel is that it is hard to tell, but the C daemon is likely timing out here. Hence the `2024/07/31 18:33:49 Run Command Timeout` message

xavierchanth:
@JeremyTubongbanua there is cjson used directly in the sshnpd code, can you also update the includes to use `<atclient/cjson.h>`

Same with any mbedtls includes in srv.

xavierchanth (approved):

PR #1237, opened 2024-08-01
feat: (c sshnpd) daemon npt support

realvarx:
**- What I did**
- Implemented `handle_npt_request`

**- How I did it**

**- How to verify it**

### DIFF
--- a/packages/c/srv/src/srv.c
+++ b/packages/c/srv/src/srv.c
@@ -154,7 +154,7 @@ int run_srv_daemon_side_multi(srv_params_t *params) {
 
     char *messagetype = NULL, *new_session_aes_key_string = NULL, *new_session_aes_iv_string = NULL;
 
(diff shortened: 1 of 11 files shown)


realvarx:
**[02 August 2024] Current Status:** If the connections are made slowly and one at a time, it works correctly up to a limited (undefined, random) number of requests. Upon reaching a certain number of requests or attempting to make multiple connections in a very short period, `C sshnpd` gives a series of errors that do not yet seem entirely conclusive (it is unknown whether the connection is dropped due to the inability to process the requests, however, there are quite a few indications that a buffer overflow occurs). After investigating the problem, it seems it could be due to a part of the code related to `monitor` management. It is believed that the problem is not with `srv` or the implementation of `multi`, as these were tested with the Dart version of `sshnpd`, and there were no issues in those circumstances.

xavierchanth:
I would like to enable the npt test for the C daemon on this branch

PR #1223, opened 2024-07-26
fix: use a linked list for process clearing

xavierchanth:
**- What I did**

- Switched the process id queue for srv from a dynamically allocated pointer array to a linked-list so memory is managed more effectively
  - This queue is used to clean up defunct processes - nearly immediately, unless the main thread is the in the middle of handling another request (which is usually a few seconds of delay at most)

**- How I did it**

**- How to verify it**

**- Description for the changelog**
fix: use a linked list for process clearing

PR #1238, opened 2024-08-01
fix: Use a more portable solution to clean up child processes

xavierchanth:
**- What I did**

- Undo #1223  and it's ancestor PRs
- Use SIGCHLD signal to detect when srv processes are finished, and then close them.

**- How I did it**

**- How to verify it**

fix: Use a more portable solution to clean up child processes

### DIFF
--- a/packages/c/sshnpd/include/sshnpd/handle_ping.h
+++ b/packages/c/sshnpd/include/sshnpd/handle_ping.h
@@ -2,6 +2,7 @@
 #define HANDLE_PING_H
 #include "sshnpd/params.h"
 #include <atclient/monitor.h>
+#include <pthread.h>
 void handle_ping(sshnpd_params *params, atclient_monitor_message *message, char *ping_response, atclient *atclient,
                  pthread_mutex_t *atclient_lock);
 #endif
--- a/packages/c/sshnpd/src/main.c
+++ b/packages/c/sshnpd/src/main.c
@@ -75,26 +75,32 @@ static bool is_child_process = false;
 
 // Signal handling
 static volatile sig_atomic_t should_run = 1;
-static void sig_handler(int _) {
-  (void)_;
+static void exit_handler(int sig) {
+  atlogger_log("exit_handler", ATLOGGER_LOGGING_LEVEL_WARN, "Received signal: %d\n");
   if (should_run == 1) {
-    atlogger_log("sig_handler", ATLOGGER_LOGGING_LEVEL_WARN, "Received SIGINT, exiting safely\n");
+    atlogger_log("exit_handler", ATLOGGER_LOGGING_LEVEL_WARN, "Received SIGINT, attempting a safe exit\n");
     should_run = 0;
   } else if (should_run == 0) {
-    atlogger_log("sig_handler", ATLOGGER_LOGGING_LEVEL_WARN, "Received SIGINT again, exiting forcefully\n");
+    atlogger_log("exit_handler", ATLOGGER_LOGGING_LEVEL_WARN, "Received SIGINT again, exiting forcefully\n");
+    exit(1);
   }
 }
-
-// Queue for all srv processes
-static struct sshnpd_process_node *process_head;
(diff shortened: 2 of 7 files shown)


packages/c/sshnpd/src/main.c
```diff
@@ -75,26 +75,32 @@ static bool is_child_process = false;
...
+    atlogger_log("exit_handler", ATLOGGER_LOGGING_LEVEL_WARN, "Received SIGINT again, exiting forcefully\n");
+    exit(1);
+  }
+}
+static void child_exit_handler(int sig) {
+  atlogger_log("child_exit_handler", ATLOGGER_LOGGING_LEVEL_WARN, "Received signal: %d\n");
+  int status;
+  pid_t pid = waitpid(-1, &status, WNOHANG);
+  if (pid > 0 && WIFEXITED(status)) {
+    atlogger_log(LOGGER_TAG, ATLOGGER_LOGGING_LEVEL_DEBUG, "pid %d exited\n", pid);
```
xavierchanth, line 93: New signal handlers

packages/c/sshnpd/src/main.c
```diff
@@ -75,26 +75,32 @@ static bool is_child_process = false;
...
-static void free_sshnpd_process_nodes(struct sshnpd_process_node *process_head);
 
 int main(int argc, char **argv) {
   int res = 0;
   int exit_res = 0;
 
   // Catch sigint and pass to the handler
-  signal(SIGINT, sig_handler);
+  signal(SIGINT, exit_handler);
+  signal(SIGCHLD, child_exit_handler);
```
xavierchanth, line 103: Signal handler setup

packages/c/sshnpd/include/sshnpd/handle_ping.h
```diff
@@ -2,6 +2,7 @@
 #define HANDLE_PING_H
 #include "sshnpd/params.h"
 #include <atclient/monitor.h>
+#include <pthread.h>
```
xavierchanth, line 5: added a missing header

packages/c/sshnpd/src/main.c
```diff
@@ -269,19 +278,19 @@ int main(int argc, char **argv) {
   if (res != 0) {
     atlogger_log(LOGGER_TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "Failed to start refresh device entry thread\n");
     exit_res = res;
-    goto cancel_atclient;
+    goto clean_username_keys;
```
xavierchanth, line 281: A bunch of these gotos were just wrong, so I also fixed the memory cleanup control flow.

packages/c/sshnpd/src/main.c
```diff
@@ -75,26 +75,32 @@ static bool is_child_process = false;
...
   } else if (should_run == 0) {
-    atlogger_log("sig_handler", ATLOGGER_LOGGING_LEVEL_WARN, "Received SIGINT again, exiting forcefully\n");
+    atlogger_log("exit_handler", ATLOGGER_LOGGING_LEVEL_WARN, "Received SIGINT again, exiting forcefully\n");
+    exit(1);
+  }
+}
+static void child_exit_handler(int sig) {
+  atlogger_log("child_exit_handler", ATLOGGER_LOGGING_LEVEL_WARN, "Received signal: %d\n");
+  int status;
+  pid_t pid = waitpid(-1, &status, WNOHANG);
```
xavierchanth, line 91: Using `waitpid(-1, &status, WNOHANG)` seems to be the most portable approach in all of the examples and research that I did.

packages/c/sshnpd/src/main.c
xavierchanth: The main addition here is `child_exit_handler`, where I have moved the `waitpid` call to - previously I was maintaining a list of process ids to check, but the use of -1 here should be more reliable.

xavierchanth:
@cpswan @cconstab my previous solution solved defunct processes on linux, but also seems to have resulted in problems on macos. This solution has been tested on both

cpswan (approved):

PR #1245, opened 2024-08-06
build: C Linux builds in Docker on glibc 2.28.0 base and Musl

### LINKED ISSUE #1220: C sshnpd binary (from c0.1.8 release) won't run on Ubuntu 20.04 due to missing glibc dependency
### Describe the bug

Running `./sshnpd` returns:

```log
./sshnpd: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by ./sshnpd)
./sshnpd: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ./sshnpd)
```

### Steps to reproduce

1. First I downloaded the [sshnp-linux-x64.tgz](https://github.com/atsign-foundation/noports/releases/download/c0.1.8/sshnp-linux-x64.tgz) tarball from the c0.1.8 release
2. Then I extracted the binary: `tar -xvf sshnp-linux-x64.tgz sshnpd`
3. And then I run the binary `./sshnpd`

### Expected behavior

Daemon shows help

### Additional context

`uname -a`

```log
Linux DB4 5.15.153.1-microsoft-standard-WSL2 #1 SMP Fri Mar 29 23:14:13 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
```

`lsb_release -a`

```log
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.6 LTS
Release:        20.04
Codename:       focal
```

`ldd --version`

```log
ldd (Ubuntu GLIBC 2.31-0ubuntu9.16) 2.31
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper.
```

cpswan:
Fixes #1220

**- What I did**

* Linux builds using glibc 2.28.0 so that binaries are compatible with older distros
* Musl based builds

**- How I did it**

* Uses `atsigncompany/cbuildimage` which is a Debian 10 base (with glibc 2.28.0) and modern Clang (18) and CMake (3.30.2)
* Alpine based Dockerfile for Musl with static linking directive

**- How to verify it**

Cut a new release (c0.1.11) after merging and then test resulting binaries on older distros.

See https://github.com/cpswan/release_automation/releases/tag/c0.1.8 for an example of what will happen.

build: C Linux builds in Docker on glibc 2.28.0 base and Musl


cpswan:
Converted to draft as I'm going to throw the Musl stuff in here too (and we need to fix the unit tests before we can merge anyway).

cpswan:
@XavierChanth this is now ready (and the unit tests are passing :) )

xavierchanth (approved):

xavierchanth:
An issue that has come up with the new unit tests:
If GNU parallel ends due to a hang, then we don't get a report back.
I will have to fix that.

Pulled the macos-arm64 build and it was linked correctly. I will do the same check with linux-x64 later today.

PR #1254, opened 2024-08-07
fix: e2e tests crashing

xavierchanth:
the dependency overrides *should* already be using ^0.6.0 as the minimum version since it is a transitive dependency of our override of dartssh2... but that would be asking too much of the toolchain.

**- What I did**

- Hopefully this fixes the e2e tests

**- How I did it**

**- How to verify it**

fix: e2e tests crashing


xavierchanth:
Tested locally and the Dart tests are passing now, the C tests are flaking which is causing a timeout long enough for GNU parallel to exit with a timeout.

```
2024-08-07 15:10:50 | ERROR:     Failed after 2 attempts
2024-08-07 15:10:52 | ERROR:     Test FAILED : exit code 1
###
###
### NoPorts e2e test run complete at 2024-08-07 15:10:52
###
### Of a possible 180, 126 were not applicable (usually version constraints)
### Executed: 54  Passed: 48  Failed: 6
```

tests/e2e_all/scripts/common/run_tests.sh
```diff
@@ -42,12 +42,15 @@ if [ $allowParallelization == "true" ] && command -v env_parallel >/dev/null 2>&
   export -f run_tests_for_daemon
   # Run a round of tests against each daemon in parallel
   parallel --jobs 5 \
+    --timeout 3m \
```
xavierchanth, line 45: This ends up being 3minutes per daemon - which may be a bit short, we will have to see.

JeremyTubongbanua (dismissed):

packages/dart/noports_core/pubspec.yaml
```diff
@@ -30,6 +30,7 @@ dependency_overrides:
     git:
       url: https://github.com/atsign-foundation/dartssh2
       ref: trunk
```
cpswan: Do we need to make a transitive dependency explicit like this? My take is that we just need to bump the lock file(s)

cpswan:
After some discussion with @gkc I've backed out the changes to `pubspec.yaml` and introduced a `pubspec.lock` change that comes from `dart pub upgrade dartssh2`.

We should probably evaluate a mechanism to ensure that `pubspec.lock` is more frequently upgraded as there are a bunch of other outstanding bumps, which include some promotions from v0.x.x to v1.x.x.

gkc (approved):

cpswan:
@XavierChanth please merge if you're happy with the revised approach (or we can discuss in arch).

xavierchanth:
FYI @gkc @cpswan @JeremyTubongbanua , beyond these changes, the e2e tests were failing because I found some non-utf8 characters in the .ssh/authorized_keys file on the test run - The Dart daemon calls `authKeys.readAsString()` (authKeys is a`File`) which does a utf8 decoding - this decode fails due to the non-utf8 characters. Not sure how this happened, but something worth noting in case we run into a similar issue.

PR #1263, opened 2024-08-14
fix: Add unit test related to sshnpd

sitaram-kalluri:
**- What I did**
- Added a unit test to notify an ssh_request to sshnpd and assert the payload in the notification value.

**- How I did it**
- When invoking the SshnpDartPureImpl.run() method, the method call `await callInitialization();` initializes both the sshnpd and sshrvd channels, which leads to the creation a lot of unnecessary mock objects in the test scenario. To address this, the 'notify' call was moved to a separate method, sendSshRequestToSshnpd, to facilitate easier unit testing.
- In the unit test,
    - Invoke `sshnp.srvdChannel.initialize()` which will fetch the host and ports from srvd. 
    - Then, invoke `sshnp.sendSshRequestToSshnpd()`  which will build the notification key - 'ssh_request` and payload as notification value which is sent to sshnpd. 
    - Assert the payload in the mockNotificationService callback.

**- How to verify it**
- All tests should pass.

### DIFF
--- a/packages/dart/noports_core/lib/src/sshnp/impl/sshnp_dart_pure_impl.dart
+++ b/packages/dart/noports_core/lib/src/sshnp/impl/sshnp_dart_pure_impl.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 
 import 'package:at_client/at_client.dart';
 import 'package:dartssh2/dartssh2.dart';
+import 'package:meta/meta.dart';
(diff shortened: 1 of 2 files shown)


gkc (approved):
LGTM thank you @sitaram-kalluri

xavierchanth (approved):

PR #1266, opened 2024-08-19
feat:  adding activation inside universal.sh

Xlin123:
**- What I did**
 introduced a check for activated / inactivated keys in order to improve the work flow of a fresh install.
**- How I did it**
  added a check inside client and device installation using at_activate.
**- How to verify it**
`git checkout xlin-activate-installer`
`./universal.sh`

**-Note**
will change the docs after a release

feat:  adding activation inside universal.sh


xavierchanth (approved):
TY

packages/dart/sshnoports/bundles/universal.sh
```diff
@@ -857,6 +880,7 @@ device() {
   if ! check_cmd sshd; then
     >&2 echo "sshd not found. Please install sshd and ensure it is running."
   fi
+  validate_activation
```
xavierchanth: Wait, we don't want to prompt them to activate their MPKAM on the device, only if it's client. The recommended (and thus automated approach) should always be APKAM to the device - only if someone know what they are doing, and thus can do it manually should they be putting MPKAM keys on a device.

xavierchanth (changes requested):
See comment above

packages/dart/sshnoports/bundles/universal.sh
Xlin123: was supposed to remove that after the arch convo on thursday, thanks for catching that

xavierchanth (approved):

Issue #1289, opened 2024-08-23
Add an e2e test where npt is running from CLI in foreground mode


gkc:
### Describe the bug

Absence of this test led to [this bug](https://github.com/atsign-foundation/noports/issues/1287) making its way into a release

### Steps to reproduce

.

### Expected behavior

.

gkc:
Adjusting sprint points from 1 to 3 as there is some refactoring to be done to prevent code duplication now that we will have three npt e2e tests, with more to come

gkc:
this is mostly done but the branch has been gathering dust while I've been focusing on #1242

PR #1290, opened 2024-08-23
fix: ensure NptImpl's `preRun` is being executed only once when npt cli is run without the `-x` flag

### LINKED ISSUE #1287: 5.6.0 npt x64 binary hanging
### Describe the bug

Whilst testing for https://github.com/atsign-foundation/operations/issues/179 I'm finding that the `npt` binary from the 5.6.0 is causing a hung connection.

### Steps to reproduce

1. First I get sshnpd installed on a test VM
2. Then I run up a test web server (in a gnu screen) with `python -m http.server`
3. And then I run a daemon that's permissioned to connect to that web server (port 8000) `sshnpd -a @bareindoornetball -m @cpswan -d ocinpd1 --po localhost:8000`
4. Then I run an `npt` client to connect: `npt -f @cpswan -t @bareindoornetball -d ocinpd1 -p 8000 -l 8000 -r @rv_eu -K`
5. Then (in another shell) I try to get something from the web server with: `curl localhost:8000`

That `curl` just hangs, whilst if I use `npt` from the 5.5.0 release it's just fine.

### Expected behavior

I get the HTML wrapped directory listing from the test VM

gkc:
**- What I did**
Fixes #1287

**- How I did it**
See commits

**- How to verify it**
- Existing automated tests pass
- I have manually tested the fix; there will be a follow-up PR to [add an e2e test](#1289) covering this scenario


cconstab (approved):
Tested and working fine with 'npt' to a remote machine and connecting to a remote daemon (sshd in my test case)

PR #1298, opened 2024-08-30
refactor: tray and window behaviours

xavierchanth:
**- What I did**

1. Decoupled the tray feature from the rest of the app - tray should know about other features, but other features should not have to know about the tray. We want the tray to pull updates from the app state, the app state shouldn't have to push updates to the tray.

2. Removed the app icon from the dock, it's just kind of there and doesn't work to create a new window when you click it. The app should be interacted with from the tray, not the dock.

> Note for future me: no. 2 should be fine on windows and macOS, but Linux needs extra dependencies depending on the DE / tray bar being used. We also potentially have to deal with X vs Wayland. This will need to be well tested before releasing. But Anything that is X + gtk should be fine if the additional dependencies for the tray are installed. So I think we can easily detect and install the missing dependency on demand for mainstream package managers (apt & dnf). Besides, anyone using Linux outside of the supported group I just mentioned is probably going to be happy with the CLI version of NoPorts.

**- How I did it**

**- How to verify it**

refactor: tray and window behaviours


packages/dart/npt_flutter/lib/app.dart
```diff
@@ -79,7 +79,7 @@ class App extends StatelessWidget {
 
           /// A cubit which manages the system tray entries
           BlocProvider<TrayCubit>(
-            create: (_) => TrayCubit()..initialize(),
+            create: (_) => TrayCubit(),
```
xavierchanth, line 82: This gets called in a more appropriate place now.

packages/dart/npt_flutter/lib/features/favorite/bloc/favorite_bloc.dart
```diff
@@ -32,7 +31,6 @@ class FavoriteBloc extends LoggingBloc<FavoriteEvent, FavoritesState> {
       return;
     }
     emit(FavoritesLoaded(favs.values));
-    App.navState.currentContext?.read<TrayCubit>().reloadFavorites();
```
xavierchanth, line 35: Decoupling - remove all "pushed" updates to the tray.

packages/dart/npt_flutter/lib/features/profile/bloc/profile_bloc.dart
```diff
@@ -38,12 +37,10 @@ class ProfileBloc extends LoggingBloc<ProfileEvent, ProfileState> {
 
     if (profile == null) {
       emit(ProfileFailedLoad(uuid));
-      App.navState.currentContext?.read<TrayCubit>().reloadFavorites();
```
xavierchanth, line 41: More of the same

packages/dart/npt_flutter/lib/features/tray_manager/cubit/tray_cubit.dart
```diff
@@ -87,34 +87,50 @@ class TrayCubit extends LoggingCubit<TrayState> {
     emit(const TrayLoaded());
   }
 
-  Future<void> reloadFavorites() async {
+  Future<void> reload() async {
```
xavierchanth, line 90: No functional changes to the reload tray method, but I did document it for clarity

packages/dart/npt_flutter/lib/features/tray_manager/cubit/tray_cubit.dart
```diff
@@ -87,34 +87,50 @@ class TrayCubit extends LoggingCubit<TrayState> {
...
           toolTip: status,
-          onClick: (_) => e.toggle(),
+          onClick: (_) => fav.toggle(),
         );
       }),
     );
+
+    /// PERF: We should conditionally call setContextMenu if there was a state
+    /// change which resulted in an actual change to the favorites list.
+    /// Currently we just force call updates which is really inefficient
```
xavierchanth, line 131: The lists that we iterate through are not expected to be long, but in the future, if someone were to have 1000s (probably an even bigger magnitude) of profiles loaded in the app we will have to determine if the tray reloading needs to be optimized - reducing useless reloads _may_ help

packages/dart/npt_flutter/lib/features/tray_manager/widgets/tray_manager.dart
```diff
@@ -16,13 +19,62 @@ class TrayManager extends StatefulWidget {
 
 class _TrayManagerState extends State<TrayManager>
     with TrayListener, WindowListener {
+  /// Must strongly type [context] here or Dart will infer the wrong type for
+  /// the [.read()] extension which causes an error
+  void reloadTray(BuildContext context, _) {
+    context.read<TrayCubit>().reload();
+  }
+
```
xavierchanth, line 27: This is the listener which triggers tray reloads whenever a piece of state changes.

packages/dart/npt_flutter/lib/features/tray_manager/widgets/tray_manager.dart
```diff
@@ -16,13 +19,62 @@ class TrayManager extends StatefulWidget {
...
+          /// 1. In an asynchronous background task (who cares)
+          /// 2. Worth it, compared to the potential maintenance costs
+          ...profiles.map((uuid) => BlocProvider<ProfileBloc>(
+                key: Key("TrayManager-$uuid"),
+                create: (context) => profileCacheCubit.getProfileBloc(uuid),
+                child: BlocListener<ProfileBloc, ProfileState>(
+                  listener: reloadTray,
+                ),
+              )),
+        ],
```
xavierchanth, line 74: All of these are the new listeners which will trigger the tray to reload whenever any of these change - could be an edited profile, favorited / unfavorited, start / stop, etc.

packages/dart/npt_flutter/lib/main.dart
```diff
@@ -6,5 +8,11 @@ import 'app.dart';
 Future<void> main() async {
   WidgetsFlutterBinding.ensureInitialized();
   windowManager.ensureInitialized();
-  runApp(const App());
+  try {
+    await windowManager.setSkipTaskbar(true); // Don't show the app icon in dock
```
xavierchanth, line 12: Disable dock icon - the dock icon wasn't functional, and made it look like the app is hung, but really the app is only functional from the tray.

packages/dart/npt_flutter/lib/features/tray_manager/widgets/tray_manager.dart
xavierchanth: I dislike that we need to iterate through each profile and add it as a listener, we could optimize by filtering to only favorited profiles, but that introduces edge cases not worth dealing with right now.

cpswan (approved):

cconstab (approved):

PR #1310, opened 2024-09-04
ci: policy docker build

xavierchanth:
**- What I did**

See https://github.com/atsign-foundation/noports/actions/runs/10708485526

**- How I did it**

**- How to verify it**

**- Description for the changelog**

PR #1313, opened 2024-09-05
build(deps): Update dependabot for new Dockerfile location

cpswan:
https://github.com/atsign-foundation/noports/pull/1310 introduced a new Dockerfile location

**- What I did**

Added the new Dockerfile location to Dependabot config so that the Dockerfiles get updated.

**- How to verify it**

Will need to be merged to trunk before we see it in Insights

build(deps): Update dependabot for new Dockerfile location


cpswan:
@gkc this should be merged before #1314

gkc (approved):

PR #1314, opened 2024-09-05
feat: BETA: policy service using atServer as backing store; admin API and UI

gkc:
**- What I did**
- new BETA policy binary, npp_atserver: fetches its rules from atServer as persistent store
- new BETA binary, np_admin: exposes an API via http for setting policy rules & seeing policy info, events etc
- new BETA webapp bundled with np_admin to provide a simple illustrative UI

Note that this is still only alpha quality but the intention is that will be beta quality in our 5.7.0 release

**- How I did it**
See commits

**- How to verify it**
- Tests pass
- multibuild and dockerfile.package work as intended

**TODO in subsequent PRs**
- take up an [at_client.AtRpc enhancement](https://github.com/atsign-foundation/at_client_sdk/pull/1388)
- add concrete types for events, logs etc rather than json/map nonsense
- add permitOpen to the NPAAuthCheckRequest so we can have policy be able to know if a request should be denied when no matching permitOpen found
- refactor: rename all the NPA/npa classes/files/directories to NPP/npp
- UI refactoring - split into pieces, reduce duplicate code
- UI<->API enhancements re events, logs etc
- Documentation
- Postgres impl


.github/workflows/multibuild.yaml
```diff
@@ -85,8 +91,21 @@ jobs:
           dart compile exe bin/activate_cli.dart -v -o sshnp/at_activate${{ matrix.ext }}
           dart compile exe bin/sshnp.dart -v -o sshnp/sshnp${{ matrix.ext }}
           dart compile exe bin/npt.dart -v -o sshnp/npt${{ matrix.ext }}
+          dart compile exe bin/npp_file.dart -v -o sshnp/npa_file${{ matrix.ext }}
+          dart compile exe bin/npp_file.dart -v -o sshnp/npp_file${{ matrix.ext }}
```
gkc, line 95: standardizing on npp_ prefix for all policy binaries; given we had npa_file out there already, preserving it for now

tools/multibuild/Dockerfile.package
```diff
@@ -0,0 +1,60 @@
...
+  mkdir -p /sshnp/debug; \
+  mkdir -p /sshnp/beta/web/admin; \
+  mkdir /tarball; \
+  dart pub get --enforce-lockfile; \
+  dart run build_runner build --delete-conflicting-outputs; \
+  dart compile exe bin/activate_cli.dart -v -o /sshnp/at_activate; \
+  dart compile exe bin/sshnp.dart -v -o /sshnp/sshnp; \
+  dart compile exe bin/npt.dart -v -o /sshnp/npt; \
+  dart compile exe bin/npp_file.dart -v -o /sshnp/npp_file; \
+  ln -s sshnp/npp_file sshnp/npa_file; \
```
xavierchanth, line 27: Multibuild did not like this:
https://github.com/atsign-foundation/noports/actions/runs/10721898228/job/29731556532

I think it is okay to make breaking changes in the alpha release and just remove npa_file and leave in npp_file

xavierchanth: I see why, needs to be absolute path, but I still don't think a symbolic link in the release is a good idea

PR #1332, opened 2024-09-09
fix: npt atSign handling

### LINKED ISSUE #1331: npt atSign handling
### Describe the bug

NptParams accepts atSigns without the `@` prefix which leads to problems later (in the relay server auth check and possibly elsewhere)

### Steps to reproduce

Run npt with one or more atSign args which are not prefixed with `@`

### Expected behavior

- NptParams object should not allow malformed atSign args

gkc:
**- What I did**
- Fixes #1331
- ensure npt rejects malformed atSign args

**- How I did it**
- added atSign parameter checking (via AtUtils.fixAtSign method) to bin/npt.dart and to the NptParams constructor
- Note: did not modify SshnpParams nor the shared superclass ClientParamsBase because of the complication of `--list-devices` which should allow blank daemon and relay atSigns

**- How to verify it**
Tests pass


xavierchanth (changes requested):
Enforcing lowercase is a breaking change.

gkc:
> Enforcing lowercase is a breaking change.

Yes this is technically a breaking change re the noports programs - however, atSigns may only ever include lowercase, so this should never break anyone's installation. Nevertheless I'll remove the strict validation and just do the coercion to lowercase silently.

xavierchanth (approved):

PR #1374, opened 2024-09-23
chore: move version to version.h and bump to 0.2.0

xavierchanth:
**- What I did**

- Removed c daemon version from all header files
- added c daemon version 0.2.0 to version.h (a new header file)

**- How I did it**

**- How to verify it**

chore: move version to version.h and bump to 0.2.0

### DIFF
--- a/packages/c/srv/src/srv.c
+++ b/packages/c/srv/src/srv.c
@@ -139,7 +139,7 @@ int run_srv_daemon_side_multi(srv_params_t *params) {
       len = res;
     }
 
-    if (&control_side.transformer != NULL) {
+    if (control_side.transformer != NULL) {
       unsigned char *output = malloc(4096 * sizeof(unsigned char));
       if (output == NULL) {
         goto exit;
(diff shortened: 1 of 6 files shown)


packages/c/srv/src/srv.c
```diff
@@ -139,7 +139,7 @@ int run_srv_daemon_side_multi(srv_params_t *params) {
       len = res;
     }
 
-    if (&control_side.transformer != NULL) {
+    if (control_side.transformer != NULL) {
```
xavierchanth, line 142: transformer is already a pointer, this was grabbing the address of address...

xavierchanth: This is exactly why I have -Werror enabled in the CI, to ensure that no matter what tooling we use across the team we catch things like this...

packages/c/sshnpd/include/sshnpd/version.h
```diff
@@ -0,0 +1,4 @@
+#ifndef SSHNPD_VERSION_H
+#define SSHNPD_VERSION_H
+#define SSHNPD_VERSION "0.2.0"
```
xavierchanth, line 3: separate header file where **only** the version number is defined

cpswan (approved):

PR #1384, opened 2024-09-24
fix(npt_flutter): sockets close properly, profile rendering bug

xavierchanth:
**- What I did**

- Fixed a bug where the socket connector stays bound to the localPort when it times out in the desktop app
  - This bug is limited to the desktop app
- Fixed a bug preventing the profiles from rendering in the dashboard of the UI

**- How I did it**

**- How to verify it**

fix(npt_flutter): sockets close properly, profile rendering bug

### DIFF
--- a/packages/dart/npt_flutter/lib/features/profile/bloc/profile_bloc.dart
+++ b/packages/dart/npt_flutter/lib/features/profile/bloc/profile_bloc.dart
@@ -109,8 +109,12 @@ class ProfileBloc extends LoggingBloc<ProfileEvent, ProfileState> {
   }
 
   Future<void> _onStart(ProfileStartEvent event, Emitter<ProfileState> emit) async {
-    if (state is! ProfileLoadedState || state is ProfileStarting || state is ProfileStopping || state is ProfileStarted)
+    if (state is! ProfileLoadedState ||
+        state is ProfileStarting ||
+        state is ProfileStopping ||
+        state is ProfileStarted) {
       return;
+    }
     // ProfileLoaded and ProfileFailedSave are both ProfileLoadedState
     var profile = (state as ProfileLoadedState).profile;
     emit(ProfileStarting(uuid, profile: profile));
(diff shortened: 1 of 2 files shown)


packages/dart/npt_flutter/lib/features/profile/bloc/profile_bloc.dart
```diff
@@ -158,26 +162,20 @@ class ProfileBloc extends LoggingBloc<ProfileEvent, ProfileState> {
...
-      sc = await npt
-          .runInline()
-          // Todo - make this timeout configurable from settings
-          .timeout(
-        const Duration(seconds: 10),
-        onTimeout: () {
-          return TimedOutSocketConnector();
-        },
-      );
+      sc = await npt.runInline();
```
xavierchanth, line 175: If the timeout interrupts npt and the underlying socketConnector while the socket is bound, then the socket will stay bound and we have no surface to unbind it from here, hence I have removed the timeout.

packages/dart/npt_flutter/lib/features/profile_list/view/profile_list_view.dart
```diff
@@ -84,15 +84,12 @@ class ProfileListView extends StatelessWidget {
                                     child: ListView.builder(
                                       itemCount: state.profiles.length,
                                       itemBuilder: (context, index) {
-                                        if (state.profiles.isEmpty) {
```
xavierchanth, line 87: This if shouldn't be here and prevents profiles from rendering when they should be. Left over from a refactor

packages/dart/npt_flutter/lib/features/profile/bloc/profile_bloc.dart
```diff
@@ -158,26 +162,20 @@ class ProfileBloc extends LoggingBloc<ProfileEvent, ProfileState> {
...
 
-      cancelSubs = () {
+      cancel = () {
         progressSub?.cancel();
         progressSub = null;
 
         errorSub?.cancel();
         errorSub = null;
+
+        if (sc is SocketConnector) sc.close();
```
xavierchanth, line 172: renamed function from cancelSubs to cancel and added an extra attempt to close the socket connector, this function is called during clean up after npt has finished. If for some reason we end up in a state where socket connector is not closed, but are in the finishing state then this ensures proper cleanup

CurtlyCritchlow (approved):
lgtm

Issue #1411, opened 2024-09-30
Fix code scanning alert - Pinned-Dependencies (npm)


cpswan:
Tracking issue for:
- [x] https://github.com/atsign-foundation/noports/security/code-scanning/24
- [x] https://github.com/atsign-foundation/noports/security/code-scanning/25
- [x] https://github.com/atsign-foundation/noports/security/code-scanning/26

Fixed the above but also got these, which also have now been fixed
- [x] https://github.com/atsign-foundation/noports/security/code-scanning/30
- [x] https://github.com/atsign-foundation/noports/security/code-scanning/31
- [x] https://github.com/atsign-foundation/noports/security/code-scanning/32

gkc:
Ack

gkc:
@cpswan Closing; alerts have been resolved

PR #1446, opened 2024-10-14
build: require npm version 10.9.0 for the policy webapp. Addresses #1411

gkc:
**- What I did**
build: require npm version 10.9.0 for the policy webapp. Addresses #1411

**- How I did it**
- run `npm install -g npm@10.9.0` everywhere before running `npm install`
- add .npmrc with `engine-strict=true`
- pin required npm version to 10.9.0 in package.json

**- How to verify it**
Code-scanning alerts referenced in #1411 should be resolved

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -44,3 +44,7 @@ updates:
     directory: "/apps/admin/webapp/"
     schedule:
       interval: "daily"
+    groups:
+      npm:
+        patterns:
+          - "*"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -40,3 +40,7 @@ updates:
       pip:
         patterns:
           - "*"
+  - package-ecosystem: 'npm'
+    directory: "/apps/admin/webapp/"
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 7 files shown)


apps/admin/webapp/package.json
cpswan, line 1: I think we need to pin versions here to fully clear the alerts.

There should also be an additional Dependabot config to keep those pinned versions up to date.

cpswan (changes requested):

apps/admin/webapp/package.json
gkc: > I think we need to pin versions here to fully clear the alerts.
> 
> There should also be an additional Dependabot config to keep those pinned versions up to date.

Both done

.github/dependabot.yml
```diff
@@ -40,3 +40,7 @@ updates:
       pip:
         patterns:
           - "*"
+  - package-ecosystem: 'npm'
+    directory: "/apps/admin/webapp/"
+    schedule:
+      interval: "daily"
```
cpswan, line 46: This should also use groups like the ones above so that we don't get lots of PRs when a bunch of stuff changes at once e.g.:
```
    groups:
      npm:
        patterns:
          - "*"
```

cpswan (changes requested):

.github/dependabot.yml
gkc: done

cpswan (approved):

PR #1451, opened 2024-10-14
build: pin npm to version 10.8.2

gkc:
**- What I did**
build: pin npm to version 10.8.2 because of [this](https://github.com/atsign-foundation/noports/actions/runs/11328424207) which is caused by [this](https://github.com/dependabot/dependabot-core/issues/9277)

**- How I did it**
See commits

**- How to verify it**
Checks pass and dependabot checks pass


cpswan (dismissed):

gkc:
@cpswan Even though `npm install` will now fail with an ugly error unless you are running npm 10.8.2, the code scanner still pops it up as an issue. This feels like a code scanning tool issue to me? since I am now doing precisely what is recommended to package.json to ensure the specific version of npm is being used? However I've just pushed another commit [as per this documentation](https://docs.npmjs.com/cli/v10/commands/npm-ci) which will ensure that package-lock.json is respected - it may be that that is the actual problem being flagged by the code scanning tool here.

cpswan (approved):

PR #1458, opened 2024-10-18
feat: C SSHNPD `--monitor-read-timeout` flag

JeremyTubongbanua:
**- What I did**
- Implemented `--monitor-read-timeout` option in C SSHNPD
- Defaults to 10 seconds

How it works:

Let's say the monitor read timeout is 10 seconds.

Then the monitor connection will do a blocking read on the network buffer for 10 seconds before giving up. Once it gives up, it will send `noop:0\r\n` to do a check if the monitor connection is alive.

With this new implementation, the monitor connection will only send a `noop:0\r\n` every *`monitor-read-timeout`* seconds, at most.

feat: C SSHNPD `--monitor-read-timeout` flag

### DIFF
--- a/packages/c/cmake/atsdk.cmake
+++ b/packages/c/cmake/atsdk.cmake
@@ -3,7 +3,7 @@ if(NOT atsdk_FOUND)
   FetchContent_Declare(
     atsdk
     GIT_REPOSITORY https://github.com/atsign-foundation/at_c.git
-    GIT_TAG 292e3842bf13e1479ed45e1a02821b806156218d
+    GIT_TAG 86c2e2fe3334050501f8ff820437227bd0a95488
   )
   FetchContent_MakeAvailable(atsdk)
   install(TARGETS atclient atchops atlogger)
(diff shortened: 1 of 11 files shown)


JeremyTubongbanua:
I added the default constant to `sshnpd.h` @XavierChanth 
Is that the correct place?

xavierchanth:
Most of the changes were in at_c where I was able to handle the socket code. It turns out mbedtls does have handlers for socket closed, but they shouldn't fully be relied on. We do check for them now, but there are also scenarios where the monitor connection is dead and the socket is still considered open, hence we have configured noop to 45 seconds in this PR.

This noop is only called if the connection has silently failed since the frequency of stats notifications is more often.

xavierchanth:
> I added the default constant to `sshnpd.h` @XavierChanth Is that the correct place?

It is the correct place, but I renamed it to main.h to be more appropriate.

xavierchanth:
I've changed the behavior in sshnpd, we don't need this to be configurable externally anymore.

PR #1494, opened 2024-11-05
fix: 32bit support 

xavierchanth:
**- What I did**

- Changed `long` cast to `int64_t` where we have a timeout which exceeds 4 bytes (30 days in milliseconds for device_info expiry)
- I also snuck in some `.clangd` changes as our recent introduction of the shared clangd settings is using the default std (some c++ version), instead of our desired c99

**- How I did it**

**- How to verify it**

fix: 32bit support

### DIFF
--- a/packages/c/.clangd
+++ b/packages/c/.clangd
@@ -2,7 +2,7 @@ CompileFlags:
   # Use gcc in development for better diagnostics
   # & Ensure that compilation doesn't fail due to warnings
   Compiler: gcc
-  Add: [-Wno-error]
+  Add: [-Wno-error, -xc, -std=c99]
 
 Index:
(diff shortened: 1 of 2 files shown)


packages/c/.clangd
```diff
@@ -2,7 +2,7 @@ CompileFlags:
   # Use gcc in development for better diagnostics
   # & Ensure that compilation doesn't fail due to warnings
   Compiler: gcc
-  Add: [-Wno-error]
+  Add: [-Wno-error, -xc, -std=c99]
```
xavierchanth, line 5: `-xc` = use the C language instead of the default `-xc++`

cpswan (approved):

PR #1498, opened 2024-11-06
fix: use immediate state for reloading the tray where possible

xavierchanth:
**- What I did**

- The only case where we may see old values is from ProfileRunningState, but ProfileState should also trigger a reload, so this is likely not an issue at all. Will observe to confirm.

**- How I did it**

**- How to verify it**

fix: use immediate state for reloading the tray where possible


packages/dart/npt_flutter/lib/features/tray_manager/widgets/tray_manager.dart
```diff
@@ -22,15 +26,30 @@ class TrayManager extends StatefulWidget {
...
+        cubit.reload(profilesRunningState: state);
+      case SettingsLoadedState _:
+        var localizations = await AppLocalizations.delegate.load(state.settings.language.locale);
+        cubit.reload(localizations: localizations);
+      case ProfileState _:
+        cubit.reload(profileState: state);
+      default:
+        cubit.reload(localizations: AppLocalizations.of(context));
+    }
   }
```
xavierchanth, line 46: This is where the magic happens... we conditionally pass the state to the appropriate field to ensure that `TrayCubit.reload` uses immediate values wherever possible, rather than pulling from possibly outdated context.

packages/dart/npt_flutter/lib/features/tray_manager/widgets/tray_manager.dart
```diff
@@ -22,15 +26,30 @@ class TrayManager extends StatefulWidget {
...
 
   @override
   Widget build(BuildContext context) {
     var trayCubit = context.read<TrayCubit>();
     if (trayCubit.state is TrayInitial) {
-      trayCubit.initialize();
+      trayCubit.initialize(localizations: AppLocalizations.of(context));
```
xavierchanth, line 52: Tray cubit also takes localizations as input in the initialize function, so that the it can immediately set the tray with the desired language on startup.

packages/dart/npt_flutter/lib/features/tray_manager/widgets/tray_manager.dart
```diff
@@ -55,6 +74,17 @@ class _TrayManagerState extends State<TrayManager> with TrayListener, WindowList
...
+              listener: reloadTray,
+              // Only call listener when the language changes in settings
+              listenWhen: (prev, next) {
+                if (prev is SettingsLoadedState && next is SettingsLoadedState) {
+                  return prev.settings.language != next.settings.language;
+                }
+                // This may cause some extra reloading (very occasionally, settings shouldn't change often)
+                // but it should catch all of the edge cases
+                return prev is SettingsLoadedState || next is SettingsLoadedState;
+              }),
```
xavierchanth, line 87: Added settings bloc listener for the locale, but it only should  trigger a reload when the language changes (`listenWhen` function)

packages/dart/npt_flutter/lib/localization/app_en.arb
```diff
@@ -1,4 +1,5 @@
 {
+
```
xavierchanth, line 2: Extracted these localizations from your branch @CurtlyCritchlow

CurtlyCritchlow (approved):

PR #1521, opened 2024-11-11
fix: clang-tidy using private macos headers

xavierchanth:
**- What I did**

- Corrected any headers starting with "sys/"
- Disabled the aggressive clang-tidy rule

**- How I did it**

**- How to verify it**

fix: clang-tidy using private macos headers


cpswan (approved):

xavierchanth:
I should not have merged that... mobile conceals the tests... I need to restore headers for waitpid

PR #1522, opened 2024-11-11
fix: undo sys/wait removal

xavierchanth:
**- What I did**

- Increased clangd diagnostics by adding the following to `~/.config/clangd/config.yaml` (these are gcc flags so the compiler must be set to gcc)

```yaml
CompileFlags:
  Add:
    - "-Wall"
    - "-Wextra"
    - "-Werror"
    - "-Werror-implicit-function-declaration"
```

- Fixed newly raised diagnostic issues
- Suppressed some issues with clang pragmas (unspecified arguments in argparse opt initializers)

**- How I did it**

**- How to verify it**


cpswan (dismissed):

xavierchanth:
It seems our compiler rules are stricter in NoPorts... and this propagating a whole bunch of error through the test suite... I will need to make a ton of changes to at_c...

I'm not sure why we are inconsistent, nor why we are only seeing these come through now.

cpswan (dismissed):

packages/c/srv/src/srv.c
```diff
@@ -350,10 +350,13 @@ int socket_to_socket(const srv_params_t *params, const char *auth_string, chunke
   return 0;
 }
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-parameter"
 int server_to_socket(const srv_params_t *params, const char *auth_string, chunked_transformer_t *encrypter,
                      chunked_transformer_t *decrypter) {
   return 0;
 }
+#pragma clang diagnostic pop
```
xavierchanth, line 359: This method is currently not implemented or used, the code path that would call it, simply returns a method that says "--bind-local-port is not supported"

packages/c/srv/src/srv.c
```diff
@@ -438,7 +441,6 @@ int aes_ctr_crypt_stream(const chunked_transformer_t *self, size_t len, const un
 
 static int process_multiple_requests(char *original, char **requests[], size_t *num_out_requests) {
   int ret = -1;
-  int num_requests = 0;
```
xavierchanth, line 441: unused variable

cpswan (approved):

PR #1539, opened 2024-11-17
build: exclude policy admin webapp assets from the docker distributions

gkc:
**- What I did**
- build: include policy admin API but exclude policy admin webapp assets from the docker distributions
- Context: (a) installing node slows down the dockerfile build quite a lot; (b) required version of node is not available as standard (i.e. via apt-get) for the platforms in question so can't (currently) build the webapp anyway (c) if required, the webapp assets can be copied from other downloads

**- How I did it**
- Removed the node install and npm build steps from tools/multibuild/Dockerfile.package
- Additionally, although unrelated, now specifying npm version to use when building the webapp assets in multibuild.yaml main_build

**- How to verify it**
Manually verified by running multibuild workflow against this branch - see https://github.com/atsign-foundation/noports/actions/runs/11878280981/job/33098772780

### DIFF
--- a/.github/workflows/multibuild.yaml
+++ b/.github/workflows/multibuild.yaml
@@ -77,6 +77,9 @@ jobs:
       - uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
         with:
           node-version: '20.17.0'
+      # setup required npm version
+      - run: |
+          npm install -g npm@10.8.2
       # create directories need for build
       - run: |
           mkdir -p sshnp/web/admin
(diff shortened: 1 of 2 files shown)


xavierchanth (approved):

xavierchanth:
We can always offer a second docker image for the web app and provide an example compose file of how to run both. Although, they have no internal dependencies so there's no special configuration, they just need the same atsign.

cpswan (approved):

gkc:
> We can always offer a second docker image for the web app and provide an example compose file of how to run both. Although, they have no internal dependencies so there's no special configuration, they just need the same atsign.

@XavierChanth yes; also could copy the web assets into the docker file from a pre-docker-file-build step

xavierchanth:
> > We can always offer a second docker image for the web app and provide an example compose file of how to run both. Although, they have no internal dependencies so there's no special configuration, they just need the same atsign.
> 
> @XavierChanth yes; also could copy the web assets into the docker file from a pre-docker-file-build step

I wasn't planning on working on it until we have an ask.

cpswan:
I can take a look at this.

PR #1552, opened 2024-11-19
fix: mandarin and cantonese state is saved across app restarts.

CurtlyCritchlow:
**- What I did**
Updated `settings.toJson()` so  mandarin and cantonese language is saved across app restarts.
**- How I did it**
run `dart run build_runner watch  build -d` in the thermal to update changes made in 'Settings' class.
**- How to verify it**


xavierchanth:
These seem like non-standard locale codes, can you explain why we've opted to drop the initial language specifier rather than propogate the region specifier?

CurtlyCritchlow:
> These seem like non-standard locale codes, can you explain why we've opted to drop the initial language specifier rather than propogate the region specifier?

It doesn't make a difference what the json value of the enum is since the app is using the language based on the enum itself and its [extension methods](https://github.com/atsign-foundation/noports/blob/0614550be0ae30c078c3c674833f76c47d41c55d/packages/dart/npt_flutter/lib/util/language.dart#L18), not based the jsonValue of the enum. That being said I can update it but it doesn't improve the app one iota.

xavierchanth:
> > These seem like non-standard locale codes, can you explain why we've opted to drop the initial language specifier rather than propogate the region specifier?
> 
> 
> 
> It doesn't make a difference what the json value of the enum is since the app is using the language based on the enum itself and its [extension methods](https://github.com/atsign-foundation/noports/blob/0614550be0ae30c078c3c674833f76c47d41c55d/packages/dart/npt_flutter/lib/util/language.dart#L18), not based the jsonValue of the enum. That being said I can update it but it doesn't improve the app one iota.

Can you double check that these don't collide with any standard locales, just in case we add more in the future

CurtlyCritchlow:
> > > These seem like non-standard locale codes, can you explain why we've opted to drop the initial language specifier rather than propogate the region specifier?
> > 
> > 
> > It doesn't make a difference what the json value of the enum is since the app is using the language based on the enum itself and its [extension methods](https://github.com/atsign-foundation/noports/blob/0614550be0ae30c078c3c674833f76c47d41c55d/packages/dart/npt_flutter/lib/util/language.dart#L18), not based the jsonValue of the enum. That being said I can update it but it doesn't improve the app one iota.
> 
> Can you double check that these don't collide with any standard locales, just in case we add more in the future

Done. I updated the json_value so there will be no potential conflict if we add additional languages in the future.

xavierchanth:
The build_runner needs to be re-run to regenerate the json model

CurtlyCritchlow:
Done, thanks for the reminder.

xavierchanth (approved):

PR #1565, opened 2024-11-22
fix: use a more aggressive handler for SIGINT and SIGCHLD

xavierchanth:
**- What I did**

- Added back the SIGINT and SIGCHLD handlers, but with no double trap

**- How I did it**

**- How to verify it**

fix: use a more aggressive handler for SIGINT and SIGCHLD


gkc (commented):
LGTM but make failing on CI

gkc (approved):

PR #1596, opened 2024-12-05
feat: Use override.conf files for systemd unit config

### LINKED ISSUE #1567: upgrading sets /etc/systemd/system/sshnpd.service back to template so reconfiguration required
### Describe the bug

Report from customer:
> Doc says to follow the install instructions e.g. run install.sh.   The end result was /etc/systemd/system/sshnpd.service was set back to the template so had to be reconfigure it.  Unless I missed something it would be much better if when upgrading it could leave the previously configured file or at a min copy the existing one?

### Steps to reproduce

- Install v5.6.1 on a Linux system choosing systemd as the run mechanism. Update sshnpd.service to be the way you want it
- Upgrade the installation to v5.7.0
- sshnpd.service has reverted to template

### Expected behavior

- If sshnpd.service already exists then it should not be updated

### Screenshots

_No response_

### Smartphones

_No response_

_No response_

### Additional context

_No response_

cpswan:
Fixes #1567 though we have more work to do to make the upgrade experience smoother

**- What I did**

* introduced `override.conf` files for systemd units
* edited `install.sh` to put in place (and not overwrite) `override.conf`
* modified `universal.sh` to place config into `override.conf` and not disrupt existing config.

**- How to verify it**

We'll have to do a pre-release once this is merged

feat: Use override.conf files for systemd unit config


cpswan:
Making draft until we've had a chance to discuss in arch - particularly that this change will impact config stored in existing systemd unit files.

cpswan:
@XavierChanth if you want to merge this then please go ahead. If not put it back to draft and I'll take another pass at how to migrate existing config.

packages/dart/sshnoports/bundles/shell/systemd/sshnpd.service
```diff
@@ -16,36 +10,7 @@ Type=simple
...
-# Comment to disable verbose logging
-Environment=v="-v"
-
-# The line below runs the sshnpd service, with the options set above.
-# You can edit this line to further customize the service to your needs.
-ExecStart=/usr/local/bin/sshnpd -a "$device_atsign" -m "$manager_atsign" -d "$device_name" "$delegate_policy" "$s" "$u" "$v"
+# The line below runs the sshnpd service, with the options set in
+# /etc/systemd/system/sshnpd.d/override.conf.
+# You can edit that config with: sudo systemctl edit sshnpd
+ExecStart=/usr/local/bin/sshnpd -a "$device_atsign" -m "$manager_atsign" -d "$device_name" "$delegate_policy" "$s" "$u" "$v" "$additional_args"
```
xavierchanth, line 16: One thing I've been meaning to address, which may be suitable as part of this changeset is that when we use policy only the manager configuration will be `-m ""` which prevents the daemon from starting.

xavierchanth: Maybe we can change it to: `$( [ -n "$manager_atsign" ] && "-m $manager_atsign" || "" )`, and have a similar wrapper for policy.

xavierchanth:
> @XavierChanth if you want to merge this then please go ahead. If not put it back to draft and I'll take another pass at how to migrate existing config.

I think we should improve the upgrade / migration path before we merge.

packages/dart/sshnoports/bundles/shell/systemd/sshnpd.service
xavierchanth: This can also be done in a later PR

packages/dart/sshnoports/bundles/shell/install.sh
```diff
@@ -214,10 +217,52 @@ post_systemd_message() {
...
 install_systemd_unit() {
   unit_name="$1"
+  systemd_unit="$systemd_dir/$unit_name"
+  systemd_config="$systemd_unit.d/override.conf"
   no_mac
-  mkdir -p "$systemd_dir"
-  dest="$systemd_dir/$unit_name"
-  cp "$script_dir/systemd/$unit_name" "$dest"
+  if [ -f "$systemd_unit" ]; then
+    # migrate old config from systemd unit file to override.conf
```
cpswan, line 224: This has been the focus of my testing for the past few days - making sure that we can migrate existing config into the new approach.

packages/dart/sshnoports/bundles/universal.sh
```diff
@@ -849,6 +849,42 @@ device() {
...
+    echo "$install_output"
+  fi
+
+  # upgrade an existing installation if we find one
+  case "$device_install_type" in
+    launchd)
+      launchd_plist="$HOME/Library/LaunchAgents/com.atsign.sshnpd.plist"
+      if [ -f "$launchd_plist" ]; then
+        echo "launchd config already in place"
+        # TODO: restart service?
```
cpswan, line 868: @XavierChanth what would be the appropriate command here?

packages/dart/sshnoports/bundles/shell/systemd/sshnpd.service
cpswan: I was thinking this should actually be something like `${manager_atsign:+-m "$manager_atsign"}`, but it turns out that [systemd doesn't support parameter expansion](https://unix.stackexchange.com/questions/216780/why-does-bash-parameter-expansion-not-work-inside-systemd-service-files)

One option might be to wrap our command line inside a shell invocation, but I'm wary of creating a deeper process tree.

xavierchanth: Seems like the safest option is to allow sshnpd to try to pull those from the environment if they aren't explicitly passed.

packages/dart/sshnoports/bundles/universal.sh
xavierchanth: From what I grok in the launchctl man page it's supposed to be this:

```
launchctl disable gui/$UID/com.atsign.sshnpd
launchctl enable gui/$UID/com.atsign.sshnpd
```

And I get a success code from both commands, but neither is actually stopping or starting the service...

xavierchanth: Ah, that doesn't work.. but this works:

`launchctl kickstart -k gui/$UID/com.atsign.sshnpd`

-k says to kill the running instance first.

xavierchanth: Confirmed by checking process ids for sshnpd

xavierchanth (approved):

xavierchanth:
systemd part LGTM

PR #1600, opened 2024-12-06
fix: error atsign doesn't exist

xavierchanth:
- Apple app reviewers don't know how to read or hit a cancel button I guess.

**- What I did**

**- How I did it**

**- How to verify it**

fix: error atsign doesn't exist

### DIFF
--- a/packages/dart/npt_flutter/pubspec.yaml
+++ b/packages/dart/npt_flutter/pubspec.yaml
@@ -16,7 +16,16 @@ publish_to: "none" # Remove this line if you wish to publish to pub.dev
 # https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html
 # In Windows, build-name is used as the major, minor, and patch parts
 # of the product and file versions while build-number is used as the build suffix.
-version: 1.1.3+0
+version: 1.1.4+4
+msix_config:
+  display_name: "NoPorts Desktop"
+  publisher_display_name: Atsign Inc
+  identity_name: TheCompany.NoPortsDesktop
+  publisher: CN=BBFE1D0B-F713-4C7F-B375-5EA851CBB1FF
+  msix_version: 1.1.4.4
+  logo_path: "assets/logo.png"
+  capabilities: internetClient
+  store: true
 
 environment:
   sdk: ^3.5.0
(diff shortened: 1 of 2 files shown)


packages/dart/npt_flutter/pubspec.yaml
```diff
@@ -16,7 +16,16 @@ publish_to: "none" # Remove this line if you wish to publish to pub.dev
 # https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html
 # In Windows, build-name is used as the major, minor, and patch parts
 # of the product and file versions while build-number is used as the build suffix.
-version: 1.1.3+0
+version: 1.1.4+0
```
CurtlyCritchlow: version number should be 1.1.4+4. The +4 indicate it is the 4th build. validation will fail if that number is not greater than the previous number.

CurtlyCritchlow (dismissed):

packages/dart/npt_flutter/pubspec.yaml
xavierchanth: Oh... you're right...

CleanShot 2024-12-06 at 20 41 10@2x

Macos is just weird, iOS' way is the normal way things tend to be done... a build number must only be unique to each version number. But for some reason, with macos... build number must be unique, regardless of the version number...

packages/dart/npt_flutter/pubspec.yaml
```diff
@@ -16,7 +16,16 @@ publish_to: "none" # Remove this line if you wish to publish to pub.dev
...
 # In Windows, build-name is used as the major, minor, and patch parts
 # of the product and file versions while build-number is used as the build suffix.
-version: 1.1.3+0
+version: 1.1.4+4
+msix_config:
+  display_name: "NoPorts Desktop"
+  publisher_display_name: Atsign Inc
+  identity_name: TheCompany.NoPortsDesktop
+  publisher: CN=BBFE1D0B-F713-4C7F-B375-5EA851CBB1FF
+  msix_version: 1.1.4.4
```
CurtlyCritchlow, line 25: The msix_version number need to be 1.1.4.0. It will fail Microsoft validation if the last digit is not 0.

CurtlyCritchlow (approved):

PR #1627, opened 2024-12-18
feat: docker demo env

xavierchanth:
**- What I did**

**- How I did it**

**- How to verify it**


cconstab:
Would love some Windows instructions too, it all works until you hit client.sh on windows

xavierchanth:
> Would love some Windows instructions too, it all works until you hit client.sh on windows

If you have wsl, docker in wsl binds to Windows host's docker socket by default, so you should be able to run everything there.

PR #1655, opened 2025-01-14
fix: Revert "build(deps): Bump the docker group across 2 directories with 1 update"

xavierchanth:
Reverts atsign-foundation/noports#1651

Jobs failing:

https://github.com/atsign-foundation/noports/actions/runs/12776419711

### DIFF
--- a/packages/dart/sshnoports/tools/Dockerfile
+++ b/packages/dart/sshnoports/tools/Dockerfile
@@ -17,7 +17,7 @@ RUN \
   dart compile exe bin/srv.dart -o ${BINARYDIR}/srv
 
 # Second stage of build FROM debian-slim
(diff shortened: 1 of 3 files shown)


cpswan:
Overtaken by #1656

cpswan:
@XavierChanth what do we need to do to recreate the things that failed to be built?

xavierchanth:
> @XavierChanth what do we need to do to recreate the things that failed to be built?

I can uptake the fixed build and we can retag as 5.8.1 or we can create a new tag as 5.8.1+1

xavierchanth:
Actually one more change needs to be made in at_c: strdup function calls

cpswan:
Then let's just do a 5.8.2

PR #1664, opened 2025-01-16
feat: Added APKAM enrollment authorisation functionality into the app

Zambrella:
**- What I did**
- Created a new authorisation page to display `AuthorisationHub` widget
- Created a repository provider for injecting `AuthorisationService`
- Created a `PendingRequestsCountCubit` for displaying pending enrollment requests
- Also updated the logic in the App Bar as to when to show the app bar actions

**- How to verify it**
1. Onboard an atsign with manager keys
2. Navigate to the new authorisation page and note down OTP (or generate a custom pin)
3. Send an enrollment request

E.g.
```bash
at_activate enroll \
-p app-name-demo-1 \
-d device-name-demo-1 \
-n namespace1:r,namespace2:rw \
-a @<your-atisng> \
-k test1 \
-s 123123
```

4. Accept/Reject request coming into the app
5. Confirm the "new" device has successfully been enrolled or rejected 

Added APKAM enrollment authorisation functionality into the app

## Demo
https://github.com/user-attachments/assets/17077446-b43f-4816-ae81-e51f89529b97


packages/dart/npt_flutter/lib/app.dart
```diff
@@ -32,6 +34,9 @@ class App extends StatelessWidget {
         RepositoryProvider<FavoriteRepository>(
           create: (_) => FavoriteRepository(),
         ),
+        RepositoryProvider<AuthorisationService>(
+          create: (_) => AuthorisationService(AtClientManager.getInstance().atClient)..init(),
```
xavierchanth: I don't think passing the atClient like this works across changing atSigns

packages/dart/npt_flutter/lib/features/authorisation/cubit/pending_requests_count_cubit.dart
```diff
@@ -0,0 +1,28 @@
+import 'dart:async';
+
+import 'package:at_client_mobile/at_client_mobile.dart';
+import 'package:flutter_bloc/flutter_bloc.dart';
+
+class PendingRequestsCountCubit extends Cubit<int> {
```
xavierchanth: Please extend the LoggingCubit class instead of Cubit, this way the built-in logging functionality will automatically work with this state

xavierchanth: The caveat to using LoggingCubit is that it requires pure state, which makes it harder to implement certain things. But this is generally a good thing, since it prevents side-effects in the app.

Zambrella: Sure I can do this.

Is there a reason to not use `BlocObserver` instead - https://bloclibrary.dev/bloc-concepts/#observing-a-cubit

xavierchanth: I didn't know it existed, it may not have when the app was first built

packages/dart/npt_flutter/lib/app.dart
Zambrella: You're right, this doesn't work when switching atSigns. Will fix.

xavierchanth (dismissed):

xavierchanth (approved):

PR #1674, opened 2025-01-21
fix: npt refresh problems resolved.

### LINKED ISSUE #1598: Refreshing a connection that is already connected leads to a bad state
### Describe the bug

When clicking to "Refresh" a connected profile it changes to Disconnected and when trying to connect again throws an error: `SocketException: Failed to create server socket (OS Error: The shared flag to bind() needs to be `true` if binding multiple times on the same (address, port) combination.)`

### Steps to reproduce

1. Connect to a profile
2. Click on the menu `...` for the profile
3. Select Refresh
4. Reconnect

### Expected behavior

I would expect that refreshing an already connected profile would maintain the connection

### Screenshots

_No response_

### Smartphones

_No response_

NoPorts

### Additional context

Does the refresh do anything? Should it be removed?

### LINKED ISSUE #1615: NoPorts Desktop (Refresh button missing if no profiles present)
### Describe the bug

If you install the app on a new machine no profiles appear and the refresh button is missing in the UI. But if you add a profile then the refresh button appears and you see the profiles you had set up.

### Steps to reproduce

1. Install
2. No profiles seen
3. Create a profile
4. refresh button appears
5. hit refresh then dlete the dummy profile !

### Expected behavior

Profiles should appear/refesh button should be visible from install

### Screenshots

_No response_

### Smartphones

- Windows on ARM

_No response_

### Additional context

_No response_

CurtlyCritchlow:
**- What I did**

1. Fixed the text selection issue
2. Fixed focus node being called after it is disposed
3. Fixed text field not showing correctly.
4. "@" automatically added to atsign fields.
5. Fixed error message being cut off in the UI.
6. Settings List items reorganised
7. refresh buttons removed from profiles and popup menu item
8. Sync progress listener added to update ui when sync completes and profile list is empty.
9. Localisations updated for new strings.

**- How I did it**

**- How to verify it**


packages/dart/npt_flutter/lib/features/onboarding/util/profile_progress_listener.dart
```diff
@@ -0,0 +1,20 @@
...
+import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:npt_flutter/app.dart';
+import 'package:npt_flutter/features/profile_list/bloc/profile_list_bloc.dart';
+
+class ProfileProgressListener extends SyncProgressListener {
+  @override
+  void onSyncProgressEvent(SyncProgress syncProgress) {
+    final profileListBlock = App.navState.currentContext!.read<ProfileListBloc>();
+
+    final profileListState = profileListBlock.state as ProfileListLoaded;
```
xavierchanth: There's no checks in place to confirm that profileListBlock.state is actually the ProfileListLoaded type, what if it's ProfileListInitial, then it throws an uncaught exception.


The if below should be something like 

```dart
... SyncStatus.success && (profileListBlock.state is! ProfileListLoaded || (profileListBlock.state as ProfileListLoaded).profiles.isEmpty)
```

packages/dart/npt_flutter/lib/features/settings/view/settings_view.dart
```diff
@@ -43,13 +43,13 @@ class SettingsView extends StatelessWidget {
...
-                          CustomTextButton.email(),
+                          gapH10,
+                          CustomTextButton.backUpYourKey(),
                           CustomTextButton.faq(),
-                          CustomTextButton.privacyPolicy(),
+                          CustomTextButton.email(),
+                          CustomTextButton.discord(),
                           CustomTextButton.feedback(),
-                          CustomTextButton.backUpYourKey(),
+                          CustomTextButton.privacyPolicy(),
```
xavierchanth, line 52: Don't rearrange this unnecessarily, it makes the videos and screenshots on the docs site out of date.

packages/dart/npt_flutter/lib/features/settings/widgets/contact_list_tile.dart
```diff
@@ -21,21 +21,23 @@ class ContactListTile extends StatelessWidget {
...
-                    leading: CircleAvatar(
-                      radius: Sizes.p18.toFont,
-                      backgroundColor: AppColor.primaryColor,
-                      backgroundImage: snapshot.data!['image'] != null ? MemoryImage(snapshot.data!['image']) : null,
-                    ),
+                    // leading: CircleAvatar(
+                    //   radius: Sizes.p18.toFont,
+                    //   backgroundColor: AppColor.primaryColor,
+                    //   backgroundImage: snapshot.data!['image'] != null ? MemoryImage(snapshot.data!['image']) : null,
+                    // ),
```
xavierchanth: nit: delete code > commenting it.

packages/dart/npt_flutter/lib/util/general_extensions.dart
```diff
@@ -0,0 +1,12 @@
+extension StringExtension on String {
+  String atsignify() {
+    var value = this;
+    if (!startsWith('@')) {
+      value = '@$this';
+    }
+    if (endsWith(' ')) {
+      value = trim();
```
xavierchanth: You should always trim and do it before checking startsWith, in case there is leading whitespace.

xavierchanth (changes requested):
Please address the comments above

packages/dart/npt_flutter/lib/features/onboarding/util/profile_progress_listener.dart
CurtlyCritchlow: My worry that `profileListBlock.state is! ProfileListLoaded` will refresh the UI while a profile is connected which will trigger the same issue Doug observed that made us remove the profileListRefreshButton.

I want `...syncStatus.success && profileListBlock.state as ProfileListLoaded).profiles.isEmpty` because I only want to refresh the UI after sync completes but profiles were missing due to local and remote secondary being out of sync. I think this condition occurs when atsign is used for the first time or when using the same atsign is used on windows and Mac versions of noPortsDesktop.

xavierchanth: Then change `is!` to `is` and the following `||` to `&&`. Either way, you need to check the type before casting.

packages/dart/npt_flutter/lib/features/settings/view/settings_view.dart
CurtlyCritchlow: @tinashe404 asked me to rearrange it since this order makes more sense logically. We'll stick with this arrangement going forward.

xavierchanth (dismissed):

xavierchanth (approved):

PR #1698, opened 2025-01-27
feat(breaking): at_activate interface for C

xavierchanth:
**- What I did**

- uptake breaking changes from atsdk

**- How I did it**

**- How to verify it**

feat(breaking): at_activate interface for C

### DIFF
--- a/packages/c/sshnpd/CHANGELOG.md
+++ b/packages/c/sshnpd/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.4.0
+
+- breaking: reimplement at_activate to conform to dart interface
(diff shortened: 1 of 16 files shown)


packages/c/sshnpd/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.3.5
```
cpswan: Since we're (breaking) we should probably make this 0.4.0

cpswan (changes requested):
Can you also please bump version in CMakeLists.txt

cpswan (approved):

PR #1706, opened 2025-01-29
fix: sync notice to user updated to be non-blocking for them.

CurtlyCritchlow:
**- What I did**

1. Updated Sync to be non-blocking for the user
2. Prevent the user from entering a badly formatted atsign when onboarding
3. Set the same column size of the profile and status column of the profiles minimum view.

**- How I did it**

**- How to verify it**

### DIFF
--- a/packages/dart/npt_flutter/lib/features/onboarding/util/profile_progress_listener.dart
+++ b/packages/dart/npt_flutter/lib/features/onboarding/util/profile_progress_listener.dart
@@ -5,9 +5,12 @@ import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:npt_flutter/app.dart';
 import 'package:npt_flutter/features/profile_list/bloc/profile_list_bloc.dart';
 
+import '../../profile_list/cubit/sync_cubit.dart';
+
 class ProfileProgressListener extends SyncProgressListener {
   @override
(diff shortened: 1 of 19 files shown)


packages/dart/npt_flutter/lib/features/onboarding/util/profile_progress_listener.dart
```diff
@@ -5,16 +5,20 @@ import 'package:flutter_bloc/flutter_bloc.dart';
...
+  void onSyncProgressEvent(SyncProgress syncProgress) async {
+    await App.navState.currentContext!.read<SyncCubit>().checkSync();
     final profileListBlock = App.navState.currentContext!.read<ProfileListBloc>();
 
     if (syncProgress.syncStatus == SyncStatus.success &&
         (profileListBlock.state is ProfileListLoaded &&
             (profileListBlock.state as ProfileListLoaded).profiles.isEmpty)) {
       profileListBlock.add(const ProfileListLoadEvent());
       log('ProfileProgressListener: ProfileListLoadEvent triggered to reload profiles');
+      await App.navState.currentContext!.read<SyncCubit>().checkSync();
```
xavierchanth: use `unawaited()` instead of `await` to let the job get handled asynchronously. Especially for the one at the top, it will ensure the profiles are loaded as quickly as possible.

xavierchanth (approved):

PR #1726, opened 2025-02-06
feat: C memcheck tools

xavierchanth:
**- What I did**

- Added a justfile with some useful commands
  - basic configure & build commands
  - valgrind container commands
  - memcheck command
    - runs sshnpd in valgrind and logs to a file
      (using `date +%s` to make each file unique)
- Added a new tool which sends a graceful_shutdown command to sshnpd
  - The notification is ignored by sshnpd by default and only enabled when the
    `SSHNPD_ENABLE_TESTING_SHUTDOWN_NOTIFICATION` compile definition is added
  - `just dev` has this definition on, but any other build (as of this point) will not

Note: The graceful shutdown tool is needed because sshnpd runs indefinitely, and SIGINT doesn't follow through on cleaning up memory properly. We previously tried to implement this, but on some platforms and in systemd, it led to the binary hanging.

**- How I did it**

**- How to verify it**

- I've already used it to confirm that the new atsdk version results in no idle leaks.
- There are leaks caused by active session creation which I will tackle in another PR

feat: C memcheck tools


cpswan (changes requested):
LGTM. Can you please bump versions so we can do a release. Either 0.4.3 or 0.5.0 depending on whether you think the graceful shutdown tool is a new feature.

xavierchanth:
> LGTM. Can you please bump versions so we can do a release. Either 0.4.3 or 0.5.0 depending on whether you think the graceful shutdown tool is a new feature.

For development purposes only, so I wouldn't consider it a new feature.

cpswan (approved):

PR #1741, opened 2025-02-13
chore: Remove Python package

cpswan:
Now that C is at 1.0.0 the Python package can be deprecated

**- What I did**

Removed Python package and associated workflows

**- How to verify it**

NA

**- Description for the changelog**

chore: Remove Python package

PR #1742, opened 2025-02-13
Revert "chore: Remove Python package"

cpswan:
Reverts atsign-foundation/noports#1741 as PyPI recommends publsihing a final version with a note in docs before archiving a package


cpswan:
Nice to see dependency review catching a vulnerable package :)

My plan here is to bump the package to take the new at_python SDK with cryptography 44.0.1, and the note about archiving and use C instead. Publish one last time, and then we can remove Python again.

xavierchanth (approved):

PR #1757, opened 2025-02-24
feat: Dockerfiles for e2e tests next iteration

### LINKED ISSUE #1444: e2e tests next iteration
### Note
**This should be a P1 once the next big NoPorts release is out the door; we can survive on what we've got until then**

Our needs have outgrown what our current e2e_all test pack can deliver

Lots of discussion over the past several months on this. Need to use standard infrastructure and bite the bullet on parallelization
- separate compile phase from test phase
- run clients and daemons on different 'hosts' (docker containers / VMs)
- run test packs in parallel
  - (a) parallel per daemon, all client tests
  - (b) parallel per client test, all daemons
- run test packs on multiple platforms

Best approach is likely a combination of features of both the current e2e_all test pack (key benefits: very quick to run tests locally; very easy to add new tests and have test variety) and the old end2end_tests pack (key benefit: everything dockerized)

_No response_

JeremyTubongbanua:
closes #1444 

**- What I did**
- Created 6 Dockerfiles, meant to run a container that holds either C binaries or Dart binaries, that are either built from a branch (or commit hash) OR taken from releases.

The 6 Dockerfiles are:

- **Dockerfile.c.branch** - build C binaries (sshnpd, srv, at_activate) from a specified branch name or commit hash 
- **Dockerfile.c.current** - builds C binaries locally 
- **Dockerfile.c.release** - download C binaries from a specified `c.*.*.*` release
- **Dockerfile.dart.branch** - build Dart binaries from a specified branch name or commit hash
- **Dockerfile.dart.current** - builds Dart binaries locally
- **Dockerfile.dart.release** - download Dart binaries from the latest release or a specified `v.*.*.*` release

In `tests/e2e_all/scripts/`:
- Created `setup_atkeys.sh` which creates a `runtime/<commit_id>/keys/` folder which will hold the client master atKeys, the client APKAM atKeys, the daemon master atKeys, and the daemon APKAM atKeys. This script only copies the master atKeys into the folder, meanwhile the APKAM atKeys will be `cp`'d by the `apkam_setup.sh` script
- In apkam_setup.sh, we actually `cp` (copy) the keys into `runtime/<commit_id>/keys/`
- New `build_docker_daemons.sh` script which builds a docker daemon for each $daemonVersion
- Modified `start_daemons.sh` to run  Docker daemon containers, be sure to call `build_docker_daemons.sh` first
- Modified `stop_daemons.sh` to stop Docker containers that start with `e2e_all-*`
- New `common/wipe_known_hosts.sh` to reduce the chance of ssh'ing into the same localhost:<port> which would have been added to the known_hosts file. We call this script at the beginning of main.sh and the beginning of every test.
- Modified `setup_binaries.sh` to setup binaries in parallel if `-p` is used
- Fixed a typo error in `tests/001_minus_s_flag`
- Added `.github/workflows/dockerhub_e2e_all_base_runtime.yml` to push e2e_all base image to Dockerhub
- Made it so that `setup_binaries.sh` and `build_docker_daemons.sh` in `main.sh` run in parallel

**- How I did it**
- CI passes

**- How to verify it**
- See README.md on usage of the Dockerfiles

feat: Dockerfiles for e2e tests next iteration


JeremyTubongbanua:
This PR is related to #1444

tests/e2e_all/dockerfiles/daemon/Dockerfile.build.branch.dart.noports
```diff
@@ -0,0 +1,31 @@
+FROM dart:3.6.0
```
cpswan: Should be 3.7.0 and pinned to a hash

JeremyTubongbanua: Where can I find instructions on how to do that again?

JeremyTubongbanua: Never mind , figured it out :)

tests/e2e_all/dockerfiles/Dockerfile.c.branch
```diff
@@ -0,0 +1,58 @@
...
+RUN set -eux ; \
+    apt-get update ; \
+    apt-get install -y openssh-server sudo vim nano iproute2 nmap tmux curl cron ; \
+    groupadd --gid ${GROUP_ID} ${USER} ; \
+    useradd --system --shell /bin/bash --home ${HOMEDIR} --uid ${USER_ID} --gid ${GROUP_ID} ${USER} ; \
+    usermod -aG sudo ${USER} ; \
+    mkdir -p ${HOMEDIR}/.ssh ${HOMEDIR}/.atsign/keys ; \
+    chown -R ${USER}:${USER} ${HOMEDIR} ; \
+    touch ${HOMEDIR}/.ssh/authorized_keys ; \
+    ex +"%s/^%sudo.*$/%sudo ALL=(ALL:ALL) NOPASSWD:ALL/g" -scwq! /etc/sudoers ; \
```
xavierchanth: why a vim ex command instead of sed?

xavierchanth: This should be the same as `sed -e 's/^%sudo.*$/%sudo ALL=(ALL:ALL) NOPASSWD:ALL/g' -i /etc/sudoers` and you don't need to install vim & the vim runtime (which is heavier than just vi)

xavierchanth (dismissed):
See my comment

tests/e2e_all/dockerfiles/Dockerfile.c.branch
JeremyTubongbanua: Done in https://github.com/atsign-foundation/noports/pull/1757/commits/64fcfeef074a17bcbc08ed0a7f3af20597744b74

JeremyTubongbanua:
Removing request for reviewers for now

JeremyTubongbanua:
CI Tests are now passing

I have updated the PR description just now.

gkc:
@JeremyTubongbanua Great work! Are you going to add parallelization in this PR? My feeling is that you should, rather than merge this PR now, since the PR as it stands increases the setup time without reducing the test execution time.

JeremyTubongbanua:
I agree parallelization should be done in this PR. The tests take longer right now cause of docker building

cpswan:
It might also be worth looking at whether we can create a base image that we rebuild periodically, and then use that to minimise build times.
Take a look at what we do with ve_base for at_server for an example of how that works already.

JeremyTubongbanua:
Managed to get test execution to work in parallel using regular bash `&`

e2e_all CI now takes ~11min30s

Now I think my next step is to reduce docker build time

gkc:
> Managed to get test execution to work in parallel using regular bash `&`
> 
> e2e_all CI now takes ~11min30s
> 
> Now I think my next step is to reduce docker build time

Nice! The actual test execution is only taking just over two minutes now

JeremyTubongbanua:
> It might also be worth looking at whether we can create a base image that we rebuild periodically, and then use that to minimise build times. Take a look at what we do with ve_base for at_server for an example of how that works already.

Would that be a base image for the _buildimage_ step or the __runtime__ step?

This is what one of my Dockerfiles look like:

```Dockerfile
FROM dart:3.7.3@sha256:1d0491432892b4f7cd216dfaf8406adef5bff216adc51a024e381023d8f22ce7 AS buildimage

ENV REPO_DIR=/app/repo
ENV OUTPUT_DIR=/output

COPY . /app/repo

RUN set -eux ; \
    mkdir -p ${REPO_DIR} ${OUTPUT_DIR} ; \
    cd ${REPO_DIR}/packages/dart/sshnoports ; \
    dart pub get ; \
    dart compile exe bin/activate_cli.dart -o ${OUTPUT_DIR}/at_activate ; \
    dart compile exe bin/npp_atserver.dart -o ${OUTPUT_DIR}/npp_atserver ; \
    dart compile exe bin/npp_file.dart -o ${OUTPUT_DIR}/npp_file ; \
    dart compile exe bin/npt.dart -o ${OUTPUT_DIR}/npt ; \
    dart compile exe bin/sshnp.dart -o ${OUTPUT_DIR}/sshnp ; \
    dart compile exe bin/sshnpd.dart -o ${OUTPUT_DIR}/sshnpd ; \
    dart compile exe bin/srv.dart -o ${OUTPUT_DIR}/srv ; \
    dart compile exe bin/srvd.dart -o ${OUTPUT_DIR}/srvd

FROM debian:stable-slim@sha256:5484adc33b4c352c5a9f4c4ae295fc49aed1cb54a7a0712a1b29674fb6f4f10f AS runtime

ENV USER=atsign
ENV HOMEDIR=/${USER}
ENV USER_ID=1024
ENV GROUP_ID=1024

COPY --from=buildimage /output/ /usr/local/bin/

RUN set -eux ; \
    apt-get update ; \
    apt-get install -y openssh-server sudo; \
    groupadd --gid ${GROUP_ID} ${USER} ; \
    useradd --system --shell /bin/bash --home ${HOMEDIR} --uid ${USER_ID} --gid ${GROUP_ID} ${USER} ; \
    usermod -aG sudo ${USER} ; \
    mkdir -p ${HOMEDIR}/.ssh ${HOMEDIR}/.atsign/keys ; \
    touch ${HOMEDIR}/.ssh/authorized_keys ; \
    ssh-keygen -t ed25519 -a 100 -f ${HOMEDIR}/.ssh/id_ed25519 -q -N '' ; \
    cat ${HOMEDIR}/.ssh/id_ed25519.pub >> ${HOMEDIR}/.ssh/authorized_keys ; \
    chmod 700 ${HOMEDIR}/.ssh ; \
    chmod 644 ${HOMEDIR}/.ssh/authorized_keys ; \
    chmod 600 ${HOMEDIR}/.ssh/id_ed25519 ; \
    chmod 644 ${HOMEDIR}/.ssh/id_ed25519.pub ; \
    sed -i '$ a %sudo ALL=(ALL:ALL) NOPASSWD:ALL' /etc/sudoers ; \
    sed -E -i 's|^#?(PasswordAuthentication)\s.*|\1 no|' /etc/ssh/sshd_config ; \
    sed -i 's/#ListenAddress 0.0.0.0/ListenAddress localhost/g' /etc/ssh/sshd_config ; \
    chown -R ${USER}:${USER} ${HOMEDIR} ;

USER ${USER}
WORKDIR ${HOMEDIR}

CMD sudo service ssh start
```

Would by `runtime` image step already be my "base image"?

cpswan:
> Would that be a base image for the _buildimage_ step or the **runtime** step?

The buildimage step doesn't waste much time and gets straight on with compiling, but the runtime step spends some time doing `apt` stuff that won't change much from run to run (or even week to week).

JeremyTubongbanua:
Managed to get the e2e tests to run in 8min21s

Setup Binaries ~1min
APKAM Setup ~1min
Daemon Docker build and run ~3.5min
Client parallel test execution ~3min

JeremyTubongbanua:
Managed to drop it down to 7min45s by parallelizing more optimally by changing the way I do the for loop:

In `run_tests.sh`:
```bash
# 2. Run the rest of the tests in parallel for all client and daemon versions
listOfPids=()
for clientVersion in $clientVersions; do
  for testToRun in $testsToRun; do
    for daemonVersion in $daemonVersions; do
      if [ "$testToRun" == "001_minus_s_flag" ]; then
        # Skip this test because it was already run above
        continue
      fi
      "$testScriptsDir/common/run_single_test.sh" $clientVersion $daemonVersion $testToRun $timeoutDuration
    done &
    pid=$!
    listOfPids+=($pid)
    sleep 0.1
  done
  # Wait for all the tests to finish
  for pid in "${listOfPids[@]}"; do
    wait $pid
  done
done
```

JeremyTubongbanua:
`setup_binaries.sh` and `build_docker_daemons.sh` now run in parallel at the very beginning. I figured out which steps depend on each other and now independent steps in parallel (see diagram below)

image

Just ran e2e tests 3 times and it took:
1. 7min27s
2. 8min24s
3. 8min5s

JeremyTubongbanua:
World record of 6min47s - https://github.com/atsign-foundation/noports/actions/runs/15169885049/job/42657488288

JeremyTubongbanua:
Another one at 6min33s - https://github.com/atsign-foundation/noports/actions/runs/15170177518/job/42657982460?pr=1757

JeremyTubongbanua:
Ready for review

I have updated the PR description

JeremyTubongbanua:
We can also wait on merging this PR and do a couple more things like:
- setting up repeat Docker images on Dockerhub so that we don't build them every time
- make the e2e tests run in the GHA runner instead of in the VM

One thing to note is the repeat docker images cache, so they aren't really "blocking" the other processes that depend on them. So I think that we will not see significant time savings by putting the Docker images in Dockerhub, but it will be a good storage saver.

cpswan (commented):
Can the various almost identical Dockerfiles be parameterised so we only have one of each?

JeremyTubongbanua:
> Can the various almost identical Dockerfiles be parameterised so we only have one of each?

Will try that. Thank you!

JeremyTubongbanua:
Something broke; going back to draft

JeremyTubongbanua:
I got it working in GHA. I ran it in the GHA runner 3 times and the CI/CD vm 3 times and here are the results:

Attempt 1:
- GHA Runner: ~6min 49s
- VM: ~6min 35s

Attempt 2:
- GHA Runner: ~6min 58s
- VM: ~7min 6s

Attempt 3:
- GHA Runner: ~6min 40s
- VM: Flaked

I think the play here is to move it to the GHA runner. It takes maybe about ~10 extra seconds than the VM and it really depends on how the VM is feeling. The GHA runner would actually be running in ~5min but it takes an extra 1min30s to download `expect`. The GHA runner will also be a little bit faster or equal to the VM time once we get the base image on Dockerhub. Another key benefit is we can run multiple GHA runners in parallel, so one PR doesn't have to wait for the other PR's e2e test to finish.

JeremyTubongbanua:
In another PR, I will:

- Get the `atsigncompany/noports_e2e_all_base_runtime` base image on Dockerhub using the workflow file I made in this PR
- Setup virtual environment to use VE atSigns instead of prod

JeremyTubongbanua:
Ready for review - e2e_all just ran in 5min45s in latest commit in GHA runner!

gkc (changes requested):
(UPDATE: I've fixed this locally, will push to branch shortly)

The new tests aren't respecting main.sh's `-r vip.ve.atsign.zone` option - I tend to run everything I need locally - e.g. atDirectory, atServers, relay(s) so this is important for me.

gkc:
> The new tests aren't respecting main.sh's `-r vip.ve.atsign.zone` option - I tend to run everything I need locally - e.g. atDirectory, atServers, relay(s) so this is important for me.

In the interests of getting this into trunk as quickly as possible, I'll happily defer getting `-r` working to a follow-up PR

gkc:
UPDATE: ignore this, this was a function of my having color-coded output from grep commands. I've made a change to cater for this, will push to branch shortly

Minor nit, `docker logs` processes aren't being killed during cleanup
```
    0 27316     1   0 12:35pm ttys027    0:00.01 sudo docker logs -f e2e_all-7362b026dcf
    0 27328 27316   0 12:35pm ttys027    0:00.04 docker logs -f e2e_all-7362b026dcf
    0 27332 27328   0 12:35pm ttys027    0:00.03 /Applications/Docker.app/Contents/Resources/bin/com.docker.cli logs -f e2e_all-7362b026dcf
    0 27334     1   0 12:35pm ttys027    0:00.01 sudo docker logs -f e2e_all-7362b026dc
    0 27341 27334   0 12:35pm ttys027    0:00.04 docker logs -f e2e_all-7362b026dc
    0 27359 27341   0 12:35pm ttys027    0:00.03 /Applications/Docker.app/Contents/Resources/bin/com.docker.cli logs -f e2e_all-7362b026dc
```

gkc:
(UPDATE: I've fixed this locally, will push to branch shortly)

The `-u` option needs to be removed, and the atsign remoteUsername hard-coded, since (1) the ssh remoteUsername in the e2e tests always needs to be "atsign", and (2) the remoteUsername is still defaulted to `whoami` .

gkc:
@JeremyTubongbanua is there a way to (re)run tests while reusing docker container images that were previously built rather than building them again?

gkc:
> @JeremyTubongbanua is there a way to (re)run tests while reusing docker container images that were previously built rather than building them again?

Maybe we could add a new flag like `-n` (or just extend the purpose of `-n` so it neither rebuilds the binaries nor the docker containers if the git ref hasn't changed since last time)

gkc:
> Minor nit, `docker logs` processes aren't being killed during cleanup
> 
> ```
>     0 27316     1   0 12:35pm ttys027    0:00.01 sudo docker logs -f e2e_all-7362b026dcf
>     0 27328 27316   0 12:35pm ttys027    0:00.04 docker logs -f e2e_all-7362b026dcf
>     0 27332 27328   0 12:35pm ttys027    0:00.03 /Applications/Docker.app/Contents/Resources/bin/com.docker.cli logs -f e2e_all-7362b026dcf
>     0 27334     1   0 12:35pm ttys027    0:00.01 sudo docker logs -f e2e_all-7362b026dc
>     0 27341 27334   0 12:35pm ttys027    0:00.04 docker logs -f e2e_all-7362b026dc
>     0 27359 27341   0 12:35pm ttys027    0:00.03 /Applications/Docker.app/Contents/Resources/bin/com.docker.cli logs -f e2e_all-7362b026dc
> ```

Ignore this, this was a function of the docker containers not being stopped on my machine because of my grep color-coding. Will push a change for this

gkc (dismissed):

gkc (approved):
Yay!

PR #1765, opened 2025-02-26
build: use `npm ci` instead of `npm install` in build scripts

gkc:
**- What I did**
build: use `npm ci` instead of `npm install` in build scripts, so that the dependency isn't flagged as being unpinned

**- How I did it**
Modified npm_build_test.yml and multibuild.yml

**- How to verify it**
line no longer flagged as unpinned by the dependency checker

### DIFF
--- a/.github/workflows/multibuild.yaml
+++ b/.github/workflows/multibuild.yaml
@@ -228,9 +228,6 @@ jobs:
         uses: actions/setup-node@1d0ff469b7ec7b3cb9d8673fde0c81c44821de2a # v4.2.0
         with:
           node-version: '20.17.0'
-      - name: setup required npm version
-        run: |
-          npm ci npm@10.8.2
       - name: Build admin webapp
         working-directory: ./apps/admin/webapp
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/multibuild.yaml
```diff
@@ -230,7 +230,7 @@ jobs:
           node-version: '20.17.0'
       - name: setup required npm version
         run: |
-          npm install -g npm@10.8.2
+          npm ci npm@10.8.2
```
cpswan: I don't think this change will persuade the scanner, as it's still not pinned to a hash

Do we actually need `npm@10.8.2` stated on this line, or can it just be picked up from the lock file?

gkc: > I don't think this change will persuade the scanner, as it's still not pinned to a hash
> 
> Do we actually need `npm@10.8.2` stated on this line, or can it just be picked up from the lock file?

Using “npm ci” by definition uses the lock file; I made a similar change in the past to eliminate this issue for the package dependencies, so my understanding is that the scanner knows about “npm ci” vs “npm i” and therefore this change should work (?)

gkc: However - actually this line is irrelevant. Will remove

cpswan (approved):

Issue #1767, opened 2025-02-26
Segmentation fault on c sshnpd (v 1.0.0) when accessed via the Noports desktop app


cconstab:
### Describe the bug

Segmentation Faults :-(

### Steps to reproduce

1. First setup a C sshnpd
2. configure Noports desktop to access port 22
3. note in logs the Segmentation error

### Expected behavior

Just to work as teh commandline does

### Screenshots

```
/home/cconstab/.local/bin/sshnpd.sh: line 8:  4824 Segmentation fault      /home/cconstab/.local/bin/sshnpd -a @ssh_1 -m @cconstab -d csshnp01
permitting open:
localhost:22
localhost:3389
[INFO] 2025-02-26 17:15:16.775542 | atclient | Missing atServer host or port. Using root.atsign.org:64 atDirectory to find atServer address
```

cconstab:
@cpswan might be worth your time confirming this.. I think this might have been reported in the past but i could noty find the ticket

cpswan:
> [@cpswan](https://github.com/cpswan) might be worth your time confirming this.. I think this might have been reported in the past but i could noty find the ticket

@cconstab confirmed that I can replicate this.

I just set up a C daemon on an OpenWrt 24.10.0 VM and here's what I see in the logs as I try to connect from the desktop app:

```log
Sun Mar  2 15:03:33 2025 daemon.info sshnpd[3893]: [INFO] 2025-03-02 15:03:33.673995 | connection | Connected
Sun Mar  2 15:03:34 2025 daemon.info sshnpd[3893]: [INFO] 2025-03-02 15:03:34.585997 | atclient | Missing atServer host or port. Using root.atsign.org:64 atDirectory to find atServer address
Sun Mar  2 15:03:36 2025 daemon.info sshnpd[3893]: [INFO] 2025-03-02 15:03:36.026849 | connection | Connected
Sun Mar  2 15:03:37 2025 daemon.info sshnpd[3893]: @cpswan,
Sun Mar  2 15:03:37 2025 daemon.info sshnpd[3893]: [INFO] 2025-03-02 15:03:37.047227 | refresh_device_entry | Saving username entries for this device
Sun Mar  2 15:03:37 2025 daemon.info sshnpd[3893]: [ERROR] 2025-03-02 15:03:37.048523 | connection | recv is null. exiting
[...]

Issue #1758, opened 2025-02-24
C daemon hung (version 1.0.0)


cconstab:
### Describe the bug

Both musl and glibc dameons hung

### Steps to reproduce

Not sure

### Expected behavior

No hangs

### Screenshots

The 

### Smartphones

On the musl

```
[INFO] 2025-02-24 13:11:42.604644 | connection | Connected
[ERROR] 2025-02-24 13:11:42.990095 | connection | recv is null. exiting
[INFO] 2025-02-24 13:11:42.990106 | sshnpd | Reconnected the monitor connection.
[INFO] 2025-02-24 13:17:13.054074 | refresh_device_entry | Refreshing device info entries for this device
[ERROR] 2025-02-24 13:17:13.054092 | atclient_put_shared_key | ctx.atserver_connection is not started
[ERROR] 2025-02-24 13:17:13.054099 | atclient_put_shared_key | atclient_put_shared_key_validate_arguments: 1
[ERROR] 2025-02-24 13:17:13.054105 | refresh_device_entry | Failed to refresh device entry for @cconstab
[INFO] 2025-02-24 14:17:14.353251 | refresh_device_entry | Refreshing device info entries for this device
[ERROR] 2025-02-24 14:17:14.353267 | atclient_put_shared_key | ctx.atserver_connection is not started
[ERROR] 2025-02-24 14:17:14.353272 | atclient_put_shared_key | atclient_put_shared_key_validate_arguments: 1
[ERROR] 2025-02-24 14:17:14.353276 | refresh_device_entry | Failed to refresh device entry for @cconstab
[INFO] 2025-02-24 15:17:15.639845 | refresh_device_entry | Refreshing device info entries for this device
[ERROR] 2025-02-24 15:17:15.639862 | atclient_put_shared_key | ctx.atserver_connection is not started
[ERROR] 2025-02-24 15:17:15.639868 | atclient_put_shared_key | atclient_put_shared_key_validate_arguments: 1
[ERROR] 2025-02-24 15:17:15.639875 | refresh_device_entry | Failed to refresh device entry for @cconstab
```

glibc
```
ERROR] 2025-02-24 16:05:48.470179 | request_options | atclient_get_public_key_request_options_is_bypass_cache_initialized: Invalid arguments
[INFO] 2025-02-24 16:05:48.612226 | srv - run | Configuring encrypter/decrypter for srv
[INFO] 2025-02-24 16:05:48.612246 | srv - run | Starting socket to socket srv
[...]

PR #1785, opened 2025-03-07
refactor: synchronize update jobs

xavierchanth:
**- What I did**

See also: https://github.com/atsign-foundation/at_c/pull/568

- Synchronize the username keys (put once) and device_info keys (put once an hour - ttln 30 days) (one less potential source for hangs #1758)
- Remove pthread & pthread locks
  - NOTE: lib pthread still needed to build sshnpd because srv uses it
- Separated the main loop to its own file: daemon.c
- Supplementary global state utils in daemon_utils.c
- global state variables in daemon.h
- use atcommons_memlist to manage memory in main.c

**- How I did it**

**- How to verify it**

- Tested under valgrind
- Still need to confirm segfault on other platforms (for #1767)

refactor: synchronize update jobs


xavierchanth:
Draft because I need to run valgrind test scenarios before its ready

packages/c/justfile
```diff
@@ -51,10 +51,12 @@ strict-docker +ARGS='':
 memcheck +ARGS='': build-dev
   #!/usr/bin/env bash
   file="valgrind-sshnpd-$(date +%s).log"
+  # child-silient-after-fork=yes: ignore session forks, we don't care
   valgrind --log-file="$file" \
   --tool=memcheck \
   --leak-check=full \
   --show-leak-kinds=all \
+  --child-silent-after-fork=yes \
```
xavierchanth, line 59: When we leave a session open, this hides the leaks which we expect to occur on the forked process: https://github.com/Mbed-TLS/mbedtls/issues/9958

packages/c/sshnpd/include/sshnpd/daemon.h
```diff
@@ -0,0 +1,65 @@
...
+extern char *home_dir;
+extern atchops_rsa_key_private_key signingkey;
+extern bool is_child_process;
+
+extern volatile sig_atomic_t should_run;
+
+// device info
+extern size_t device_info_pos;
+extern time_t *device_info_last_sent;
+extern uint8_t device_info_attempts;
```
xavierchanth, line 46: global state variables shared by `main.c:main()` and `daemon.c:main_loop()`

packages/c/sshnpd/include/sshnpd/background_jobs.h
xavierchanth, line 1: Deleted file: replaced by daemon and device_info headers.

packages/c/sshnpd/include/sshnpd/handle_npt_request.h
```diff
@@ -2,9 +2,7 @@
 #define HANDLE_NPT_REQUEST_H
 #include "sshnpd/params.h"
 #include <atclient/monitor.h>
-#include <pthread.h>
```
xavierchanth, line 5: no more pthread required anywhere except srv package.

mutexes have been removed everywhere as well.

packages/c/sshnpd/src/background_jobs.c
xavierchanth, line 1: Deleted file : replaced with device_info.c

packages/c/sshnpd/src/daemon.c
```diff
@@ -0,0 +1,250 @@
...
+char *home_dir;
+atchops_rsa_key_private_key signingkey;
+bool is_child_process;
+
+volatile sig_atomic_t should_run;
+
+// device info
+size_t device_info_pos;
+time_t *device_info_last_sent;
+uint8_t device_info_attempts;
```
XavierChanth, line 57: initalizers for all global state defined in `daemon.h`

xavierchanth, line 57: extern global state variables defined here. If they are defined directly in the header then certain compilers will see it as a duplicated definition.

packages/c/sshnpd/src/daemon.c
```diff
@@ -0,0 +1,250 @@
...
+
+void main_loop() {
```
xavierchanth, line 59: Previously defined in `main.c`.

packages/c/sshnpd/src/daemon_utils.c
```diff
@@ -0,0 +1,97 @@
...
+#include <sshnpd/run_srv_process.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+#define LOGGER_TAG "daemon"
+
+int set_worker_hooks() {
```
xavierchanth, line 31: All functions here previously defined in `main.c`.

packages/c/sshnpd/src/device_info.c
```diff
@@ -0,0 +1,166 @@
+#include "atlogger/atlogger.h"
+#include "sshnpd/daemon.h"
+#include "sshnpd/params.h"
+#include <atclient/atclient.h>
+#include <sshnpd/device_info.h>
+#include <string.h>
+#include <time.h>
+
+int handle_username_keys(atclient *atclient, const char **atsigns, size_t num_atsigns, const char *username,
+                         const char *device_name, const char *device_atsign, bool make_visible) {
```
xavierchanth, line 10: previously defined in background_jobs.c as part of the background thread.

This puts or deletes (during startup of sshnpd) all of the unix username keys, depending on whether the hide flag was passed to the program.

packages/c/sshnpd/src/device_info.c
```diff
@@ -0,0 +1,166 @@
...
+        atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "Failed to delete username key for %s\n", atsign);
+        continue;
+      }
+    }
+  }
+
+  return ret;
+}
+
+void send_next_device_info(atclient *atclient, sshnpd_params *params) {
```
xavierchanth, line 81: Function which puts (or deletes if hide==true) the next device info key if an hour has passed since the last time it was put by the current process (not determined by atkey metadata).

xavierchanth: The downside to beware of is that each iteration of this is between monitor timeouts (5 seconds), so if there are a ton of manager atsigns, it will take `5s * number of managers` to fully update on startup.

packages/c/sshnpd/src/main.c
```diff
@@ -37,55 +33,9 @@
...
-static atclient_atkeys atkeys;
-static sshnpd_params params;
-static atclient monitor_ctx;
-static char *regex;
-static FILE *authkeys_file;
-static char *authkeys_filename;
-static char *ping_response;
-static char *home_dir;
-static atchops_rsa_key_private_key signingkey;
-static bool is_child_process = false;
```
xavierchanth, line 85: moved to daemon.c

packages/c/sshnpd/src/main.c
```diff
@@ -100,9 +50,30 @@ static void child_exit_handler(int sig) {
...
 int main(int argc, char **argv) {
   int res = 0;
-  int exit_res = 0;
+
+  // setup initial values for global variables
+  is_child_process = false;
+  should_run = 1;
+  device_info_pos = 0;
+  device_info_last_sent = NULL;
+  device_info_attempts = 0;
```
xavierchanth, line 69: initialize global state which was previously static

packages/c/sshnpd/src/main.c
```diff
@@ -100,9 +50,30 @@ static void child_exit_handler(int sig) {
...
+
+  // device info
+  struct atcommons_memlist memlist = atcommons_memlist_create(32);
+#define safe_memlist(x) res = x;
+  if (res > 0) {
+    atcommons_memlist_failure_free(&memlist);
+    return res;
+  }
```
xavierchanth, line 77: Now using atcommons_memlist to manage all of the setup & teardown of memory in the main process.

packages/c/sshnpd/src/main.c
```diff
@@ -116,6 +87,23 @@ int main(int argc, char **argv) {
...
   }
 
+  // explicitly pass free_fn here because it is okay for these params to be null sometimes
+  // normally this would be an error
+  res = atcommons_memlist_add(&memlist, params.manager_list, true, free_if_not_null);
+  // res won't overflow from summation as the function returns a max value of 2
+  res += atcommons_memlist_add(&memlist, params.permitopen_hosts, true, free_if_not_null);
+  res += atcommons_memlist_add(&memlist, params.permitopen_ports, true, free_if_not_null);
+  res += atcommons_memlist_add(&memlist, params.permitopen_str, true, free_if_not_null);
+  res += atcommons_memlist_add(&memlist, NULL, true, mbedtls_psa_crypto_free);
```
xavierchanth, line 98: These values can be null sometimes, so we pass the `free_if_not_null` function, otherwise we get a state error depending on the program inputs. Intended feature of atcommons_memlist_add to prevent bugs, you must be explicit as done here. Although in most cases it is treated as a noop,`free(NULL)` is undefined behavior.

packages/c/sshnpd/src/main.c
```diff
@@ -256,78 +270,68 @@ int main(int argc, char **argv) {
...
-  }
-
-  for (size_t i = 0; i < params.manager_list_len; i++) {
-    atclient_atkey_init(usernamekeys + i);
-  }
-
-  struct refresh_device_entry_params refresh_params = {
-      &worker, &atclient_lock, &refresh_cond, &params, ping_response, username, &should_run, infokeys, usernamekeys};
-  res = pthread_create(&refresh_tid, NULL, refresh_device_entry, (void *)&refresh_params);
+  res = handle_username_keys(&worker, (const char **)params.manager_list, params.manager_list_len, username,
```
xavierchanth, line 289: Instead of starting the background thread, we only need to handle_username_keys during initial startup now.

packages/c/sshnpd/src/main.c
```diff
@@ -338,351 +342,33 @@ int main(int argc, char **argv) {
...
-  atclient_atkeys_free(&atkeys);
-
-exit:
-  free(params.manager_list);
-  free(params.permitopen_hosts);
-  free(params.permitopen_ports);
-  free(params.permitopen_str);
-  mbedtls_psa_crypto_free();
-  exit(exit_res);
-}
```
xavierchanth, line 409: All of these messy labels & goto statements are now handled by a single function call, and we no longer have to deal with dynamic stack allocations being denied by the compiler due to them being after a goto statement (due to stack position becoming dependent on program state - even simple gotos like this have been proven to be ugly).

packages/c/sshnpd/src/main.c
```diff
@@ -338,351 +342,33 @@ int main(int argc, char **argv) {
...
-  }
-  return ret;
-}
-
-static int unlock_atclient(int ret) {
-  ret = pthread_mutex_unlock(&atclient_lock);
-  if (ret != 0) {
-    atlogger_log(LOGGER_TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "Failed to release atclient lock\n");
-    exit(1);
+    atcommons_memlist_success_free(&memlist);
```
xavierchanth, line 370: we are using success here to handle freeing the child process memory, since we can optionally not free certain things (like the mbedtls sockets which will get closed if we do) on a success free.

tests/e2e_all/scripts/main.sh
```diff
@@ -110,19 +110,19 @@ export commitId
 remoteUsername=$(whoami)
 identityFilename="$HOME/.ssh/e2e_all.${commitId}"
 
-daemonStartWait=15
+daemonStartWait=20
```
xavierchanth, line 113: Increased wait time, as the c daemon synchronously puts username keys before starting monitor now.

packages/c/.clangd
```diff
@@ -2,7 +2,7 @@ CompileFlags:
   # Use gcc in development for better diagnostics
   # & Ensure that compilation doesn't fail due to warnings
   Compiler: gcc
-  Add: [-Wno-error, -xc, -std=c99]
+  Add: [-Wno-error, -xc, -std=c99, -DATSDK_DEBUG_MODE]
```
xavierchanth, line 5: Enable the debug logger macro in at_c

packages/c/sshnpd/src/main.c
xavierchanth: Doing this is mostly pointless, but if we need to memcheck srv, there will be unavoidable leaks on that process, but at least we can minimize the number that occur, which makes it easier to find leaks that actually matter.

cpswan (approved):

PR #1817, opened 2025-03-19
fix: use cpack for c_release source tarballs

xavierchanth:
**- What I did**

This is copied from the work I did for packaging the atsdk, then modified as needed.
See also: https://github.com/atsign-foundation/at_c/pull/574

- Added presets to generate a source tarball
- Added preset to build from the source tarball with

```sh
cmake --workflow --preset from-package-source
```

openwrt specific flags (`-fPIC -fhonour-copts`) can use a custom workflow from the same source tar to build:
```sh
cmake --workflow --preset openwrt
```

**- How I did it**

**- How to verify it**

Tested by emptying build cache, disabled internet, then ran the workflow preset noted above.

fix: use cpack for c_release source tarballs

### DIFF
--- a/.github/workflows/c_release.yml
+++ b/.github/workflows/c_release.yml
@@ -159,10 +159,11 @@ jobs:
         working-directory: ./packages/c
         run: |
           cmake --workflow --preset package-source
+          mv build/release-static/noports-source.tar.gz ../../tarball/csshnpd-${{ github.ref_name }}.tar.gz
       - uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4.6.1
         with:
           name: csshnpd-src-${{github.ref_name}}-${{github.run_number}}-${{github.run_attempt}}
-          path: ./packages/c/build/release-static/noports-source.tar.gz
+          path: ./tarball/csshnpd-${{ github.ref_name }}.tar.gz
 
   github-release:
     if: startsWith(github.ref, 'refs/tags/c')
--- a/.github/workflows/c_release.yml
+++ b/.github/workflows/c_release.yml
@@ -159,11 +159,10 @@ jobs:
         working-directory: ./packages/c
         run: |
           cmake --workflow --preset package-source
-          mv build/release-static/noports-source.tar.gz ../../tarball/csshnpd-${{ github.ref_name }}.tar.gz
       - uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4.6.1
         with:
           name: csshnpd-src-${{github.ref_name}}-${{github.run_number}}-${{github.run_attempt}}
-          path: ./tarball/csshnpd-${{ github.ref_name }}.tar.gz
+          path: ./packages/c/build/release-static/noports-source.tar.gz
 
   github-release:
     if: startsWith(github.ref, 'refs/tags/c')
--- a/.github/workflows/c_release.yml
+++ b/.github/workflows/c_release.yml
@@ -79,13 +79,11 @@ jobs:
           mv build/at_activate .
       - if: ${{ matrix.os == 'macOS-13' || matrix.os == 'macos-14'}}
         name: Zip the build
-        run:
-          zip tarball/${{ matrix.output-name }}.zip sshnpd at_activate
+        run: zip tarball/${{ matrix.output-name }}.zip sshnpd at_activate
       - name: Upload archives
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         with:
-          name:
-            ${{ matrix.output-name }}_${{ matrix.compiler
+          name: ${{ matrix.output-name }}_${{ matrix.compiler
             }}-${{  github.ref_name
             }}-${{github.run_number}}-${{github.run_attempt}}
           path: ./packages/c/sshnpd/tarball
@@ -144,8 +142,7 @@ jobs:
       - name: Upload tarballs
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         with:
-          name:
-            ${{
+          name: ${{
             matrix.output-name  }}-${{  github.ref_name  }}-${{  github.run_number
             }}-${{  github.run_attempt  }}
           path: ./packages/c/tarballs/${{ matrix.output-name }}.tgz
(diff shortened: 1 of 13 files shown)


.github/workflows/c_release.yml
```diff
@@ -159,15 +156,13 @@ jobs:
...
-          mkdir tarball
-          cd ./packages
-          mv c csshnpd-${{ github.ref_name }}
-          tar -cvzf ../tarball/csshnpd-${{ github.ref_name }}.tar.gz csshnpd-${{ github.ref_name }}
+          cmake --workflow --preset package-source
       - uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4.6.1
         with:
           name: csshnpd-src-${{github.ref_name}}-${{github.run_number}}-${{github.run_attempt}}
-          path: ./tarball/csshnpd-${{ github.ref_name }}.tar.gz
+          path: ./packages/c/build/release-static/noports-source.tar.gz
```
cpswan: I'd like to preserve the naming convention in terms of the tarball filename, and also the contents within it.

So csshnpd-c1.0.4.tar.gz rather than noports-source.tar.gz

And csshnpd-c1.0.4 rather than noports-source

Whether we do that with CPack cleverness, or just some script here doesn't bother me.

packages/c/CMakeLists.txt
```diff
@@ -1,6 +1,19 @@
...
+option(NOPORTS_MBEDTLS_PATH "Local mbedtls path" OFF)
+option(NOPORTS_GS_TOOL "Include graceful-shutdown-tool" OFF)
+set(ATSDK_CJSON_PATH ${NOPORTS_CJSON_PATH})
+set(ATSDK_MBEDTLS_PATH ${NOPORTS_MBEDTLS_PATH})
+
 cmake_minimum_required(VERSION 3.19)
 set(CMAKE_C_STANDARD 99)
 cmake_policy(SET CMP0135 NEW)
-project(csshnpd VERSION 1.0.3 LANGUAGES C)
+project(noports VERSION 1.0.3 LANGUAGES C)
```
cpswan: This is going to be a 1.0.4 release, so let's bump versions

packages/c/CMakePresets.json
```diff
@@ -50,6 +50,81 @@
...
       "inherits": ["strict"],
       "generator": "Ninja"
+    },
+    {
+      "name": "release-static",
+      "binaryDir": "${sourceDir}/build/release-static",
+      "cacheVariables": {
+        "BUILD_SHARED_LIBS": "OFF",
+        "CMAKE_BUILD_TYPE": "Release",
+        "CMAKE_C_FLAGS": "-std=c99 -Wno-error -Wextra -Werror-implicit-function-declaration",
```
cpswan, line 60: Since we're defining C_FLAGS here we need to migrate them from the Makefile that's previously specified them. (Is there a way still to pass in additional C_FLAGS?)

Right now my OpenWrt Makefile has: `-fPIC -fhonour-copts -Wno-calloc-transposed-args -Wno-error -pthread -lrt`

cpswan: I'm not sure we need all of those flags, but the first two have been added for OpenWrt specific build issues.

cpswan (changes requested):

packages/c/CMakePresets.json
xavierchanth: I can add an openwrt specific preset

.github/workflows/c_release.yml
xavierchanth: I misunderstood, I thought the name var was renaming the artifact as it gets uploaded. I'll do a mv

packages/c/CMakePresets.json
```diff
@@ -50,6 +50,127 @@
...
+        "ENABLE_CJSON_TEST": "OFF",
+        "CJSON_OVERRIDE_BUILD_SHARED_LIBS": "ON",
+        "CJSON_BUILD_SHARED_LIBS": "OFF",
+        "BUILD_SHARED_AND_STATIC_LIBS": "OFF",
+        "ENABLE_TARGET_EXPORT": "OFF",
+        "ENABLE_CJSON_VERSION_SO": "OFF"
+      }
+    },
+    {
+      "name": "openwrt",
```
xavierchanth, line 100: new openwrt build, which can be used to generate the binaries from the source-package we push to the github release.

.github/workflows/c_release.yml
```diff
@@ -159,10 +159,11 @@ jobs:
         working-directory: ./packages/c
         run: |
           cmake --workflow --preset package-source
+          mv build/release-static/noports-source.tar.gz ../../tarball/csshnpd-${{ github.ref_name }}.tar.gz
```
cpswan: This deals with the tarball filename, but not the root directory name within it. We probably have to untar it, rename, then tar it up again.

packages/c/CMakePresets.json
```diff
@@ -79,6 +79,45 @@
         "NOPORTS_MBEDTLS_PATH": "deps/mbedtls-src",
         "NOPORTS_CJSON_PATH": "deps/cjson-src",
         "NOPORTS_ATSDK_PATH": "deps/atsdk-src"
```
cpswan: Missing , here?

```
CMake Error: Could not read presets from /home/chris/git/github.com/atsign-foundation/noports/packages/c:
JSON Parse Error: /home/chris/git/github.com/atsign-foundation/noports/packages/c/CMakePresets.json:
* Line 82, Column 9
  Missing ',' or '}' in object declaration
* Line 128, Column 6
  Extra non-whitespace after JSON value.
```

cpswan (changes requested):

packages/c/CMakePresets.json
xavierchanth: Yep, new commit with more changes

.github/workflows/c_release.yml
xavierchanth: Ah, maybe this is best handled in the cpack afterall then

packages/c/CMakePresets.json
xavierchanth: Also, is openwrt using any cmake to compile this? If so, which distribution, source or static?

cpswan: OpenWrt is using CMake. The git repo just comes with Makefiles that pull down the source, whilst the SDKs are a little more canned. It's with the latter that I've run into trouble with CMake binaries that won't fetch from GitHub.

.github/workflows/c_release.yml
xavierchanth: [76a8ae0](https://github.com/atsign-foundation/noports/pull/1817/commits/76a8ae05a768e450c018ec1cd8163122b703f576)

source package: csshnpd-c1.0.4.tar.gz
static package(currently unused): csshnpd-static-c1.0.4.tar.gz

xavierchanth:
This was some git weirdness... there was a merge conflict, but the conflict was my patch on base directly against base, which means the patchset should have just applied fine. I did mess up my local and do a reset to HEAD though, which likely means it's my fault. But I should have seen those conflicts on local first.

.github/workflows/c_release.yml
cpswan: Maybe I wasn't clear that both the tarball name and the root directory it contains need to follow the same convention

cpswan (changes requested):

xavierchanth:
Nevermind. There was a change, upload-artifact was updated. I just pushed another commit restoring the updated version SHA.

.github/workflows/c_release.yml
xavierchanth: I took the wrong line in conflict resolution

xavierchanth:
> Nevermind. There was a change, upload-artifact was updated. I just pushed another commit restoring the updated version SHA.

Same thing for upload artifact, restored the latest path.

cpswan (approved):

Issue #1821, opened 2025-03-23
srvd resilience can fail due to lack of file handles


cconstab:
### Describe the bug

If an srvd cannot get a file handle for a new socket it still grabs the mutex.

### Steps to reproduce

1. First run srvd
2. hammer the srvd with half open connections
3. eventually it will run out of file handles (def on prod at 1024)
4. srvd still grabs mutex then files to get socket handle

### Expected behavior

get the socket first then the mutex (if possible) or some way to release the mutex (nasty)

### Screenshots

```
SHOUT|2025-03-19 17:20:54.887268| srvd |😎 Will handle request from @cconstab; acquired mutex 9b8070c1-f4cd-411b-807b-4798fe73d5c7.session_mutexes.sshrvd@rv_am
Unhandled exception:
SocketException: Failed to create server socket (OS Error: Too many open files, errno = 24), address = 0.0.0.0, port = 0
#0      _NativeSocket.bind (dart:io-patch/socket_patch.dart:1218)
<asynchronous suspension>
#1      _RawServerSocket.bind.<anonymous closure> (dart:io-patch/socket_patch.dart:2157)
<asynchronous suspension>
#2      _ServerSocket.bind.<anonymous closure> (dart:io-patch/socket_patch.dart:2513)
<asynchronous suspension>
#3      SocketConnector.serverToServer (package:socket_connector/src/socket_connector.dart:320)
<asynchronous suspension>
#4      socketConnector (package:noports_core/src/srvd/socket_connector.dart:94)
<asynchronous suspension>
```

### Smartphones

- srvd 5.9.0

### Were you using an atApplication when the bug was found?

srvd

gkc:
It is the spawned isolate which binds the ports and then sends the ports back to the main isolate. The resolution would be for each relay to first spawn its isolate, and only try to acquire the mutex when it has heard back from the spawned isolate. If it acquires the mutex, great, respond to the noports client and all is well. If it does not acquire the mutex then it should kill the isolate.

gkc:
@cconstab I've got a fix ready to go for this; the relay now spawns its isolate but has a timeout when awaiting the port pair info. If it doesn't hear back from the spawned isolate within one second, it logs an exception and returns without trying to grab the mutex. One second should be more than enough to avoid having false negatives where the isolate spawns successfully but the main isolate times out while waiting for the spawned to send the port pair info.

gkc:
[...]

PR #1825, opened 2025-03-24
fix: Have relays bind ports before locking the response mutex

gkc:
**- What I did**
fix: Have the relays acquire their ports before attempting to lock the respond-to-client mutex

**- How I did it**
See commits

**- How to verify it**
See #1821

### DIFF
--- a/packages/dart/noports_core/lib/src/srvd/srvd_impl.dart
+++ b/packages/dart/noports_core/lib/src/srvd/srvd_impl.dart
@@ -152,9 +152,11 @@ class SrvdImpl implements Srvd {
 
     logger.shout('Acquiring port pair');
     PortPair ports;
+    // ignore: unused_local_variable
     Isolate spawned;
+    SendPort sendToSpawned;
     try {
-      (ports, spawned) = await _spawnSocketConnector(
+      (ports, spawned, sendToSpawned) = await _spawnSocketConnector(
         0,
         0,
         sessionParams,
@@ -188,7 +190,7 @@ class SrvdImpl implements Srvd {
       if (err.toString().toLowerCase().contains('immutable')) {
         logger.shout('🤷‍♂️ Will not handle request from ${notification.from}'
             '; did not acquire mutex $mutexKey');
-        spawned.kill();
+        sendToSpawned.send('kill');
       } else {
         logger.shout('Will not handle; did not acquire mutex $mutexKey : $err');
       }
@@ -234,7 +236,7 @@ class SrvdImpl implements Srvd {
   /// once the socketConnector has spawned and is ready to accept connections
   /// it sends back the port numbers to the main isolate
   /// then the port numbers are returned from this function
-  Future<(PortPair, Isolate)> _spawnSocketConnector(
+  Future<(PortPair, Isolate, SendPort)> _spawnSocketConnector(
     int portA,
     int portB,
     SrvdSessionParams srvdSessionParams,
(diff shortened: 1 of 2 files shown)


packages/dart/noports_core/lib/src/srvd/srvd_impl.dart
```diff
@@ -148,6 +150,22 @@ class SrvdImpl implements Srvd {
 
     logger.info('New session request params: $sessionParams');
 
+    logger.shout('Acquiring port pair');
```
gkc, line 153: Moved this code from where it was earlier, and wrapped with a try-catch

packages/dart/noports_core/lib/src/srvd/srvd_impl.dart
```diff
@@ -170,19 +188,13 @@ class SrvdImpl implements Srvd {
       if (err.toString().toLowerCase().contains('immutable')) {
         logger.shout('🤷‍♂️ Will not handle request from ${notification.from}'
             '; did not acquire mutex $mutexKey');
+        spawned.kill();
```
gkc: If we are not handling the request, then let's kill the spawned isolate so we don't unnecessarily hang on to file descriptors

packages/dart/noports_core/lib/src/srvd/srvd_impl.dart
```diff
@@ -241,17 +253,25 @@ class SrvdImpl implements Srvd {
...
-    unawaited(Isolate.spawn<ConnectorParams>(socketConnector, parameters));
+    Isolate spawned =
+        await Isolate.spawn<ConnectorParams>(socketConnector, parameters);
 
-    PortPair ports = await receivePort.first;
+    logger.info('Waiting for isolate to send its port pair info');
+    PortPair ports;
+    try {
+      ports =
+          await receivePort.first.timeout(Duration(milliseconds: timeoutMs));
```
gkc: Time out if port info not received within `timeoutMs`

packages/dart/noports_core/lib/src/srvd/srvd_impl.dart
```diff
@@ -241,17 +253,25 @@ class SrvdImpl implements Srvd {
...
+    } on TimeoutException catch (_) {
+      throw TimeoutException('Timed out after ${timeoutMs}ms');
+    }
 
     logger.info('Received ports $ports in main isolate'
         ' for session ${srvdSessionParams.sessionId}');
 
-    return ports;
+    return (ports, spawned);
```
gkc: Returning the spawned isolate as well as the ports, so that the calling code can kill the isolate if it is not needed

packages/dart/noports_core/lib/src/srvd/srvd_impl.dart
xavierchanth: Has this been checked to ensure that it properly releases the file descriptors?

xavierchanth (dismissed):
Just the one question about file descriptors, if the answer is yes, then LGTM.

packages/dart/noports_core/lib/src/srvd/srvd_impl.dart
gkc: Great question! No. I made an assumption. Will double-check.

gkc: @XavierChanth Indeed you were correct; the ports are still being LISTENed to after the isolate is killed, and I don't think they would ever get cleaned up while the srvd process is still running.

I've made a change so the main isolate, instead of calling `kill` on the spawned isolate, instead sends a control message to the spawned isolate, which then cleans itself up correctly.

I created 10 npt sessions with everything running locally, which should result in 30 new ports being bound : 10 times 1 for each side of the relay, and 1 for the npt local port. With the most recent changes this works as expected. 30 additional ports bound (but no more) and all cleaned up when the npt sessions have exited.
```
gary@gkc2019 ~ % date; lsof -PiTCP -sTCP:LISTEN | wc -l
Tue 25 Mar 2025 09:59:12 GMT
      28
gary@gkc2019 ~ % date; lsof -PiTCP -sTCP:LISTEN | wc -l
Tue 25 Mar 2025 09:59:26 GMT
      58
gary@gkc2019 ~ % date; lsof -PiTCP -sTCP:LISTEN | wc -l
Tue 25 Mar 2025 09:59:53 GMT
      28
```

xavierchanth (approved):

PR #1831, opened 2025-03-27
fix: only atsign in teapot state are required to back up keys.

### LINKED ISSUE #1819: Backup key not implemented correctly.
### Describe the bug

Backup keys dialog is shown if the user never backed up their keys from the dialog. This means old users will be required to backup their keys even if they already did in the past.

### Steps to reproduce

1. Onboard with an atsign you already backup
2. You will be show a back up dialog that cannot be dismissed unless 
3. And then '...'
4.

### Expected behavior

Backup key dialog should be shown only if atsign is not activated.

### Screenshots

_No response_

### Smartphones

- Device: [MacOS]

NoPorts Desktop

### Additional context

_No response_

CurtlyCritchlow:
**- What I did**
Modify the back up key implementation so only atsign in teapot are allowed to backup their key.
**- How I did it**

**- How to verify it**


xavierchanth (approved):

xavierchanth:
LGTM to me, the only nit is that the naming of the variables is opposite to what the bool represents.
But everything is functionally correct, so it's fine.

PR #1840, opened 2025-04-01
fix: universal sh

### LINKED ISSUE #1829: universal.sh installs 64bit binaries onto Raspberry Pi OS with 32bit userspace
### Describe the bug

There are versions of Raspberry Pi OS that ship with a 64bit kernel and 32bit userspace.

We detect those systems as 64bit, since they report `aarch64` when we do `uname -m`

But the 64bit binaries fail to find the 64bit glibc they're linked against, as the userspace only has 32bit libraries.

### Steps to reproduce

1. First I install Raspberry Pi OS that has 64bit kernel and 32bit userspace
2. Then I download and run `universal.sh`
3. And then I try to run a NoPorts binary such as `at_activate`, which reports `No such file or directory`

### Expected behavior

Universal installer correctly identifies the userland glibc and downloads correct tarball

### Additional context

On my system where I replicated this, I see:

```sh
$ at_activate
-bash: /home/chris/.local/bin/at_activate: No such file or directory
```

```sh
$ file $(which at_activate)
/home/chris/.local/bin/at_activate: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, for GNU/Linux 3.7.0, stripped
```

The installer has downloaded 64bit binaries

```sh
$ uname -a
Linux raspberrypi 6.1.21-v8+ #1642 SMP PREEMPT Mon Apr  3 17:24:16 BST 2023 aarch64 GNU/Linux
```

We have a 64bit kernel

```sh
$ dpkg --print-architecture
armhf
```

But a 32bit userspace

```sh
$ file $(which ls)
/usr/bin/ls: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-armhf.so.3, BuildID[sha1]=81004d065160807541b79235b23eea0e00a2d44e, for GNU/Linux 3.2.0, stripped
```

So e.g. `ls` is linked against a 32bit glibc

## Workaround:

Download the latest release tarball for 32bit arm:

```sh
wget https://github.com/atsign-foundation/noports/releases/download/v5.9.0/sshnp-linux-arm.tgz
```

Untar it:

```sh
tar -xvf sshnp-linux-arm.tgz
```

Copy the binaries into place:

```sh
cp sshnp/* .local/bin/
```

Optionally tidy up:

```sh
rm -r sshnp sshnp-linux-arm.tgz
```

### LINKED ISSUE #1830: universal.sh doesn't find NoPorts binaries if `.local/bin` isn't already on PATH
### Describe the bug

We use `.local/bin` for client installs but the .profile in a typical Debian derived distro will only add it to the PATH if it's there already.

```sh
# set PATH so it includes user's private bin if it exists
if [ -d "$HOME/.local/bin" ] ; then
    PATH="$HOME/.local/bin:$PATH"
fi
```

`universal.sh` will attempt to do activation, but won't find the binaries it's just installed, as they're not (yet) on the PATH

### Steps to reproduce

1. First I get a freshly installed Ubuntu system
2. Then I download and run universal.sh, choosing a client install
3. And then it fails to find the binaries that have just been installed as they're not yet on the path
4. If I restart my shell then `.profile` finds `~/.local/bin` and adds it to the PATH, and the NoPorts binaries can be found OK.

### Expected behavior

`universal.sh` should ensure that the install directory is on the PATH

### LINKED ISSUE #1836: remove np.sh from universal.sh script
- It confuses people during install
- There are better ways of using NoPorts since this was introduced

xavierchanth:
fixes https://github.com/atsign-foundation/noports/issues/1829
fixes https://github.com/atsign-foundation/noports/issues/1830
fixes https://github.com/atsign-foundation/noports/issues/1836

Also fixes device name constraints (no issue created) we relaxed the constraints, but this script was never updated, as stricter is fine, but less strict is not.

**- What I did**

- Arm binaries now check the userspace sizing before determining which binaries should be used. i.e. 64 bit kernel with 32 bit userspace will use arm not arm64.
- remove np.sh "magic script" from universal.sh install
- retain root ownership when installing to /usr/local/bin - don't chown it to the user
- Add ~/.local/bin/ to PATH in ~/.profile
- Relax device constraints to match where we've relaxed them elsewhere

**- How I did it**

**- How to verify it**

- Will be testing on several machines, install.sh also needs to be tested, making this a bit tricky.

fix: universal sh


xavierchanth:
TODO: update device name constraints

Will resume tomorrow by
- complete the TODO above
- testing

packages/dart/sshnoports/bundles/shell/install.sh
```diff
@@ -154,12 +154,13 @@ install_single_binary() {
...
       mkdir -p "$user_bin_dir"
-      chown -R $user:$user "$user_bin_dir" || chown -R $user "$user_bin_dir"
     fi
 
     if [ -f "$dest/$1" ]; then
       ln -sf "$dest/$1" "$user_bin_dir/$1"
-      chown $user:$user "$user_bin_dir/$1" || chown $user "$user_bin_dir/$1"
+      chown root:wheel "$dest/$1" 2>/dev/null || chown root:root "$dest/$1" 2>/dev/null || chown root "$dest/$1" 2>/dev/null
+      chmod o+x "$dest/$1"
+
```
xavierchanth, line 163: When running as root, root should own the binaries in /usr/local/bin and they should be made executable by others.

packages/dart/sshnoports/bundles/universal.sh
```diff
@@ -741,89 +758,21 @@ client() {
...
-  cp "$extract_path"/sshnp/magic/sshnp.sh "$magic_script"
-  chmod +x "$magic_script"
-  if is_root && [ -f "$bin_path"/np.sh ]; then
-    ln -sf "$bin_path"/np.sh "$user_bin_dir"/np.sh
-    if [ $verbose = true ]; then
-      echo "=> Linked $user_bin_dir/np.sh to $bin_path/np.sh"
-    fi
-  fi
-
-  # get the inputs for the magic script
```
xavierchanth, line 755: No more np.sh

packages/dart/sshnoports/bundles/universal.sh
```diff
@@ -741,89 +758,21 @@ client() {
...
+  echo "Please enter atSigns you would like activated"
+  # get the inputs for client & device atsign to do activation
   if [ -z "$client_atsign" ]; then
-    get_atsign "client"
+    get_atsign_manually_once "client"
```
xavierchanth, line 764: Make these auto activate atSign inputs optional.

packages/dart/sshnoports/bundles/universal.sh
```diff
@@ -741,89 +758,21 @@ client() {
...
-          else
-            devices="$devices,$device_name"
-          fi
-        fi
-      done
-    fi
+  if [ -n "$device_atsign" ] || [ -n "$client_atsign" ]; then
+    echo "Ensuring your atSigns are activated..."
+    validate_activation
   fi
```
xavierchanth, line 818: These inputs were only needed for np.sh.

packages/dart/sshnoports/bundles/universal.sh
```diff
@@ -910,58 +859,60 @@ device() {
...
     printf "Enter device name: "
     read -r device_name
-    if ! echo "$device_name" | grep -Eq '^[a-z][a-z0-9_]{0,14}$'; then
-      echo "Device name must be in snake case and max 15 characters"
+    # capitals are allowed here, sshnpd will convert to lowercase
+    if ! echo "$device_name" | grep -Eq '^[A-Za-z0-9_][A-Za-z0-9_-]{0,35}$'; then
+      echo "Device name must only contain alphanumeric characters, '-', or '_' and may not start with '-'."
+      echo "Device name may have a maximum length of 36 characters"
       device_name=""
     fi
```
xavierchanth, line 867: updated validation. We allow captials here, sshnpd already converts to lowercase.

packages/dart/sshnoports/bundles/universal.sh
```diff
@@ -648,13 +659,19 @@ suggest_sudo() {
...
+  echo "then press Ctrl+C and then run the following command:"
   echo
   echo "sh universal.sh --no-sudo"
   echo
-  exit 0
+  echo "NB: Issues with installation through --no-sudo is not covered by support."
+  echo "    It is not a reliable installation method and it is NOT RECOMMENDED."
+  echo
+  echo "Executing 'sudo $original_command':"
+  sudo $original_command
```
xavierchanth, line 673: Automatically re-execute the script with the same arguments as sudo

packages/dart/sshnoports/bundles/universal.sh
```diff
@@ -648,13 +659,19 @@ suggest_sudo() {
...
+  exit $?
```
xavierchanth, line 674: propagate the exit code from the sudo run

packages/dart/sshnoports/bundles/shell/install.sh
```diff
@@ -512,6 +513,20 @@ tmux() {
...
+  if ! [ -d "$user_home/.profile" ]; then
+    touch "$user_home/.profile"
+    chown $user:$user "$user_home/.profile" 2>/dev/null || chown $user "$user_home/.profile" 2>/dev/null
+  fi
+
+  echo 'Adding $HOME/.local/bin/ to the PATH in ~/.profile'
+  echo 'PATH="$PATH:$HOME/.local/bin"' >>"$user_home/.profile"
+}
+
```
xavierchanth, line 529: detect & add ~/.local/bin to PATH in ~/.profile if there is no entry detected.

packages/dart/sshnoports/bundles/shell/install.sh
```diff
@@ -535,10 +550,22 @@ main() {
...
+      add_home_local_bin_to_path
+      ;; # strips debug_ prefix from the command input
+    debug)
+      install_debug_binaries
+      add_home_local_bin_to_path
+      ;;
+    all)
+      install_all_binaries
+      add_home_local_bin_to_path
+      ;;
```
xavierchanth, line 571: Add ~/.local/bin to the PATH in every install flow.

packages/dart/sshnoports/bundles/shell/install.sh
```diff
@@ -114,7 +114,7 @@ usage() {
 setup_authorized_keys() {
   mkdir -p "$user_ssh_dir"
   touch "$user_ssh_dir/authorized_keys"
-  chown $user:$user "$user_ssh_dir/authorized_keys" || chown $user "$user_ssh_dir/authorized_keys"
+  chown $user:$user "$user_ssh_dir/authorized_keys" 2>/dev/null || chown $user "$user_ssh_dir/authorized_keys" 2>/dev/null
```
xavierchanth, line 117: suppress errors to chown in all cases, we try as many combinations as possible, we do our best to guess the group name, but it can be different on different platforms. Setting just the user ownership is good enough in most cases.

packages/dart/sshnoports/bundles/universal.sh
xavierchanth: np.sh was the "quick picker" script for sshnp sessions. But with both openssh and putty integrations documented, this script only offers the negative impact of a more challenging install experience.

gkc (approved):

PR #1882, opened 2025-04-28
feat: srvd: default `per-session-storage` to true

gkc:
**- What I did**
feat: srvd: default the `per-session-storage` command-line arg to true. (There is no current reason to use the same storage each time, and using ephemeral storage by default provides a level of insulation from potential local-storage-related bugs in the future.)

**- How I did it**
See commits

**- How to verify it**
Tests pass


gkc:
@cconstab I think this is a sensible change (and maintains the current prod behaviour) ...

cconstab (approved):

PR #1897, opened 2025-05-02
feat: Created Dockerfile for srvd for relay hosts and included the bu…

cconstab:
**- What I did**
Created a Dockerfile that builds a standalone srvd container
Created bash script to update the container if the image has been updated
**- How I did it**
Modified existing Dockerfile steps but to create a srvd container
**- How to verify it**
ran locally and tested for example
```
 docker run  -it -v C:\Users\colin\.atsign\keys:/atsign/.atsign/keys  --network host  cconstab/srvd -v -a "@stream" -i 192.168.1.19
 ```
Created srvd container

### DIFF
--- a/packages/dart/sshnoports/tools/Dockerfile.srvd
+++ b/packages/dart/sshnoports/tools/Dockerfile.srvd
@@ -1,5 +1,5 @@
 # Dockerfile.srvd
-# Build image for a containerized call of the at_activate binary
+# Build image for a containerized call of the srvd and srv binaries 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/dart/sshnoports/tools/Dockerfile.srvd [hunk 1]
```diff
@@ -0,0 +1,40 @@
+# Dockerfile.activate
```
cpswan: Comments left over from another file?

cpswan (changes requested):

packages/dart/sshnoports/tools/Dockerfile.srvd
cconstab: Thank you, good catch

cpswan (approved):

PR #1903, opened 2025-05-08
fix: Ensure at_activate is on PATH and that it writes to correct dir

cpswan:
Fixes re-opened #1830 

**- What I did**

* Ensure that at_activate binary is on PATH
* Write files to $user_home so they don't end up in /root/...

**- How I did it**

Iterative testing against a clean VM snapshot

**- How to verify it**

Run the universal.sh from this branch on a clean test VM

fix: Ensure at_activate is on PATH and that it writes to correct dir

### DIFF
--- a/packages/dart/sshnoports/bundles/universal.sh
+++ b/packages/dart/sshnoports/bundles/universal.sh
@@ -723,7 +723,7 @@ validate_activation() {
     if [ "$device_status" -eq 3 ]; then
       echo $device_output
     fi
-    at_activate onboard -a $device_atsign
+    at_activate onboard -a $device_atsign -k "${user_home}/.atsign/keys/${device_atsign}_key.atKeys"
   fi
 
   if [ "$client_status" -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


xavierchanth (approved):

packages/dart/sshnoports/bundles/universal.sh
```diff
@@ -755,6 +755,11 @@ client() {
   "$extract_path"/sshnp/install.sh -b "$bin_path" -u "$user" srv
   "$extract_path"/sshnp/install.sh -b "$bin_path" -u "$user" at_activate
 
+  # set PATH so it includes user's private bin if it exists
+  if [ -d "$user_home/.local/bin" ] ; then
+    PATH="$user_home/.local/bin:$PATH"
```
xavierchanth, line 760: Don't we need export here?

cpswan: no need for export as it's within the subshell for the script that we need this.

xavierchanth (approved):
accidental comment, please ignore

PR #1907, opened 2025-05-12
chore: Bump version in CMake files

cpswan:
c1.0.5 release didn't build correctly as the source tarball was name c1.0.4

**- What I did**

Bumped version in CMake files

**- How to verify it**

Re-release c1.0.5

chore: Bump version in CMake files


CurtlyCritchlow (approved):
lgtm

PR #1922, opened 2025-05-16
feat: added --root-domain url:portnumber leaving port 64 as default

cconstab:
**- What I did**
Added the option of adding a port number (e.g. 443) to the root-domain flag

**- How I did it**
Altered the atClient generation to include a root port from the URL and leave a default of 64
**- How to verify it**
Compiled and tested by hand for both sshnp and sshnpd

feat: added --root-domain url:portnumber leaving port 64 as default


cconstab:
npt needs work :-(

gkc:
Not required. All NoPorts Dart programs already work with an atServer proxy, by adding the --root-domain flag with proxy syntax e.g. `--root-domain proxy:my_atserver_proxy.my_domain:443`

PR #1934, opened 2025-05-22
feat: changes for sending a heartbeat over the npt control socket

gkc:
**- What I did**
- feat: changes for sending a heartbeat over the npt control socket. Heartbeats are intended to persuade zealous network intermediaries that the control socket is not in fact dead.
- fix: Also took care of some technical debt in the control socket message handling.

**- How I did it**
- See commits

**- How to verify it**
- Automated tests pass
- Manually tested for races etc by setting a heartbeat interval of 250ms for an npt session and then continually refreshing my web browser
- Manually tested with npt version 5.2.0 to verify backwards compatibility of control message handling

### DIFF
--- a/packages/dart/noports_core/lib/src/srv/srv_impl.dart
+++ b/packages/dart/noports_core/lib/src/srv/srv_impl.dart
@@ -13,6 +13,9 @@ import 'package:mutex/mutex.dart';
 import 'package:noports_core/srv.dart';
 import 'package:noports_core/sshnp.dart';
 import 'package:socket_connector/socket_connector.dart';
+import 'package:at_commons/at_commons.dart' as at_commons;
+
+const newLineCodeUnit = 10;
 
 @visibleForTesting
 class SrvImplExec implements Srv<Process> {
@@ -48,6 +51,9 @@ class SrvImplExec implements Srv<Process> {
   @override
   final Duration timeout;
 
+  @override
+  final Duration heartbeat;
+
   SrvImplExec(
     this.streamingHost,
     this.streamingPort, {
@@ -59,6 +65,7 @@ class SrvImplExec implements Srv<Process> {
     this.sessionIVString,
     required this.multi,
     required this.timeout,
+    required this.heartbeat,
   }) {
     if (localPort == null) {
       throw ArgumentError('localPort must be non-null');
@@ -92,6 +99,8 @@ class SrvImplExec implements Srv<Process> {
       localHost ?? 'localhost',
       '--timeout',
       timeout.inSeconds.toString(),
+      '--heartbeat',
+      heartbeat.inSeconds.toString(),
     ];
     if (multi) {
       rvArgs.add('--multi');
@@ -193,6 +202,9 @@ class SrvImplInline implements Srv<SSHSocket> {
   @override
   final Duration timeout;
 
+  @override
+  final Duration heartbeat;
+
   SrvImplInline(
     this.streamingHost,
     this.streamingPort, {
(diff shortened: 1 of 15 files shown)


packages/dart/noports_core/lib/src/srv/srv_impl.dart
```diff
@@ -801,11 +846,86 @@ class SrvImplDart implements Srv<SocketConnector> {
...
+          } else {
+            rcvBuffer.addByte(data[element]);
+            // Backwards compatibility for clients prior to 5.6.1
+            // IF we're at the LAST byte in the received data
+            // AND the rcvBuffer length is currently *precisely* an exact
+            // multiple of the length of
+            //   'connect:$socketAESKey:$socketIV'
+            //       7   1     44      1   24
+            // i.e. an exact multiple of 77
+            // THEN we should also go ahead and process the request (or requests)
```
gkc, line 904: Note: The algorithm above can fail in the unlikely event that
- 1) a _partial_ fragment of a given control request is received by the socket listener, and
- 2) after we add the very last byte of the fragment to the receive buffer, the buffer length is an exact multiple of 77 bytes

Open to suggestions for improvement

gkc:
One thing left to do here which is to make this a daemon feature and only send heartbeats if the daemon supports them

xavierchanth:
> One thing left to do here which is to make this a daemon feature and only send heartbeats if the daemon supports them

I would rather we mark this as control protocol version (v1) in the daemon features.

gkc:
> > One thing left to do here which is to make this a daemon feature and only send heartbeats if the daemon supports them
> 
> I would rather we mark this as control protocol version (v1) in the daemon features.

Do you mean you would like to rename the controlChannelHeartbeats daemon feature to  controlProtocolV1 or controlProtocol_v1 or something like that?

cconstab (dismissed):
tested and works as advertised, watched the protocol flow using WireShark, all looked good.

gkc:
@XavierChanth please take a look at the changes in the latest commit

xavierchanth:
> @XavierChanth please take a look at the changes in the latest commit

My concern with this approach is that the order and cadence of features added to other daemon implementations (i.e. C) may not match the Dart implementation. Thus potentially needed features may be gate-kept behind unneeded ones.

gkc:
> > @XavierChanth please take a look at the changes in the latest commit
> 
> 
> 
> My concern with this approach is that the order and cadence of features added to other daemon implementations (i.e. C) may not match the Dart implementation. Thus potentially needed features may be gate-kept behind unneeded ones.

I don't see why. Implementations should add the features they want, and advertise them. It's up to the clients to decide what to do given what the daemon advertises

xavierchanth:
> > > @XavierChanth please take a look at the changes in the latest commit
> > 
> > 
> > My concern with this approach is that the order and cadence of features added to other daemon implementations (i.e. C) may not match the Dart implementation. Thus potentially needed features may be gate-kept behind unneeded ones.
> 
> I don't see why. Implementations should add the features they want, and advertise them. It's up to the clients to decide what to do given what the daemon advertises



> > > @XavierChanth please take a look at the changes in the latest commit
> > 
> > 
> > My concern with this approach is that the order and cadence of features added to other daemon implementations (i.e. C) may not match the Dart implementation. Thus potentially needed features may be gate-kept behind unneeded ones.
> 
> I don't see why. Implementations should add the features they want, and advertise them. It's up to the clients to decide what to do given what the daemon advertises

I just had a re-read, I mistook the since value as absolute, assuming the features list had been removed.

xavierchanth (approved):

PR #1960, opened 2025-06-02
fix: convert device name to lowercase

xavierchanth:
**- What I did**

- Fixed cjson build issue : https://github.com/atsign-foundation/at_c/pull/592
- removed unused test variable
- Added a loop which converts uppercase chars to lowercase in devicename during the parsing phase

**- How I did it**

**- How to verify it**
Added temporary printf:
```c
  // Repeat for permit-open
  // Convert devicename to lower case
  for (char *c = params->device; c[0] != '\0'; c++) {
    if (*c >= 'A' && *c <= 'Z') {
      *c += ('a' - 'A');
    }
  }

  printf("dev: '%s'\n", params->device);
  return 0;
```
command input / printf output:
```
 ./build/dev-linux-x86_64/sshnpd/sshnpd -a @xchan -m @snooker
25 -d ASDSD_123Z
permitting open:
localhost:22
localhost:3389
dev: 'asdsd_123z'
```
fix: convert device name to lowercase

### DIFF
--- a/packages/c/cmake/atsdk.cmake
+++ b/packages/c/cmake/atsdk.cmake
@@ -12,7 +12,7 @@ if(NOT atsdk_FOUND)
     FetchContent_Declare(
       atsdk
       GIT_REPOSITORY https://github.com/atsign-foundation/at_c.git
-      GIT_TAG e62f9b41b7b7cd5b554c129d3ab38ad4f8d8aa80
+      GIT_TAG 129baf255544c69a9ebd9544a22e74d080a19598
     )
   endif()
   FetchContent_MakeAvailable(atsdk)
--- a/packages/c/cmake/atsdk.cmake
+++ b/packages/c/cmake/atsdk.cmake
@@ -12,7 +12,7 @@ if(NOT atsdk_FOUND)
     FetchContent_Declare(
       atsdk
       GIT_REPOSITORY https://github.com/atsign-foundation/at_c.git
-      GIT_TAG 129baf255544c69a9ebd9544a22e74d080a19598
+      GIT_TAG b134eaa3e1092a41f107d3abcb12d8bd6eca7b17
     )
   endif()
   FetchContent_MakeAvailable(atsdk)
--- a/packages/c/cmake/atsdk.cmake
+++ b/packages/c/cmake/atsdk.cmake
@@ -12,7 +12,7 @@ if(NOT atsdk_FOUND)
     FetchContent_Declare(
       atsdk
       GIT_REPOSITORY https://github.com/atsign-foundation/at_c.git
-      GIT_TAG b134eaa3e1092a41f107d3abcb12d8bd6eca7b17
+      GIT_TAG d6898eb2885acb7968be631b39d91994c6a15953
     )
   endif()
   FetchContent_MakeAvailable(atsdk)
--- a/packages/c/sshnpd/src/params.c
+++ b/packages/c/sshnpd/src/params.c
@@ -179,5 +179,11 @@ int parse_sshnpd_params(sshnpd_params *params, int argc, const char **argv) {
   }
 
   // Repeat for permit-open
+  // Convert devicename to lower case
+  for (char *c = params->device; c[0] != '\0'; c++) {
+    if (*c >= 'A' && *c <= 'Z') {
+      *c += ('a' - 'A');
+    }
+  }
   return 0;
 }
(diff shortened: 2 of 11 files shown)


packages/c/sshnpd/tests/test_permit_open.c
```diff
@@ -44,8 +44,6 @@ int star_star_test() {
   params.permitopen_ports = ports;
   params.permitopen_len = 1;
 
-  bool allow;
```
xavierchanth, line 47: These became unused due to a test refactor

packages/c/cmake/atsdk.cmake
```diff
@@ -12,7 +12,7 @@ if(NOT atsdk_FOUND)
     FetchContent_Declare(
       atsdk
       GIT_REPOSITORY https://github.com/atsign-foundation/at_c.git
-      GIT_TAG e62f9b41b7b7cd5b554c129d3ab38ad4f8d8aa80
+      GIT_TAG 129baf255544c69a9ebd9544a22e74d080a19598
```
xavierchanth: https://github.com/atsign-foundation/at_c/pull/592/files

packages/c/sshnpd/src/params.c
```diff
@@ -179,5 +179,11 @@ int parse_sshnpd_params(sshnpd_params *params, int argc, const char **argv) {
   }
 
   // Repeat for permit-open
+  // Convert devicename to lower case
+  for (char *c = params->device; c[0] != '\0'; c++) {
```
xavierchanth, line 183: This reads: start the pointer at the beginning of the device name, move it forward once each iteration until we reach the null terminator.

packages/c/sshnpd/src/params.c
```diff
@@ -179,5 +179,11 @@ int parse_sshnpd_params(sshnpd_params *params, int argc, const char **argv) {
...
+    if (*c >= 'A' && *c <= 'Z') {
+      *c += ('a' - 'A');
+    }
```
xavierchanth, line 186: Then check if the value of the character at the current pointer position is an uppercase ascii char, if it is, we convert it in place to lowercase.

packages/c/sshnpd/src/params.c
xavierchanth: This string is expected to always be null-terminated, unless a bug occurs in argparse, or we change how our defaults are set (which they should still be null-terminated even if we do change them).

xavierchanth:
We need the `patch` command in the docker container now, I assume that's why e2e tests are failing to build. I'm on a time constraint, will pick that up in the morning.

cpswan:
> We need the `patch` command in the docker container now, I assume that's why e2e tests are failing to build. I'm on a time constraint, will pick that up in the morning.

There's no `patch` in either the e2e base image or the CMake image that we use. Looking at where we do the build it seems that we need to get it into the CMake image.

```log
$ sudo docker run -it atsigncompany/cbuildimage:CMake-3.31.4@sha256:0b443da3bccd746e23323eb374d19ba156b497ecbee6c5e84f97d77954515cbf /bin/bash
Unable to find image 'atsigncompany/cbuildimage:CMake-3.31.4@sha256:0b443da3bccd746e23323eb374d19ba156b497ecbee6c5e84f97d77954515cbf' locally
docker.io/atsigncompany/cbuildimage@sha256:0b443da3bccd746e23323eb374d19ba156b497ecbee6c5e84f97d77954515cbf: Pulling from atsigncompany/cbuildimage
fa953b1ccc1b: Pull complete
3768bb6c77bd: Pull complete
0b520bb3838f: Pull complete
Digest: sha256:0b443da3bccd746e23323eb374d19ba156b497ecbee6c5e84f97d77954515cbf
Status: Downloaded newer image for atsigncompany/cbuildimage@sha256:0b443da3bccd746e23323eb374d19ba156b497ecbee6c5e84f97d77954515cbf
root@ef9988e5e6a8:/# which patch
root@ef9988e5e6a8:/# patch
bash: patch: command not found
```

cpswan (dismissed):

cpswan:
@XavierChanth e2e tests now passing with `patch` in the 3.31.7 release of the CMake build image :)

I'm happy for this to be merged, but:
* we need to do a c1.0.12 release (with all the usual bumps)
* there should be a new test for the coercion to lowercase

So those things can go into this PR or another...

xavierchanth:
> * we need to do a c1.0.12 release (with all the usual bumps)
> 
> * there should be a new test for the coercion to lowercase

Both of these are now included in this PR

cpswan (approved):

PR #1964, opened 2025-06-03
ci: add c:1.0.12 to e2e_all

### LINKED ISSUE #1790: e2e tests: Add C daemon tags
- Now that C Daemon is 1.0.0 and we are about to release a new version, we should regression test against c 1.0.0 & c branch

JeremyTubongbanua:
closes #1790 

**- What I did**
- Added `c:1.0.12` to list of `daemonVersions` in e2e_all `main.sh`
- Modified `*.release` Dockerfiles so that they accept a version without a prepending `v` or `c`: e.g. `--build-arg release=1.0.12` (I implied it that `v` version is for Dart and `c` version is for C)
- Implemented a `downloadCBInaries` function for future proofing
- Made it so that `setup_binaries.sh` will only download client versions, as we only use locally downloaded client binaries to execute `./sshnp` or `./npt` binaries, so daemon versions don't matter.

**- How to verify it**
See comment https://github.com/atsign-foundation/noports/pull/1964#issuecomment-2949427100

test: add c:1.0.12 to e2e_all

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,6 +14,7 @@ updates:
     directories:
       - "/packages/c/sshnpd/tools/"
       - "/packages/dart/sshnoports/tools/"
+      - "/tests/e2e_all/dockerfiles/"
       - "/tests/end2end_tests/image/"
       - "/tools/multibuild/"
     schedule:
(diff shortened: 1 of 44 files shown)


JeremyTubongbanua:
`daemons/` folder
image

By adding c:1.0.12 to the list of daemons, the number of tests was increased to 62

image

Samples of c1.0.12 tests running. As you can see from image below, some pass and some are not applicable.
image

cpswan (approved):

PR #1965, opened 2025-06-03
cd: add more platforms to e2e all dockerhub workflow

JeremyTubongbanua:
**- What I did**
- Push more platforms to dockerhub

**- How I did it**

**- How to verify it**

cd: add more platforms to e2e all dockerhub workflow


cpswan (approved):

cpswan:
When it comes to using these new images I'd suggest using an OS matrix that has regular and arm runners.

e.g. in [.github/workflows/dockerhub_sshnpd.yml](https://github.com/atsign-foundation/noports/blob/trunk/.github/workflows/dockerhub_sshnpd.yml)

```yaml
jobs:
  docker:
    strategy:
      matrix:
        os: [ubuntu-24.04-arm, ubuntu-24.04]
```

PR #1984, opened 2025-06-11
fix: Use correct tag for releases and don't fail fast

cpswan:
The [Docker image builds for pre-release v5.10.0](https://github.com/atsign-foundation/noports/actions/runs/15569502840) failed due to the tag section looking for `l` (which was used in testing) rather than `v`

**- What I did**

* Swapped `l` for `v`
* Also set matrix strategy to not fail fast

**- How to verify it**

@gkc I'd like to re-release 5.10.0 by deleting the existing tag and recreating the release. Alternatively we can do a 5.10.1. LMK?

https://github.com/atsign-foundation/noports/actions/runs/15569502840

### DIFF
--- a/.github/workflows/dockerhub_sshnpd.yml
+++ b/.github/workflows/dockerhub_sshnpd.yml
@@ -14,6 +14,7 @@ permissions: # added using https://github.com/step-security/secure-workflows
 jobs:
   docker:
     strategy:
+      fail-fast: false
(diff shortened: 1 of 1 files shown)


gkc:
@cpswan let's delete v5.10.0 and recreate it

gkc (approved):

PR #2029, opened 2025-07-03
feat: improve time-to-session

gkc:
**- What I did**

**- How I did it**

**- How to verify it**


gkc:
I did actually write a proper description but must not have saved to GitHub 

Basically
1. Use latest at_onboarding_cli which avoids an unnecessary creation and authentication of another at lookup connection 
2. Inform srvd if there are keys which should be pre-fetched to save on latency during socket auth verification 

See commit messages and changeset for more detail

cconstab (approved):

PR #2047, opened 2025-07-18
build: npt_flutter: Add direct dependency on at_client ^3.6.0

gkc:
**- What I did**
Add direct dependency on at_client (npt_flutter does directly use at_client features, and we want to set a minimum version)

**- How I did it**
Updated pubspec.yaml, ran pub upgrade to update pubspec.lock

**- How to verify it**
Review the commit


gkc:
Rendered obsolete by https://github.com/atsign-foundation/noports/pull/2046

PR #2058, opened 2025-07-29
chore: cleanup universal.sh

xavierchanth:
**- What I did**

Simplify universal.sh:
- Automatically re-runs itself as root with sudo
- No longer depends on install.sh (though it is still around since we use it in the manual installation guide)
- No longer requires additional inputs
- No longer configures / starts services for you
- No more nohup/tmux install

Whilst I agree that this makes it "harder" to install, it does make installs that go wrong easier, because we can easily determine what went wrong. Whereas if we try to do things automatically, making assumptions about the os, it becomes much more challenging.

However, the only additional skills required by this are:
- edit a file
- start/manage a service

If you can't already do either of those things, then how do you maintain your machine?

**- How I did it**

**- How to verify it**

Will test tomorrow.

chore: cleanup universal.sh


xavierchanth:
@cpswan is now the right time to replace macos with a homebrew package of noports that has similar behavior to this?

Brew does have a `brew services` subcommand. Overall maintaining macos separately would be easier long term.

cpswan:
> @cpswan is now the right time to replace macos with a homebrew package of noports that has similar behavior to this?

I think yes; and it might be the right place to continue our journey into the world of package managers.

xavierchanth:
Brew tap created at https://github.com/atsign-foundation/homebrew-tap

Next steps:

- Drop macos here and recommend brew when universal.sh is run on a mac.
- Update documentation for macos and linux

xavierchanth:
Closing this PR, instead I'm going to create a separate script to go with a new flow. universal.sh will remain as is, after we migrate documentation to the new script we can consider deprecating universal.sh.

PR #2066, opened 2025-08-07
docs(automated): Update docs from Gitbook

atsignbot:
Sync Gitbook docs branch to trunk


xavierchanth:
[26b0e1d](https://github.com/atsign-foundation/noports/pull/2066/commits/26b0e1dd2f1251e9e21893eeabf766dccf17f331) fixes gitbook using `core.ignorecase = false` in their git config server side (terrible default since a majority of users are macos/windows).

APFS at least has a case sensitive version, but it is not the default fs.

xavierchanth (approved):

PR #2028, opened 2025-07-02
feat: npt_flutter basic policy + ui top bar + 443 checkbox in profile edit form

JeremyTubongbanua:
closes #2018 
closes #2089 

**- What I did**

3 Major New Features

- Basic Policy (what was in the web app is now in our Flutter desktop app)
- Sleak top UI bar 
- 443 check box in the proifle edit screen which enables `--443` and `--relay-auth-mode escr`

**- How I did it**
- using Claude

**- How to verify it**
- See snippets which verify 443 checkbox is working- https://github.com/atsign-foundation/noports/pull/2028#issuecomment-3203892955
- See images of UI changes and new features- https://github.com/atsign-foundation/noports/pull/2028#issuecomment-3235295144

**- Description for the changelog**
feat: npt_flutter basic policy + ui top bar + 443 checkbox in profile edit form

PR #2124, opened 2025-08-29
feat: NoPorts Desktop keep alive checkbox

### LINKED ISSUE #1977: NoPorts Desktop persistent connections

When using npt you can specify a persistent connection (-K), I would like this to be available in the Desktop App

This is useful for API connections that only get used on occasion and so often time out after a while on non use.

migrating to npt, works but not friendly for the MCP UI user.

### Additional context

_No response_

JeremyTubongbanua:
closes #1977 

**- What I did**
- Remove white space in some of the policy source code files
- **NEW** Keep alive checkbox widget
- **NEW** Profile business logic for retrying a connection
- Updated profile tests to take into account new `keepAlive` variable

**- How to verify it**

[Video demo](https://www.youtube.com/watch?v=Ka-_PYn3NV8)

- left terminal = doing `curl`
- middle window = NoPorts Desktop 
- right terminal = `flutter run -d macos`

In the first half video demo, I uncheck the "Keep Alive" option. I show that the tunnel will drop when I do `curl localhost:1111`. The reason it drops is because the host exists, but there is no service running on the port, so (I assume) my client is attempting to write to a TCP socket that won't work so therefore closes it(?). In the second half of the video, I check the "Keep Alive" option and show that the connection will recover and show logs to prove it.

feat: NoPorts Desktop keep alive checkbox


JeremyTubongbanua:
I would like to merge https://github.com/atsign-foundation/noports/pull/2028 before I open this for review

JeremyTubongbanua:
See https://www.youtube.com/watch?v=Ka-_PYn3NV8

Also read PR description (I just updated it)

xavierchanth:
Nothing to do with your PR, but I very much dislike how this: 
image

doesn't trigger an actual rebase, making the commit history inaccurate to the diffs

packages/dart/npt_flutter/lib/features/profile/bloc/profile_bloc.dart
```diff
@@ -256,12 +256,108 @@ class ProfileBloc extends LoggingBloc<ProfileEvent, ProfileState> {
...
+
+  Future<void> _retryConnection(
+    Emitter<ProfileState> emit,
+    Profile profile,
+    AtClient atClient,
+    String atSign,
+    Settings settings,
+  ) async {
+    // Wait 5 seconds before retrying, similar to npt binary
+    await Future.delayed(const Duration(seconds: 5));
```
xavierchanth, line 285: Why do we wait 5 seconds in npt binary? Is this something we even need to do. A short delay is fine, but 5 seconds seems excessive.

xavierchanth: cc: @gkc

packages/dart/npt_flutter/pubspec.lock
```diff
@@ -324,7 +324,7 @@ packages:
     source: hosted
     version: "8.12.0"
   chalkdart:
-    dependency: transitive
+    dependency: "direct overridden"
```
xavierchanth, line 327: I think you need to do another pub get and commit the changes, there are no changes to the pubspec.yaml but I'm seeing that a dependency override was added at some point. It may also be that a previous pr didn't check in the pubspec.lock.

xavierchanth (approved):

packages/dart/npt_flutter/pubspec.lock
JeremyTubongbanua: I did a flutter pub get and dart pub get in `packages/dart/npt_flutter` and the pubspec.lock didn't change

PR #2145, opened 2025-09-10
feat: created mutex for sshnpd for load balancing/redundancy etc

cconstab:
**- What I did**
Added functionality to sshnpd to be able to run multiple sshnpd for redundancy or load balancing

**- How I did it**
Used the Mutex code from srvd and added it to the sshnpd code base so only one sshnpd replies

**- How to verify it**
Tested by hand and passed tests

sshnpd load balacing / redundancy support similar to the srvd existing solution using mutex code

### DIFF
--- a/docs/sshnpd-load-balancing.md
+++ b/docs/sshnpd-load-balancing.md
@@ -1,0 +1,73 @@
+# Session-Based Mutex for sshnpd Load Balancing
+
+## Overview
+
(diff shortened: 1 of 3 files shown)


gkc:
lgtm @cconstab, nice and clean! Would like @XavierChanth to take a look also

gkc (approved):

cconstab:
Sorry hit commit.. @XavierChanth take a look and we can pull if need be

xavierchanth:
LGTM

PR #2163, opened 2025-09-27
feat: Added --local-host <IP/FQDN> to NPT

cconstab:
**- What I did**
Added the optional --lh <IP> and moved the default value from 0.0.0.0 to localhost
***** Possible Breaking change ******

**- How I did it**
Worked with Claude and checked everything, as code was added.
**- How to verify it**
Tested by hand added tests and tests pass
Added the optional --lh <IP> and moved the default value from 0.0.0.0 to localhost


cconstab:
Two qustions...

 Should we push this chnage as I think binding to 0.0.0.0 is VERY bad... Or should we wait until we have tests to make sure we never bind to 0.0.0.0 by accident again?
 
 Should we default to `localhost` binding or to '127.0.0.1' ? I prefered localhost as folks move to IPv6 this still works in the distant future..

Original issue asked for a comma delimetered IP list to bind to. Given then above and that to introducae that would mean updating the connector lib, I decided that a single IP per instance, including 0.0.0.0 to bind to all interfaces. For the moment at least a single IP bind. If multiple  IPs need to be binded then use multiple npt instances.

xavierchanth:
I think this one needs an arch discussion.

cconstab:
Sure what would be the topic?

xavierchanth:
> Sure what would be the topic?

1. Whether to change the default host (breaking change, but probably yes)

2. How to handle ipv4/6 and which should be the default resolution strategy, also, should we bind to both if provided a hostname.

gkc:
My two cents:
- We should bind to `localhost` by default, allowing the host to determine whether ipv4 or ipv6 is prioritized when connecting to `localhost`
- Allow provision of a single IP address to which to bind; run multiple npts if you need to bind to more than one

cconstab:
Updated to bind to `localhost` and let the local OS decide what that is. Interestingly MacOS decides that ::1 is localhost no matter what you put in the /etc/hosts. That said it all works as it shoud. Linux works just as you would expect and defaults localhost to IPv6. 
In all cases npt now outputs where it has bound in hostname and IP address terms..

```
╭─cconstab@tarial in ~/eem 
╰$ /tmp/npt -f @cconstab -t @ssh_1 -r @rv_am -d iot_device01 -p 22 -l 2222 -T 0 -K --local-host tarial 
Connecting ... Connected
2025-09-28 13:02:29.080823 : Requested "never" timeout: set to 365 days
2025-09-28 13:02:29.088165 : Will bind to local host: tarial
2025-09-28 13:02:29.091244 : Sending daemon feature check request
2025-09-28 13:02:29.091337 : Fetching host and port from srvd
2025-09-28 13:02:30.767117 : Received host and port from srvd
2025-09-28 13:02:30.767157 : Waiting for daemon feature check response
2025-09-28 13:02:30.814870 : Received daemon feature check response
2025-09-28 13:02:30.827698 : Will send control channel heartbeats
2025-09-28 13:02:30.941982 : Required daemon features are supported
2025-09-28 13:02:30.942411 : Sending session request to the device daemon
2025-09-28 13:02:31.028594 : Waiting for response from the device daemon
2025-09-28 13:02:31.442763 : Received response from the device daemon
2025-09-28 13:02:31.446182 : Will use local port 2222
2025-09-28 13:02:31.446190 : Creating connection to socket rendezvous
2025-09-28 13:02:31.484950 : npt is listening on tarial:2222 (127.0.0.1:2222)
```
Or
```
╭─cconstab@tarial in ~/eem 
╰$ /tmp/npt -f @cconstab -t @ssh_1 -r @rv_am -d iot_device01 -p 22 -l 2222 -T 0 -K --local-host 0.0.0.0
Connecting ... Connected
2025-09-28 13:03:25.210814 : Requested "never" timeout: set to 365 days
2025-09-28 13:03:25.210832 : Will bind to local host: 0.0.0.0
2025-09-28 13:03:25.212536 : Sending daemon feature check request
2025-09-28 13:03:25.212642 : Fetching host and port from srvd
2025-09-28 13:03:26.818306 : Received host and port from srvd
2025-09-28 13:03:26.818358 : Waiting for daemon feature check response
2025-09-28 13:03:26.858028 : Received daemon feature check response
2025-09-28 13:03:26.871316 : Will send control channel heartbeats
2025-09-28 13:03:26.938267 : Required daemon features are supported
2025-09-28 13:03:26.938666 : Sending session request to the device daemon
2025-09-28 13:03:27.049897 : Waiting for response from the device daemon
2025-09-28 13:03:27.451928 : Received response from the device daemon
2025-09-28 13:03:27.455697 : Will use local port 2222
2025-09-28 13:03:27.455708 : Creating connection to socket rendezvous
2025-09-28 13:03:27.491249 : npt is listening on 0.0.0.0:2222 (0.0.0.0:2222)
```

I did think about have an `all` option but I think the 0.0.0.0 is standard network bind and is now noted in help text

packages/dart/sshnoports/bin/npt.dart
```diff
@@ -112,6 +112,17 @@ void main(List<String> args) async {
...
+        'local-host',
+        aliases: ['lh'],
+        help:
+            'Local IP address to bind to, or comma-separated list with fallbacks.'
+            ' When specified, npt will act as a gateway by binding to the'
+            ' first available IP instead of the default localhost (127.0.0.1).'
+            ' If multiple IPs are provided, only the first valid one is used.'
+            ' Example: --local-host 192.168.1.100,10.0.0.50'
+            ' Alias: --lh',
+      );
```
JeremyTubongbanua, line 128: Is there a reason we're not using the "defaultsTo" argument?

```
defaultsTo: 'localhost',
```

https://pub.dev/documentation/args/latest/args/ArgParser/addOption.html

JeremyTubongbanua (commented):
+1 for npt should bind to localhost by default

packages/dart/sshnoports/bin/npt.dart
cconstab: let me fix that !
TY

packages/dart/sshnoports/bin/npt.dart
```diff
@@ -364,6 +378,52 @@ void main(List<String> args) async {
...
+          .split(',')
+          .map<String>((ip) => ip.trim())
+          .where((String ip) => ip.isNotEmpty)
+          .toList();
+
+        // Find the first valid IP address from the list
+        for (String ip in parsedHosts) {
+          bool isValid = false;
+          try {
+            // Try to parse as IP address - this will throw if invalid
```
gkc: Suggest use [tryParse](https://api.flutter.dev/flutter/dart-io/InternetAddress/tryParse.html) instead

packages/dart/noports_core/test/npt/local_host_edge_cases_test.dart
```diff
@@ -0,0 +1,297 @@
+import 'dart:io';
+import 'package:test/test.dart';
+
+void main() {
```
gkc: There are a lot of test cases here but they are arguably of zero actual value since they are largely testing behaviour of dart's `InternetAddress`.

I'd suggest all of these tests be deleted and instead
1. refactor the new `Parse and validate local host` code in sshnoports/bin/npt.dart into a function in noports_core
2. write a small number of unit tests in noports_core to veify behaviour of that function

packages/dart/noports_core/test/npt/local_host_test.dart
```diff
@@ -0,0 +1,169 @@
+import 'dart:io';
+import 'package:test/test.dart';
+import 'package:args/args.dart';
+
+void main() {
+  group('Local Host CLI Argument Tests', () {
+    late ArgParser parser;
+
+    setUp(() {
+      // Create a basic parser similar to what npt uses
```
gkc: omg no.

gkc (changes requested):
LGTM except for the ai-generated unit test slop

packages/dart/noports_core/test/npt/local_host_edge_cases_test.dart
cconstab: I think Claude got over excited

cconstab:
I was thinking to drop most/all tests too..

cconstab:
Bogus tests removed and migrated to tryParse.. Thanks @gkc

packages/dart/sshnoports/bin/npt.dart
cconstab: moved to tryParse

cconstab:
As per suggestion moved the host validation to core and have sensible tests

cconstab:
After arch call `localhost` as default agreed and pre bind is fine..

Also add the --4 --6 flags like ssh to only use IPv4/IPv6

cconstab:
@XavierChanth @gkc I think everything is in order now..

gkc:
Thinking about this some more, I wonder if, when using a local-host arg of `localhost`, we should default to using ipv4 rather than letting the host decide. I'm a bit torn on this; I feel like most people would expect ipv4 as default, but some people might be annoyed by it not taking the host's preference, and others might expect it to bind to both ipv4 and ipv6. @cconstab I can look into socket_connector to see how feasible it is to add support for more than one ServerSocket

cconstab:
I am also torn but if people use localhost as a name it will work as itvstands they just need to be consistent. 
This is a Linux distro issue  so I do not feel we have to fix it.
We should allow the OS to pick.

gkc:
> I am also torn but if people use localhost as a name it will work as itvstands they just need to be consistent. This is a Linux distro issue so I do not feel we have to fix it. We should allow the OS to pick.

OK I'm good with this. FYI re socket_connector - it's straightforward enough to add a new static function `SocketConnector.serversToSockets`, however making the changes to ripple all of that through into NoPorts is a pain and I think it better to let our decision stand (run multiple npt if you want multiple serverSockets bound)

gkc (dismissed):
LGTM but I cannot merge as I pushed a commit to update versions

xavierchanth:
> Thinking about this some more, I wonder if, when using a local-host arg of `localhost`, we should default to using ipv4 rather than letting the host decide. I'm a bit torn on this; I feel like most people would expect ipv4 as default, but some people might be annoyed by it not taking the host's preference, and others might expect it to bind to both ipv4 and ipv6. @cconstab I can look into socket_connector to see how feasible it is to add support for more than one ServerSocket

We can definitely bind to multiple sockets very easily, all we need to do is update the arg parser and merge a series of streams, since a server socket is just a Stream<Socket> in dart.

xavierchanth:
Will review this in a bit

xavierchanth:
There's something a bit off based on the way we handle lookups and bind checks, you can get a valid lookup which is not valid to bind to. Thus, we shouldn't consolidate down to a single host to bind to too early, but also, we shouldn't do all of our lookups first. The pattern should be:

```pseudo
for host in hosts:
  addrinfo = lookup(host)
  if addrinfo is valid:
    sock = bind(host)
    if sock is not null:
      close(sock)
      return host 
```

For reference, I'm seeing multiple resolution strategies in this changeset, but we should normalize on the one in the pseudo code above in my opinion. This minimizes the number of syscalls while still providing the most accurate resolution of valid, bindable hosts from the list.

This is all due to the fact that you can lookup another host, resulting in a valid lookup but invalid bind. We only want to lookup and bind to hosts for the current machine. This is a very possible scenario, especially if there's a dhcp server which reissues all of its leases.

cconstab:
I think we check hosts in order for lookup and binding. The first to succeed binding is used..

packages/dart/noports_core/lib/src/npt/npt.dart
```diff
@@ -305,16 +305,20 @@ class _NptImpl extends NptBase
...
+
     if (params.localPort == 0) {
       sendProgress('Finding an available local port');
 
-      /// Find a port to use
-      final server = await ServerSocket.bind(InternetAddress.anyIPv4, 0);
+      /// Find a port to use - params.localHost is now a resolved IP address
+      final bindAddress =
+          InternetAddress.tryParse(params.localHost!) ??
+          InternetAddress.loopbackIPv4; // Fallback if somehow not a valid IP
```
xavierchanth, line 315: This only takes a single host.

packages/dart/noports_core/lib/src/common/host_validator.dart
```diff
@@ -0,0 +1,130 @@
+import 'dart:io';
+
+/// Utility class for validating and parsing host addresses
+class HostValidator {
```
xavierchanth, line 4: This is implemented, but most of it isn't being used in the rest of the PR, and there's still plenty of other places interacting with the Dart:io APIs directly when it should be using this class.

packages/dart/noports_core/lib/src/srv/srv_impl.dart
```diff
@@ -1363,19 +1370,17 @@ class SrvImplDart implements Srv<SocketConnector> {
...
-        hostToLookup,
-        type: InternetAddressType.IPv6,
-      );
-    }
     if (candidates.isEmpty) {
       throw Exception("Cannot resolve address for $hostToLookup");
     }
+    logger.info('Resolved local host $hostToLookup to ${candidates[0]}');
     return candidates[0];
   }
```
xavierchanth, line 1385: Here could also be using your new host_validator.dart instead of hard implementation.

xavierchanth:
Am I missing something? I see an implementation (host_validator.dart) which (mostly) does what we've discussed, but that API isn't being used in the npt binary where I would expect it to be called.

packages/dart/noports_core/lib/src/common/host_validator.dart
cconstab: Ok gotcha let me take a look...I justvdid not want to hold up other stuff

cconstab: Ok gotcha let me take a look...I justvdid not want to hold up other stuff

cconstab:
Thatbis called in the npt.dart in the sshnoports dir which calls the npt.dart in core.
The idea was to keep the core simple and the outer npt.dart decide what to send to tge core lib.

xavierchanth (dismissed):

cconstab:
Just a quick note that after chatting with @XavierChanth we decided to move to the standard -4/-6 as per ssh with --ipv4/--ipv6 as alternatives

gkc (approved):

xavierchanth (approved):

PR #2202, opened 2025-10-10
fix: npt_flutter read the correct heartbeat key namespace and unify the heartbeat rpc notifications to use `sshnp` namespace

JeremyTubongbanua:
**- What I did**

## **first thing is the heartbeat key fix**

You get this error message when trying to read the Policy Server's heartbeat on NoPorts Desktop (`heartbeat.noports@` <- incorrect)

image

But that key will never exist because the actual key's name is this: `heartbeat.sshnp@` <- correct

image

## **second thing is changing the atrpc base namespace from `sshnp.noports` to just `sshnp`**

After some thought, I think that both the NoPorts Desktop client and Policy server should just be using the `sshnp` namespace because it is guaranteed that the client's APKAM keys using either of these services will be allowed on the `sshnp` namespace, but might not always be the `noports` namespace.

For some context, the `noports` namespace is used in NoPorts Desktop and `sshnp` is used in `sshnoports/bin/` applications. These two types of applications will surely be used by APKAM keys that have access to the `sshnp` space, but might not always have access to the `noports` namespace.

For some more additional context, we had a `namespaceAware=true` bug that was fixed in `at_client: ^3.8.0` which is why it was `sshnp.noports` in the first place, but now that it is fixed, we should just change it to `sshnp` to make things simple.

A good take away from this is that when you are writing same atSign to same atSign notifications code within NoPorts , developer should use the `sshnp` namespace so that everything works good

**- Description for changelog**
fix: npt_flutter read the correct heartbeat key namespace and unify the heartbeat rpc notifications to use `sshnp` namespace

### DIFF
--- a/packages/dart/npt_flutter/lib/features/policy/cubit/status_light/policy_status_light_cubit.dart
+++ b/packages/dart/npt_flutter/lib/features/policy/cubit/status_light/policy_status_light_cubit.dart
@@ -19,7 +19,7 @@ final class PolicyStatusLightCubit
       final AtKey atKey = AtKey()
         ..key = 'heartbeat'
         ..sharedBy = atClient.getCurrentAtSign()
-        ..namespace = atClient.getPreferences()!.namespace!;
+        ..namespace = 'sshnp';
 
       final AtValue atValue = await atClient.get(
         atKey,
--- a/packages/dart/npt_flutter/lib/features/policy/cubit/status_light/policy_status_light_cubit.dart
+++ b/packages/dart/npt_flutter/lib/features/policy/cubit/status_light/policy_status_light_cubit.dart
@@ -124,7 +124,7 @@ final class PolicyStatusLightCubit
     final rpc = AtRpcClient(
       serverAtsign: atClient.getCurrentAtSign()!,
       atClient: atClient,
-      baseNameSpace: 'sshnp.noports',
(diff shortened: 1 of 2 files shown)


JeremyTubongbanua:
After changing the atRPC notifications, it still works:

NoPorts desktop light:

image

npp_atserver logs:

```
INFO|2025-10-10 15:54:01.748691|AtRpc|Subscribing to request.\d+.auth_checks.__rpcs.sshnp@ 
...

SHOUT|2025-10-10 15:55:53.778883| npp |Sending AtRpcResp... 
INFO|2025-10-10 15:55:53.779038|AtRpc|Sending notification @tastelessbanana:success.1760126153560516.npp_atserver_heartbeat.__rpcs.sshnp@tastelessbanana with payload {reqId: 1760126153560516, respType: success, payload: {success: true}, message: Successfully forced heartbeat.} 
INFO|2025-10-10 15:56:01.814893| npp |Put timestamp key `heartbeat.sshnp@tastelessbanana`: 2025-10-10 15:56:01.750098, success: true 
SHOUT|2025-10-10 15:56:19.055514|PolicyServiceWithAtClient|Received device heartbeat from @soccer99 
INFO|2025-10-10 15:56:19.055818| npp |daemonAtSigns is now {@soccer99, @device} 
```

xavierchanth (approved):

PR #2217, opened 2025-10-17
ci: Add Docker image signing and more SLSA

cpswan:
Progresses https://github.com/atsign-foundation/operations/issues/304

**- What I did**

* Added image signing
* Added further SLSA attestations 

**- How I did it**

Based on:

* https://github.com/atsign-company/at_swarm_load/pull/75
* https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/generic/README.md#a-different-attestation-for-each-iteration

**- How to verify it**

As new images are built check for signing and attestations in Docker Hub.

ci: Add Docker image signing

### DIFF
--- a/.github/workflows/c_release.yml
+++ b/.github/workflows/c_release.yml
@@ -131,7 +131,7 @@ jobs:
             output-name: sshnpd-linux-riscv64-musl
             dockerfile: Dockerfile.musl
     ## 20240806: Clang and CMake that we use aren't packaged for
-    ## armv7 or riscv64 (also no Debian 10 for RISC-V)
+    ## armv7 or riscv64 (also no Debian 11 for RISC-V)
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
--- a/.github/workflows/dockerhub_sshnpd.yml
+++ b/.github/workflows/dockerhub_sshnpd.yml
@@ -209,7 +209,6 @@ jobs:
           echo "::notice::Generated SLSA Matrix JSON: ${MATRIX_JSON}"
 
   slsa_provenance:
-    runs-on: ubuntu-latest
     needs: [aggregate_digests]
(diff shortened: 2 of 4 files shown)


.github/workflows/c_release.yml
```diff
@@ -131,7 +131,7 @@ jobs:
             output-name: sshnpd-linux-riscv64-musl
             dockerfile: Dockerfile.musl
     ## 20240806: Clang and CMake that we use aren't packaged for
-    ## armv7 or riscv64 (also no Debian 10 for RISC-V)
+    ## armv7 or riscv64 (also no Debian 11 for RISC-V)
```
cpswan, line 134: Caught this on the way past.

.github/workflows/dockerhub_sshnpd.yml
```diff
@@ -133,8 +154,27 @@ jobs:
           docker buildx imagetools create \
             -t atsigncompany/${{ matrix.name }}:${{ env.TAG2 }} \
             --append atsigncompany/${{ matrix.name }}:amd64-${{ env.TAG2 }}
+      - name: Save digest to file and sign combined manifests
```
cpswan, line 157: As we're doing a matrix build we can't just pass digests between jobs. We need to save them to a file, aggregate them, then use the aggregate for a matrix definition.

.github/workflows/promote_canary.yml
```diff
@@ -43,6 +43,13 @@ jobs:
           docker buildx imagetools create -t \
             atsigncompany/${{ github.event.inputs.image }}:prod-wd-gha${{ github.run_number }} \
             atsigncompany/${{ github.event.inputs.image }}:canary
+          LATEST="atsigncompany/${{ github.event.inputs.image }}:latest"
```
cpswan, line 46: similar to the CANARY block in the dockerhub_sshnpd.yml workflow

gkc (approved):

PR #2265, opened 2025-10-31
fix(windows-msi): fix config and event viewer errors

Xlin123:
**- What I did**
 - permanent and neveroverride config
 - removed additional lines on event viewer
    - I wanted to switch it to WiX but it was somehow even worse than what we have setup so...
    
**- How I did it**

**- How to verify it**

fix(windows-msi): fix config and event viewer errors


Xlin123:
please squash this commit because it has excess commits that are already included in trunk... my bad

xavierchanth (approved):

PR #2287, opened 2025-11-07
fix: policy cache on edit/save

### LINKED ISSUE #2285: Policy info in-app cache is not updated when you 'save' a role update
### Describe the bug

Policy info in-app cache is not updated when you 'save' a role update

### Steps to reproduce

1. I update a rule and hit save
2. If I then select the rule again, it shows the old info
3. But if I hit refresh, and then select the rule again, it shows the correct info

### Expected behavior

Shouldn't need to refresh

### Screenshots

_No response_

_No response_

### Additional context

_No response_

JeremyTubongbanua:
closes #2285 

**- What I did**
- Fixed Policy form so you don't have to refresh to get the latest data

**- How I did it**
Using Claude

- Policy cubit has new `reloadAndSelectRole` function which is meant to be used in onSuccess.
- When editing an existing role, `reloadAndSelectRole` is called on the existing role's id
- When creating a new role, `putNewRole` is called and returns the roleId so that when a new role is created, we can call `reloadAndSelectRole` to select that newly created role.
- When we delete a role, no role is selected in the main view

**- How to verify it**
Cases that were tested

1. Editing an existing role
2. Creating a new role
3. Deleting a role

In the first two cases, they will go to the current role and have updated cache.
In third case, no role will be selected

fix: policy cache on edit/save

### DIFF
--- a/packages/dart/npt_flutter/lib/features/policy/repositories/role_repository.dart
+++ b/packages/dart/npt_flutter/lib/features/policy/repositories/role_repository.dart
@@ -59,8 +59,8 @@ class RoleRepository {
     return roles;
   }
 
-  Future<bool> putNewRole(final RoleInProgress roleInProgress) async {
-    
+  Future<String?> putNewRole(final RoleInProgress roleInProgress) async {
+
     final AtClient atClient = AtClientManager.getInstance().atClient;
     String? currentAtSign = atClient.getCurrentAtSign();
 
@@ -68,7 +68,7 @@ class RoleRepository {
       App.log(
         '[ERROR] updateExistingRole: Current atSign is null'.loggable,
       );
-      return false;
+      return null;
     }
 
     // ensure currentAtSign starts with '@'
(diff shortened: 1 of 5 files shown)


packages/dart/npt_flutter/lib/features/policy/repositories/role_repository.dart
```diff
@@ -112,10 +112,10 @@ class RoleRepository {
           );
         }
       }
-      return success;
+      return success ? newId.toString() : null;
```
JeremyTubongbanua, line 115: This is what `newId` is

```dart
final int maxId = await getMaxGroupId();
final int newId = maxId + 1;
```

JeremyTubongbanua:
```dart
Future<int> getMaxGroupId() async {
    final List<FetchedRole> roles = await fetchRoles();
    if (roles.isEmpty) return 0;

    int maxId = 0;
    for (final role in roles) {
      final parsedId = int.tryParse(role.id) ?? 0;
      if (parsedId > maxId) {
        maxId = parsedId;
      }
    }
    return maxId;
}
```

gkc (approved):
Works perfectly. LGTM

PR #2288, opened 2025-11-07
feat: config based installs

xavierchanth:
## TODOs

- [ ] Auto update tags for the new windows-msi installer
- [ ] Packages
  - [ ] deb package
  - [ ] rpm package
  - [ ] aur package
  - [ ] winget package
- [ ] authorized keys file, windows installer

**- What I did**

**- How I did it**

**- How to verify it**


.github/workflows/multibuild.yaml
```diff
@@ -392,7 +396,104 @@ jobs:
...
+        run: |
+          gh pr create \
+            --base 'trunk' \
+            --head "multibuild-${{github.run_number}}" \
+            --reviewer 'xavierchanth,cpswan' \
+            --title 'chore(automated): Updated tags from multibuild release' \
+            --body 'Sync tags from multibuild ${{ github.run_number }}' || \
+          echo "Already created?";
+
+  brew_tap:
```
xavierchanth, line 468: TODO: This job needs to be updated, it needs to:
1. checkout the homebrew-tap repo
2. Replace the noports.rb file in Casks/noports.rb with our updated copy.
3. Submit a PR to homebrew-tap

.github/workflows/multibuild.yaml
```diff
@@ -363,6 +340,40 @@ jobs:
       - name: Generate SHA256 checksums
         working-directory: tarballs
         run: sha256sum * > checksums.txt
+      - name: Update SHA256 checksums in universal.sh
```
xavierchanth, line 344: We have a chicken and egg problem now @cpswan (with the way jobs are currently laid out):

Universal.sh needs to be updated with the new checksums (as I've added checksum verification), but this then requires us to regenerate the checksums for universal.sh before provenance occurs.

We'll have the same problem with the rest of our manifests if we decide to upload them to the release (which I think we've already discussed doing).

xavierchanth: What we ultimately need is:
- Generate SHA256 checksums for the build archives
- Update the manifests with said checksums
- Generate SHA256 checksums for the manifests
- Continue with provenance

cpswan:
Overtaken by #2528

PR #2304, opened 2025-11-13
feat: Noports CLI implementation

### LINKED ISSUE #1976: Make APKAM easier with "otp and approve" pseudo command
**Context**:
In the APKAM flow, we ask the user to 1. go to client machine and generate an OTP, 2. go to device machine and enroll, and 3. go BACK to the client machine and approve.

**What I want**:
Create either a pseudo command in at_activate, or create a script, or something along those lines where the new APKAM flow would be:
1. user goes to client machine and runs the "generate OTP and auto approve" pseudo function, where the user would specify the device atSign and device name
2. user goes to device machine and enrolls 
3. tada, the apkam flow is completed and the device machine has the enrolled keys

**Some cool features maybe to add**:
- On step 1 when user generates OTP and auto approve, maybe give the user the correct command to run (aka "spoon feed them") on their device machine. But this would require an extra parameter asking for their device operating system/situation (desktop or terminal)
- there should be a timeout (30s-2min) 
- OTP is 6 alphanumeric so there is no real risk of brute force attacks
- Add a feature in the NoPorts registrar website to give them these simple commands to make installation easier

srieteja:
Closes https://github.com/atsign-foundation/noports/issues/1976

**- What I did**
- Introduced Noports CLI with `activate` and `issue-keys` commands
- Implemented a facade pattern for both commands and added individual fromArgs factories to make instantiation easier.
- `issue-keys` generates copy-pasteable activation strings, `activate` accepts them.
- Created a CLILoggingHandler to display cleaner logs for CLIs

**- How I did it**
#### Core Architecture:

- Created Activate class with CRAM and APKAM enrollment support
- Created IssueKeys class to generate enrollment strings and auto-approve requests
- Implemented NoPortsCommand enum for type-safe command routing in noports.dart

#### Activate Command:

- Added noports activate supporting formats: <atsign>:cram:<cramkey> and <atsign>:enroll:<otp>[:name:<devicename>]
- Created the ActivateParams class with parsing logic and validation
- Implemented regex-based parsing for activation strings
- Implemented help flag handling with HelpRequestedException throw-catch
- Added --keyfile option for custom atKeys file paths
- Prompts the user for a file path if a keys file already exists at the default or provided atKeys path.
- Supports verbose and debug flags

#### Issue-Keys Command:

- Generates an enrollment OTP and constructs an activation string using it. 
- Waits for the enrollment with the specific app and device name to be created; approves it once created.
- Introduced `IssueKeysParams` that has its arg parser impl and parsing
- Uses default values for namespace and appName
- Accepts user input for deviceName (fallback: noports_<OTP>), keyfile, and passphrase. Also supports verbose and debug flags.
- Support for resuming command: Checks for existing pending enrollments with given appName and deviceName. If present, it approves it and exits (so re-running the command would simply resume the flow)
- Maintains 1-hour OTP expiry with loop termination on expiration

**- How to verify it**
- Run ./buildBinaries in `sshnoports/` to generate binaries
- Run `noports issue-keys <@atsign>` to generate an enrollment string
- Copy the generated string and run `noports activate <string>` to activate
- Verify enrollment is auto-approved with correct parameters, and the keys file is generated at the user-inputted/default atKeys location

feat: introduce noports CLI with support for commands: 'activate' and 'issue-keys'.


packages/dart/sshnoports/bin/noports.dart
```diff
@@ -0,0 +1,88 @@
...
+
+enum NoportsCommand {
+  activate('activate'),
+  issueKeys('issue-keys');
+
+  final String commandName;
+
+  const NoportsCommand(this.commandName);
+
+  static NoportsCommand fromString(String value) {
```
srieteja: An enum that holds available commands and a fromString() for the same

packages/dart/sshnoports/lib/src/noports_cli/activate/np_activate.dart
```diff
@@ -0,0 +1,111 @@
+import 'dart:io';
+
+import 'package:at_auth/at_auth.dart';
+import 'package:at_onboarding_cli/at_onboarding_cli.dart';
+import 'package:sshnoports/src/noports_cli/activate/np_activate_params.dart';
+
```
srieteja: activate commands implementation classes and abstractions

packages/dart/sshnoports/lib/src/noports_cli/activate/np_activate_params.dart
```diff
@@ -0,0 +1,69 @@
+import 'package:sshnoports/src/noports_cli/util/constants.dart';
+import 'package:sshnoports/src/noports_cli/util/regex.dart';
```
srieteja: A params class that holds all currently necessary parameters and a fromArgs() method

packages/dart/sshnoports/lib/src/noports_cli/issue_keys/np_issue_keys.dart
```diff
@@ -0,0 +1,167 @@
...
+  ///
+  /// Combines atsign, otp, and optional deviceName and atKeysFilePath
+  /// into a format compatible with 'noports activate' command.
+  ///
+  /// Returns: activation string in format `<atsign>:enroll:otp:<otp>[:name:<deviceName>[:keyfile:<path>]]`
+  String _generateEnrollCommand(NoportsParams params) {
+    StringBuffer cbuf = StringBuffer();
+    cbuf.append(NPIssueKeys.baseEnrollCommand
+        .replaceFirst('<atsign>', params.atsign)
+        .replaceFirst('<otp>', params.otp!));
```
srieteja: baseEnrollCommand is `<atsign>:enroll:otp:<otp>`; replace placeholders with actual values

packages/dart/sshnoports/lib/src/noports_cli/issue_keys/np_issue_keys.dart
```diff
@@ -0,0 +1,167 @@
...
+      ..namespace = defaultEnrollmentNamespaces.toString()
+      ..enrollmentListFilter = [EnrollmentStatus.pending];
+
+    writeInfoMessage('Listening...');
+    while (true) {
+      List<Enrollment> enrollRequests = await _enrollmentService!
+          .fetchEnrollmentRequests(enrollmentListParams: enrollListParams);
+
+      if (enrollRequests.isEmpty) {
+        // keeps the loop active until enrollment fetched or terminated
```
srieteja: Currently keeps retrying forever until successful or terminated. Should we consider having a retry limit?

packages/dart/sshnoports/lib/src/create_at_client_cli.dart
```diff
@@ -5,7 +5,7 @@ import 'package:path/path.dart' as path;
 
 Future<AtClient> createAtClientCli({
   required String atsign,
-  required String atKeysFilePath,
```
srieteja, line 8: OnboardingCli knows the defaultAtKeysPath, and will assume that when this param is null.

packages/dart/sshnoports/buildBinaries
```diff
@@ -11,10 +11,11 @@ echo "Compiling sshnp"; dart compile exe --verbosity error bin/sshnp.dart -o bin
 echo "Compiling npt"; dart compile exe --verbosity error bin/npt.dart -o bin/npt &
 echo "Compiling npp_file"; dart compile exe --verbosity error bin/npp_file.dart -o bin/npp_file &
 echo "Compiling npp_atserver"; dart compile exe --verbosity error bin/npp_atserver.dart -o bin/npp_atserver &
+echo "Compiling noports_cli"; dart compile exe --verbosity error bin/noports.dart -o bin/noports &
 echo "Compiling demo/npa_always_deny"; dart compile exe --verbosity error bin/demo/npa_always_deny.dart -o bin/demo/npa_always_deny &
```
srieteja, line 15: Calling this impl `noports_cli` for now, to not confuse it with the actual noports product

packages/dart/sshnoports/pubspec.lock
```diff
@@ -50,7 +50,7 @@ packages:
     source: hosted
     version: "2.13.0"
   at_auth:
-    dependency: transitive
+    dependency: "direct main"
```
srieteja: Auto-generated, since at_auth was explicitly added as a dependency

packages/dart/sshnoports/buildBinaries
xavierchanth: Renaming files is more confusing, it complicated our build process.

packages/dart/sshnoports/lib/src/noports_cli/activate/np_activate.dart
```diff
@@ -0,0 +1,111 @@
...
+/// Handles activation of atSign accounts via CRAM authentication or APKAM enrollment
+sealed class NPActivate {
```
xavierchanth: This doesn't need to be a sealed class. And to make it a proper facade, it should have a factory which returns an instance of the implementation.

packages/dart/sshnoports/buildBinaries
srieteja: Okay, I will change that to noports

packages/dart/sshnoports/bin/noports.dart
```diff
@@ -0,0 +1,88 @@
...
+    printUsage();
+    exit(1);
+  }
+
+  final command = NoportsCommand.fromString(args[0]);
+  if (args.length == 1) {
+    writeError('You must supply an argument string');
+    printUsage(command: command);
+    exit(1);
+  }
```
gkc, line 49: Once we have determined that args[0] is a valid command, we should remove args[0] from args so that the subsequent `wrappedMain` can concern themselves solely with the arguments they are expecting

packages/dart/sshnoports/lib/src/noports_cli/issue_keys/np_issue_keys.dart
gkc: Nope; but we should consider having a resume if the user accidentally terminated before successful

gkc (commented):
Have left a couple of comments but looks pretty good so far; in order to complete my review I will have to try it out for real and review the code in an IDE. I will try to get to that tomorrow

packages/dart/sshnoports/bin/noports.dart
```diff
@@ -0,0 +1,88 @@
...
+import 'package:sshnoports/src/noports_cli/util/np_utils.dart';
+import 'package:sshnoports/src/noports_cli/util/usage_messages.dart';
+
+enum NoportsCommand {
+  activate('activate'),
+  issueKeys('issue-keys');
+
+  final String commandName;
+
+  const NoportsCommand(this.commandName);
```
gkc: I'd suggest rename NoportsCommand to NoPortsCommand

packages/dart/noports_core/lib/src/activate/np_activate.dart
```diff
@@ -0,0 +1,111 @@
+import 'dart:io';
```
gkc: I'd suggest you move all of this code into the noports_core package so it can be reused elsewhere (since we don't publish the sshnoports package

packages/dart/sshnoports/lib/src/noports_cli/activate/np_activate.dart
gkc: What's the reason for this being a sealed class?

packages/dart/sshnoports/lib/src/noports_cli/activate/np_activate.dart
```diff
@@ -0,0 +1,111 @@
...
+  ///
+  /// Returns: [AtEnrollmentResponse] containing enrollment status and details
+  /// Throws: [ArgumentError] if otp is missing
+  Future<AtEnrollmentResponse> enroll(NoportsParams params);
+}
+
+/// Determines the type of activation: CRAM or APKAM enrollment
+enum NPActivateType { cram, enroll }
+
+class NPActivateImpl implements NPActivate {
```
gkc: Currently the only way of getting an instance of NPActivate is to construct an NPActivateImpl. What is the thinking behind this design?

packages/dart/sshnoports/lib/src/noports_cli/activate/np_activate_params.dart
```diff
@@ -0,0 +1,69 @@
...
+
+class NoportsParams {
```
gkc: Shouldn't be called NoportsParams; the `noports` binary is ultimately intended as a wrapper which starts up any of the noports programs. It has no parameters of its own. Since this class is specifically about the `noports activate` command, it should be named accordingly.

packages/dart/sshnoports/bin/noports.dart
```diff
@@ -0,0 +1,88 @@
...
+      return NoportsCommand.values
+          .firstWhere((cmd) => cmd.commandName == value);
+    } catch (e) {
+      throw ArgumentError('Invalid command: $value');
+    }
+  }
+}
+
+final NPActivate npActivate = NPActivateImpl();
+final NPIssueKeys issueKeys = NPIssueKeysImpl();
```
gkc: Why are these instances in the top-level scope rather than local within the `main` function?

packages/dart/sshnoports/lib/src/noports_cli/util/usage_messages.dart
```diff
@@ -0,0 +1,36 @@
+class UsageMessages {
+  static final mainMenu = '''
+  
+Usage: noports <command> <argument_string>
```
gkc: `<argument_string>` suggests there is only one argument string. But it's actually going to be `noports <command> [arg1, arg2, ...]>`

packages/dart/sshnoports/lib/src/noports_cli/util/np_utils.dart
```diff
@@ -0,0 +1,31 @@
+import 'dart:io';
+
+import 'package:chalkdart/chalk.dart';
+
+void writeError(String message) {
```
gkc: Why are we doing all of this instead of just using an AtSignLogger?

packages/dart/sshnoports/bin/noports.dart
```diff
@@ -0,0 +1,88 @@
...
+
+  if (args.isEmpty) {
+    writeError('You must supply a command');
+    printUsage();
+    exit(1);
+  }
+
+  final command = NoportsCommand.fromString(args[0]);
+  if (args.length == 1) {
+    writeError('You must supply an argument string');
```
gkc: You should leave this up to the parser for the actual command.

packages/dart/sshnoports/bin/noports.dart
```diff
@@ -0,0 +1,88 @@
...
+  displayBanner();
+  int exc = 0;
+
+  if (args.isEmpty) {
+    writeError('You must supply a command');
+    printUsage();
+    exit(1);
+  }
+
+  final command = NoportsCommand.fromString(args[0]);
```
gkc: If you run `noports foo` then you get an ugly stack trace. Need to wrap in its own try-catch block like
```
  final NoportsCommand command;
  try {
    command = NoportsCommand.fromString(args[0]);
  } catch (e) {
    writeError(e.toString());
    printUsage();
    exit(1);
  }
```

packages/dart/sshnoports/bin/noports.dart
```diff
@@ -0,0 +1,88 @@
...
+    writeError(e.toString());
+    stderr.writeln();
+    printUsage(command: command);
+    stderr.writeln();
+    exit(2);
+  } catch (e) {
+    /// ToDo: parse exception message to remove unnecessary Exception prefixes
+    writeError(e.toString());
+    stderr.writeln();
+    stderr.writeln('Please try again or contact support@atsign.com');
```
gkc: You do not need to output this message.

gkc (changes requested):

packages/dart/noports_core/lib/src/activate/np_activate.dart
xavierchanth: +1, we will need some of this on the flutter side.

packages/dart/sshnoports/lib/src/noports_cli/util/usage_messages.dart
xavierchanth: I suggest using `brew --help` / `brew commands` / `brew help [COMMAND]` as an example for a good experience for a router/wrapper cli app

packages/dart/noports_core/lib/src/common/cli_logging_handler.dart
```diff
@@ -0,0 +1,33 @@
...
+  String _getColoredLevel(Level level) {
+    switch (level) {
+      case Level.WARNING:
+        return chalk.yellow('WARN');
+      case Level.SEVERE:
+      case Level.SHOUT:
+        return chalk.red('ERR');
+      case Level.INFO:
+      default:
+        return chalk.blueBright('INFO');
```
gkc: I'd rather you had explicit handling here for FINER and FINEST

packages/dart/sshnoports/bin/noports.dart
srieteja: This has been handled through the latest commits. Note that the removal was not possible as `args` is a fixed-length list. To address this, the main passes a sublist to the `wrappedMain` of individual commands, excluding the first argument.

packages/dart/noports_core/lib/src/activate/np_activate.dart
srieteja: I agree. But since a review is already in progress here, I think waiting until the review is complete and then relocating the code would be efficient.

packages/dart/noports_core/lib/src/commands/issue_keys/issue_keys.dart
```diff
@@ -0,0 +1,251 @@
...
+    );
+
+    final req = EnrollmentListRequestParam()
+      ..deviceName = params.deviceName
+      ..appName = defaultAppName
+      ..namespace = defaultEnrollmentNamespaces.toString()
+      ..enrollmentListFilter = [EnrollmentStatus.pending];
+
+    logger.info('Listening...');
+
```
srieteja, line 205: I've added a retry limit. This is needed because the OTP has a TTL — default is about 5 minutes. Right now, if the user runs issue-keys and then runs activate after the OTP expires, we end up in an endless wait: issue-keys keeps waiting forever for an enrollment that will never show up, and activate can’t create the enrollment because the OTP is already expired. By adding a retry limit, we can control when the waiting stops, handle the timeout cleanly, and tell the user exactly what's going on.

Also note that the `maxRetries` is calculated by dividing `otpExpiry` by `retryInterval`(both in seconds). This syncs the `retries` to the `otpExpiry` .

packages/dart/noports_core/lib/src/common/cli_logging_handler.dart
srieteja: Explicit handling for FINER and FINEST has been added in the latest commit.

packages/dart/sshnoports/lib/src/noports_cli/issue_keys/np_issue_keys.dart
srieteja: The resume mechanism has been added through the latest commits. The mechanism works by serializing the `NPActivateParams` and then reading when available.

packages/dart/sshnoports/bin/noports.dart
srieteja: This change has been made through the latest commits.

packages/dart/sshnoports/lib/src/noports_cli/activate/np_activate_params.dart
srieteja: This class has been renamed to ActivateParams in the latest commits.

packages/dart/sshnoports/bin/noports.dart
srieteja: The first draft has other functions that needed access to these instances. I have now moved this into main() as they're no longer needed in that scope.

packages/dart/sshnoports/lib/src/noports_cli/util/usage_messages.dart
srieteja: I changed the <argument_string> to [options] for a broader scope. I have also updated the help messages to be more user-friendly.

packages/dart/sshnoports/bin/noports.dart
srieteja: Has been addressed through the latest commits.

packages/dart/sshnoports/lib/src/noports_cli/util/np_utils.dart
srieteja: I have introduced a custom logging handler as we discussed.

packages/dart/sshnoports/bin/noports.dart
srieteja: Most edge cases have been identified and are being appropriately handled. This snippet has also been included to handle potential exceptions from NoPortsCommand.fromString().

packages/dart/sshnoports/bin/noports.dart
srieteja: This message has been removed in the latest commits.

packages/dart/noports_core/lib/src/activate/np_activate.dart
srieteja: I have moved to code to noports_core/activate through [8c7f4a3](https://github.com/atsign-foundation/noports/pull/2304/commits/8c7f4a35583c142ed27216b4d84cc84dbe0c5ffe).

srieteja:
Changes in the last batch of commits:

1. Moved the code to noports_core/src/commands/
2. Created new exception `HelpRequestedException` that is thrown when help is requested, help messages displayed when caught.
3. Created a separate arg parser + params class for IssueKeys: `IssueKeysParams`
4. Modified the existing activate params class to a hybrid one: Regex based reading of params from activation_string, then an arg parser for the rest.
5. Removed the state file based resume mechanism; replaced with a check for matching pending enrollments before the usual flow. 
6. Updated help messages to be more descriptive and included examples

packages/dart/noports_core/lib/src/commands/activate/activate_impl.dart
```diff
@@ -0,0 +1,146 @@
...
+  /// 4. If no file exists, returns without modification (safe to proceed)
+  ///
+  ///
+  /// TODO: Replace [AtOnboardingServiceImpl] after ensuring that
+  /// [AtOnboardingService] throws a specific exception when keyFile exists at defaultPath
+  Future<void> _validateAndPrepareKeysFile() async {
+    final String? keysFilePath;
+
+    if (_params.atKeysFilePath == null) {
+      // Use the default path from onboarding service preferences
```
srieteja: Workaround to avoid conflicts with creating atKeys files. Currently fetches the defaultPath from OnboardingServiceImpl.

Changes required: OnboardingService should throw a dedicated exception for key file name collisions, which can then be caught and handled.

packages/dart/noports_core/lib/src/commands/activate/activate_params.dart
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    if (results.rest.isEmpty) {
+      throw ArgumentError(
+        'Activation string is required (e.g. @alice:cram:secret or'
+        ' @alice:enroll:otp:123456)',
+      );
+    }
+
+    final activationString = results.rest.first;
```
xavierchanth: Why .first instead of .single?

packages/dart/noports_core/lib/src/commands/activate/activate_params.dart
```diff
@@ -0,0 +1,134 @@
...
+      )
+      ..addFlag(
+        'help',
+        abbr: 'h',
+        negatable: false,
+        help: 'Show usage information',
+      );
+  }
+
+  static String? _parseAtsign(String input, ActivateType type) {
```
xavierchanth: at_commons has an AtSign type which can be used now. This type calls the equivalent of AtUtils.FixAtsign under the hood.

packages/dart/noports_core/lib/src/commands/issue_keys/issue_keys.dart
```diff
@@ -0,0 +1,225 @@
...
+  /// 1. Initialize AtClient connection
+  /// 2. Generate OTP if not provided
+  /// 3. Set default device name
+  /// 4. Display activation command for user
+  /// 5. Wait for and approve the enrollment request
+  ///
+  /// Returns: 0 on success, 1 on failure
+  Future<int> wrappedMain() async {
+    await _initAtClient();
+
```
srieteja, line 87: If existing matching enrolment present, approves it

packages/dart/noports_core/lib/src/commands/activate/activate_impl.dart
```diff
@@ -0,0 +1,146 @@
...
+
+class Activate {
+  final ActivateParams _params;
+
+  final AtOnboardingService _onboardingService;
+
+  final logger = AtSignLogger('Activate', loggingHandler: CLILoggingHandler())
+    ..level = 'info';
+
+  Activate._(this._onboardingService, this._params);
```
xavierchanth: Nit: the strongly constructor shouldn't be private, most non-binary consumers of this api will prefer this over the factory constructor.

packages/dart/noports_core/lib/src/commands/issue_keys/issue_keys_impl.dart
```diff
@@ -0,0 +1,225 @@
...
+  late final AtClient _atClient;
+
+  final IssueKeysParams _params;
+
+  final AtSignLogger logger = AtSignLogger(
+    'IssueKeys',
+    loggingHandler: CLILoggingHandler(),
+  )..level = 'info';
+
+  IssueKeys._(this._params);
```
xavierchanth: Same here.

packages/dart/noports_core/lib/src/commands/utils/regex.dart
```diff
@@ -0,0 +1,21 @@
+/// Regular expressions for parsing noports command arguments.
+class ActivateRegex {
+  // CRAM authentication: <atsign>:cram:<secret>
+  static final cram = RegExp(r'^(?<atsign>[^:]+):cram:(?<secret>.+)$');
+
+  // Enrollment: <atsign>:enroll:otp:<otp>[:name:<device>:keyfile:<path>]
+  static final enroll = RegExp(
+    r'^(?<atsign>[^:]+):enroll:otp:(?<otp>[A-Za-z0-9]{6})'
```
xavierchanth: This is ordered, which is fine simple is best. Just make sure the Desktop app respects the same order  when it gets implemented there.

Ideally the app consumes the same classes.

packages/dart/noports_core/lib/src/commands/utils/usage_messages.dart
```diff
@@ -0,0 +1,74 @@
...
+  
+Run 'noports ${chalk.cyan('<command>')} --help' for more information
+''';
+
+  static final activateHelp =
+      '''
+${chalk.bold('Usage:')} noports activate ${chalk.cyan('<activation_string>')} [options]
+
+${chalk.bold('Options:')}
+  -k, --keyfile <path>        Path to save atKeys file
```
xavierchanth: Looks like the help wasn't updated

packages/dart/noports_core/lib/src/common/cli_logging_handler.dart
xavierchanth: Why not upstream this to atlogger hidden behind a boolean switch?

packages/dart/noports_core/lib/src/common/help_requested_exception.dart
```diff
@@ -0,0 +1,3 @@
+class HelpRequestedException implements Exception{
```
xavierchanth: Your Pr needs formatting

xavierchanth (dismissed):
Minor changes and some questions.

JeremyTubongbanua:
Feedback:

- `--version` or show version in usage instructions

So there's 3 functions to the NoPorts CLI that I'm aware of

- `noports activate` (CRAM Onboarding)
- `noports issue-keys` (OTP and Auto Approve)
- `noports enroll` (we want this to be a new command, separate from activate)

Right now 1 and 3 are in the same `activate` command, but I think it should be separate commands

packages/dart/noports_core/lib/src/commands/activate/activate_params.dart
srieteja: Addressed through [90a00f6](https://github.com/atsign-foundation/noports/pull/2304/commits/90a00f64245a572d0c4a5daca989208fb03e2ec9)

packages/dart/noports_core/lib/src/commands/activate/activate_impl.dart
srieteja: Constructor made public through [90a00f6](https://github.com/atsign-foundation/noports/pull/2304/commits/90a00f64245a572d0c4a5daca989208fb03e2ec9)

packages/dart/noports_core/lib/src/commands/issue_keys/issue_keys_impl.dart
srieteja: Constructor made public through [90a00f6](https://github.com/atsign-foundation/noports/pull/2304/commits/90a00f64245a572d0c4a5daca989208fb03e2ec9)

packages/dart/noports_core/lib/src/commands/utils/regex.dart
srieteja: This was left behind. The keys file is not accepted through the activation_string anymore; it can be passed as a separate arg with `-k (or) -key-file`. I'll make sure the desktop app works as expected.

packages/dart/noports_core/lib/src/common/help_requested_exception.dart
srieteja: addressed in my last commit

packages/dart/noports_core/lib/src/commands/utils/usage_messages.dart
srieteja: I did update it earlier, I also reworked them again through ([8a4a9d7](https://github.com/atsign-foundation/noports/pull/2304/commits/8a4a9d7ed63ad7593a82669e699a1dbda2b53f6f)).

xavierchanth:
> Right now 1 and 3 are in the same `activate` command, but I think it should be separate commands

Then it no longer solves the original goal.

The original goal is to create a string format that encodes all of the information required to transmit an atSign between machines for use with NoPorts. As soon as you separate the commands, or add options and flags, you have now coupled the use-case to the CLI and it is no longer compatible with the desktop app.

JeremyTubongbanua:
> > Right now 1 and 3 are in the same `activate` command, but I think it should be separate commands
> 
> Then it no longer solves the original goal.
> 
> The original goal is to create a string format that encodes all of the information required to transmit an atSign between machines for use with NoPorts. As soon as you separate the commands, or add options and flags, you have now coupled the use-case to the CLI and it is no longer compatible with the desktop app.

Makes sense,

It was just some feedback I had during my 30min call with Srie. If there was an original vision that I wasn't aware of, then we can ignore my suggestion

packages/dart/noports_core/lib/src/commands/activate/activate_params.dart
srieteja: .first keeps the cli lenient by avoiding errors when extra values are passed. But .single seems clearer and  more straighforward, so I've replaced .first with .single through [90a00f6](https://github.com/atsign-foundation/noports/pull/2304/commits/90a00f64245a572d0c4a5daca989208fb03e2ec9)

xavierchanth:
> > > Right now 1 and 3 are in the same `activate` command, but I think it should be separate commands
> > 
> > 
> > Then it no longer solves the original goal.
> > The original goal is to create a string format that encodes all of the information required to transmit an atSign between machines for use with NoPorts. As soon as you separate the commands, or add options and flags, you have now coupled the use-case to the CLI and it is no longer compatible with the desktop app.
> 
> Makes sense,
> 
> It was just some feedback I had during my 30min call with Srie. If there was an original vision that I wasn't aware of, then we can ignore my suggestion

This is covered in my onboarding document.

packages/dart/sshnoports/lib/src/noports_cli/activate/np_activate.dart
srieteja: In the intial design, I introduced abstraction to guide callers on which methods were intended for use; similar to the current `AtClient` and `AtOnboardingService` patterns. Although after further consideration and discussion on this PR, I have moved away from that pattern and adoped a 'facade with a factory' pattern which fits the usecase better.

packages/dart/sshnoports/lib/src/noports_cli/activate/np_activate.dart
srieteja: I was going for abstraction in the earlier design and I had to prevent the abstract class from being implemented externally. Since the design has moved away from that pattern, there is no longer a need for it to be a sealed class.

packages/dart/noports_core/lib/src/common/cli_logging_handler.dart
srieteja: Will remove CLILoggingHandler from here once a new version of at_utils is published.

packages/dart/noports_core/lib/src/commands/activate/activate_impl.dart
```diff
@@ -0,0 +1,145 @@
...
+      // Use the user-provided path
+      keysFilePath = _params.atKeysFilePath;
+    }
+
+    if (keysFilePath != null && !File(keysFilePath).existsSync()) {
+      return;
+    }
+
+    logger.shout('KeysFile exists at $keysFilePath');
+    _params.atKeysFilePath = promptUser(
```
gkc: what happens if the user puts the same path in again?

packages/dart/noports_core/lib/src/commands/activate/activate_impl.dart
gkc: Need to address this issue fully in this PR. Find the places that MUST have file existence checking (onboard, enroll), ensure the checking always happens there. Remove file existence checking from elsewhere and verify that all the ways we create atKeys files do end up having a file existence check so we don't ever overwrite keys files. As a final safeguard, put a check into the place we finally write an AtKeys file and do not permit overwriting.

packages/dart/noports_core/lib/src/commands/activate/activate_params.dart
```diff
@@ -0,0 +1,131 @@
...
+      )
+      ..addFlag(
+        'help',
+        abbr: 'h',
+        negatable: false,
+        help: 'Show this usage info',
+      );
+  }
+
+  static String? _parseAtsign(String input, ActivateType type) {
```
gkc: Return type should be `Atsign?` not `String?`

packages/dart/noports_core/lib/src/commands/activate/activate_params.dart
```diff
@@ -0,0 +1,131 @@
...
+    final match = ActivateRegex.enroll.firstMatch(input);
+    return match?.namedGroup(ActivateRegexGroups.otp);
+  }
+
+  static String? _parseDeviceName(String input) {
+    final match = ActivateRegex.enroll.firstMatch(input);
+    return match?.namedGroup(ActivateRegexGroups.deviceName);
+  }
+
+  Map<String, String?> toJson() {
```
gkc: Why is `type` not part of the json?

gkc: You are missing a test somewhere to ensure this kind of problem doesn't happen

packages/dart/noports_core/lib/src/commands/issue_keys/issue_keys.dart
```diff
@@ -0,0 +1,211 @@
...
+  /// Returns: 0 on success, 1 on failure
+  Future<int> wrappedMain() async {
+    await _init();
+
+    // Check for matching pending enrollment, approve if found | works like a resume
+    final existingEnrollment = await _fetchMatchingEnrollment();
+    if (existingEnrollment != null) {
+      await _approveEnrollment(existingEnrollment);
+      return 0;
+    }
```
gkc, line 98: personal preference, I would prefer the rest of this code were in an `else` block

packages/dart/noports_core/lib/src/commands/issue_keys/issue_keys.dart
```diff
@@ -0,0 +1,211 @@
...
+      ..enrollmentListFilter = [EnrollmentStatus.pending];
+
+    final results = await _enrollmentService.fetchEnrollmentRequests(
+      enrollmentListParams: requestParam,
+    );
+    logger.finer('Found matching enrollments: ${results.toString()}');
+
+    if (results.isNotEmpty) {
+      logger.info('Enrollment found with id: ${results.first.enrollmentId}');
+      return results.first;
```
gkc, line 236: Should we be asserting there be just one? Or at least an explanation that we can only expect at most one result.

packages/dart/noports_core/lib/src/commands/activate/activate_params.dart
gkc: after further discussion we concluded that toJson is no longer needed and can be removed from this class

packages/dart/noports_core/lib/src/common/cli_logging_handler.dart
gkc: OK

packages/dart/sshnoports/bin/noports.dart
```diff
@@ -0,0 +1,99 @@
...
+  }
+  printVersion();
+  return;
+}
+
+AtSignLogger setupLogging() {
+  displayBanner();
+  AtSignLogger.root_level = 'severe';
+  return AtSignLogger('NoPorts', loggingHandler: CLILoggingHandler())
+    ..level = 'info';
```
gkc, line 100: No way for the user to change the log level?

gkc: add `--verbose | -v` and `--debug` as we have for sshnpd

gkc (changes requested):

packages/dart/noports_core/lib/src/commands/activate/activate_params.dart
srieteja: This has been fixed through b548404653fd76f9109b95787139125c51b39ad7

packages/dart/noports_core/lib/src/commands/activate/activate_impl.dart
srieteja: This would fail again. As discussed previously, I’m working on a better solution in AtOnboardingCli. Once that work is in place, this check will be updated to use the collision handler.

packages/dart/noports_core/lib/src/commands/activate/activate_params.dart
srieteja: The toJson method has now been removed.

packages/dart/sshnoports/bin/noports.dart
srieteja: Verbose and debug args have been added through b548404653fd76f9109b95787139125c51b39ad7.

packages/dart/noports_core/lib/src/commands/issue_keys/issue_keys.dart
srieteja: Yes, I've added the check to ensure there can only be one enrollment request for a given device name. In the case of multiple enrollments, an AtEnrollmentException is now thrown.

packages/dart/noports_core/lib/src/commands/issue_keys/issue_keys.dart
srieteja: I've put the rest of the code in a else block through b548404653fd76f9109b95787139125c51b39ad7.

packages/dart/sshnoports/lib/src/noports_cli/issue_keys/np_issue_keys.dart
gkc: @srieteja As I understand it, this resume is for the approving side. Maybe I'm missing something but, how is resume handled on the enrolling side? e.g. I issue the enroll: command on the enrolling side, then I hit ctrl-C. The enrollment now exists so, if I issue the same command again, I will get an exception. We discussed persisting the AtKeys file in a "pending approval" state so the enrolling side can check if that exists for the same parameters, and if so, goes into its "waiting for approval" loop without re-issuing the enroll command

packages/dart/noports_core/lib/src/commands/activate/activate_impl.dart
gkc: @srieteja are you including the conflict handling in this PR or in a different PR?

packages/dart/sshnoports/lib/src/noports_cli/issue_keys/np_issue_keys.dart
srieteja: Yes, this currently does not handle resume support on the enrolling side. I’ve created this [ticket](https://github.com/atsign-foundation/at_client_sdk/issues/1747)
 to track the feature in the onboarding CLI. I’ll work on that ticket next. Since the resume logic will be part of AtOnboardingService.enroll(), it will be picked up by this tool automatically once a new version of the Onboarding CLI is published.

gkc (approved):
LGTM. Please follow up with a PR to change `.github/workflows/multibuild.yaml` to add in the noports binary into our releases. Please also update the `buildArchive` script in that PR.

PR #2305, opened 2025-11-13
feat: support to run multiple Policy Servers at the same time

### LINKED ISSUE #2280: Add ability to run two instances of policy service on the same host as the same user with the same atSign

When testing redundancy, it is very inconvenient to have to use two separate environments when the limiting factor is just the lock on hive storage. 

Other noports programs allow the hive storage location to be overridden; the policy service needs this too

_No response_

### Additional context

_No response_

srieteja:
Closes https://github.com/atsign-foundation/noports/issues/2280

**- What I did**
- Introduced support for running multiple policy servers on the same system

**- How I did it**
- Introduced a new storagePath parameter in NPAParams.
- Updated npp_atserver to use this parameter, allowing users to specify an alternate storage location for each additional Policy Server instance.
- By enabling separate storage paths, each Policy Server can maintain its own keystore/local secondary. Previously, all instances shared a single commitLog keystore location, which prevented multiple instances from running due to the lockfile.

**- How to verify it**
- Tests pass here.
- Run a Policy Server using the `sshnoports/bin/npp_atserver.dart` and provide a storage location using `-s` argument, and the Policy Server should create its storage instance at the provided location.

feat: support to run multiple Policy Servers at the same time

### DIFF
--- a/packages/dart/noports_core/lib/src/npa/npa_params.dart
+++ b/packages/dart/noports_core/lib/src/npa/npa_params.dart
@@ -93,8 +93,9 @@ class NPAParams {
       'storage-path',
       abbr: 's',
       mandatory: false,
-      help: 'Path to atsign storage directory',
-      // hide: true,
+      help: 'Path to atsign storage directory. Defaults to "~/.atsign/storage/<atSign>/shhnp/single/". '
+          'Running multiple CLI atClient programs with the same storage path is not supported. '
+          'An alternate storage directory can be passed through this argument when running multiple instances.',
     );
 
     return parser;
--- a/packages/dart/noports_core/lib/src/npa/npa_params.dart
+++ b/packages/dart/noports_core/lib/src/npa/npa_params.dart
@@ -11,6 +11,7 @@ class NPAParams {
   final bool verbose;
   final String rootDomain;
   final String homeDirectory;
+  final String? storagePath;
 
   // Non param variables
   static final ArgParser parser = _createArgParser();
@@ -21,6 +22,7 @@ class NPAParams {
     required this.verbose,
     required this.rootDomain,
     required this.homeDirectory,
+    this.storagePath,
   });
 
   static Future<NPAParams> fromArgs(List<String> args) async {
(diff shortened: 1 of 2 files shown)


packages/dart/noports_core/lib/src/npa/npa_params.dart
```diff
@@ -85,6 +89,14 @@ class NPAParams {
       hide: true,
     );
 
+    parser.addOption(
+      'storage-path',
+      abbr: 's',
+      mandatory: false,
+      help: 'Path to atsign storage directory',
+      // hide: true,
```
srieteja: Should this be hidden?

packages/dart/noports_core/lib/src/npa/npa_params.dart
```diff
@@ -33,11 +35,13 @@ class NPAParams {
     return NPAParams(
       authorizerAtsign: authorizerAtsign,
       daemonAtsigns: r['daemon-atsigns'].toString().split(',').toSet(),
-      atKeysFilePath: r['key-file'] ??
+      atKeysFilePath:
+          r['key-file'] ??
           getDefaultAtKeysFilePath(homeDirectory, authorizerAtsign),
```
srieteja, line 40: ignore: dart formatted change

packages/dart/noports_core/lib/src/npa/npa_params.dart
gkc: No, but you should provide explanation in the help text (1) what will this default to (2) explain that you may not run two CLI atClient programs concurrently unless they are each using their own storage directory

gkc (changes requested):

packages/dart/noports_core/lib/src/npa/npa_params.dart
srieteja: The above-mentioned explanation has now been included in the help message through [2accff5](https://github.com/atsign-foundation/noports/pull/2305/commits/2accff5a5c70ec2480fcc63010d7f6a6ad0588b0)

gkc (approved):

PR #2334, opened 2025-11-25
fix: allow policy default value to fallthrough

xavierchanth:
**- What I did**

Based on the issue @kimdimick found while using policy with the config file. 

**- How I did it**

**- How to verify it**

fix: allow policy default value to fallthrough


cpswan (approved):

gkc (approved):
LGTM if Kim d has verified

PR #2378, opened 2025-12-16
fix: Update appName in EnrollmentRequest to use Constants.namespace.

### LINKED ISSUE #2344: NoPorts Desktop default enrollment appname is NoPorts
### Describe the bug

There is inconsistent naming for -A/--arx/appName in npt_flutter for our enrollment requests.

I noticed it here, [onboarding_apkam_dialog.dart](https://github.com/atsign-foundation/noports/blob/8c9a931f4f0d80d044dd5e5217162ceb203114af/packages/dart/npt_flutter/lib/features/onboarding/widgets/onboarding_apkam_dialog.dart#L184)

and since it works end to end I'm assuming the same appName is used for approval and etc.

### Steps to reproduce

1. Open NoPorts Desktop and start an enrollment
2. try and approve via cli with 'noports' as app name
3. fails
4. check `at_activate list -a ...` and the appName is "NoPorts".

### Expected behavior

The appName should be noports, which is uniform across all other cli tools.

### Screenshots

Image

### Smartphones

Latest NoPorts Desktop 
Any CLI

_No response_

### Additional context

_No response_

CurtlyCritchlow:
**- What I did**

-  Updated `appName` in the Apkam Dialog to use `constants.namespace` which store the string 'noports'.

**- How I did it**

**- How to verify it**


packages/dart/npt_flutter/lib/features/onboarding/widgets/onboarding_apkam_dialog.dart
```diff
@@ -182,7 +182,7 @@ class OnboardingApkamDialogState extends State<OnboardingApkamDialog> {
     App.log('Device Name: $deviceName'.loggable);
 
     final enrollmentRequest = EnrollmentRequest(
-      appName: 'NoPorts',
+      appName: Constants.namespace!,
```
Xlin123: can we use a `static String` instead of `static String?` so we don't have to worry about the constants being null

Xlin123 (approved):

PR #2386, opened 2025-12-18
fix: disable background when enrolling

### LINKED ISSUE #1986: NoPorts Desktop - Disable background when authenticating

When authenticating in the desktop app, if someone accidentally clicks on the greyed out background, it closes the authentication dialogue box and the person has to start again by re-entering their atSign. 

Can we make it so that if someone clicks on the greyed out background, nothing happens
Example screen...
Image

_No response_

### Additional context

_No response_

CurtlyCritchlow:
**- What I did**
Prevent Enrolment Dialog from Being disabled
**- How I did it**
Set disable dialog to false and added a back button so user can go back before initiating enrolment.
**- How to verify it**


packages/dart/npt_flutter/lib/features/onboarding/widgets/onboarding_apkam_dialog.dart
```diff
@@ -234,6 +234,27 @@ class OnboardingApkamDialogState extends State<OnboardingApkamDialog> {
...
+      ),
+      onPressed: onboardingStatus != OnboardingStatus.pendingApproval
+          ? () {
+              Navigator.of(context).pop();
+            }
+          : null,
+      child: Text(strings),
+    );
+  }
+
```
xavierchanth: Never use a closure for a widget for "convenience". In doing so, you lose all benefits of Flutter's Widget tree.

There has to be a real reason for doing so, and this doesn't really have one. You can do everything that this function does in a statelessW.

xavierchanth (dismissed):
See my comment about closures

gkc (approved):

PR #2392, opened 2025-12-22
fix: noports desktop switching atsigns refreshes authorisation and policy screen.

### LINKED ISSUE #2195: NoPorts Desktop - switching atSigns doesn't refresh current page
### Describe the bug

This is related to current trunk

When you switch atSigns using the switch atSign feature on the top right of the nav bar, it doesn't refresh the current page (which means you see old state from the old atSign that you originally switched from).

### Steps to reproduce

1. First I '...'
2. Then I '...'
3. And then '...'
4.

### Expected behavior

When you switch atSigns, it should update the state of the current page so that you see the new page and data that is related to the atSign you switched to

### Screenshots

_No response_

### Smartphones

_No response_

_No response_

### Additional context

_No response_

CurtlyCritchlow:
**- What I did**
Made Policy and Authorization Screen refreshes after switching atsign.
**- How I did it**
Updated `post_onboarding.dart` and made `PolicyCubit` global so it can be updated outside of the Policy Screen.
**- How to verify it**

1. Run app, 
2. Navigate to the policy screen, switch atSign and the policy screen will refresh without having to leave the screen.
3. Navigate to the Authenticator screen, switch atSign and the Authenticator screen will refresh without having to leave the screen.

### DIFF
--- a/packages/dart/npt_flutter/lib/app.dart
+++ b/packages/dart/npt_flutter/lib/app.dart
@@ -7,7 +7,6 @@ import 'package:npt_flutter/features/authorisation/cubit/pending_requests_count_
 import 'package:npt_flutter/features/back_up_key/cubit/backup_key_cubit.dart';
 import 'package:npt_flutter/features/back_up_key/repository/backup_key_repository.dart';
 import 'package:npt_flutter/features/features.dart';
-import 'package:npt_flutter/features/policy/repositories/role_repository.dart';
 import 'package:npt_flutter/features/profile_list/cubit/sync_cubit.dart';
 import 'package:npt_flutter/localization/app_localizations.dart';
 import 'package:npt_flutter/routes.dart';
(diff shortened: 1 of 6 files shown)


packages/dart/npt_flutter/lib/app.dart
```diff
@@ -109,6 +106,9 @@ class App extends StatelessWidget {
           BlocProvider<SyncCubit>(create: (_) => SyncCubit()),
           // A cubit which tracks if the atkey is backed up
           BlocProvider<BackupKeyCubit>(create: (ctx) => BackupKeyCubit()),
+          BlocProvider<PolicyCubit>(
+            create: (ctx) => PolicyCubit(ctx.read<RoleRepository>()),
+          ),
         ],
```
CurtlyCritchlow, line 112: Cubit added globally so it is accessible anywhere in the app.

packages/dart/npt_flutter/lib/features/policy/view/policy_view.dart
```diff
@@ -18,11 +18,10 @@ class PolicyView extends StatelessWidget {
...
-    final strings = AppLocalizations.of(context)!;
-    return BlocProvider(
-      create: (context) =>
-          PolicyCubit(context.read<RoleRepository>())..loadRoles(strings),
-      child: PolicyContent(atSign: atSign),
+    return BlocBuilder<PolicyCubit, PolicyState>(
+      builder: (context, state) {
+        return PolicyContent(atSign: atSign);
+      },
     );
```
CurtlyCritchlow, line 25: Instead of creating the policy cubit here, it was referenced from the global policy cubit that was created in `app.dart`

gkc (approved):
LGTM

srieteja (approved):

PR #2403, opened 2026-01-01
feat: NoPorts Desktop. Automatic open of URI if selected after connection established

cconstab:
**- What I did**
Added an option to open an URI once the connection is established
**- How I did it**
Worked with Claude to develop a solution and ensure that everything was extended into profiles
**- How to verify it**
Built and tested on Mac and Windows. - Note URI opening is dependent on OS configurations. Mac OSX has trouble with RDP (windows works fine).
Added Application auto open via URI's once the connection is made. Widgets added to create the URI in the connection config page


gkc (approved):
LGTM

PR #2433, opened 2026-01-16
fix: found and wrapped failed notification with a try block

cconstab:
**- What I did**
Investigated why srvd was crashing every so often and dropping connections
**- How I did it**
Back trace on carsh dump and put in place a try block to catch malformed notifications
**- How to verify it**
tests complete & local testing
prevent crash of srvd on malformed / late notifications

### DIFF
--- a/packages/dart/noports_core/lib/src/srvd/srvd_impl.dart
+++ b/packages/dart/noports_core/lib/src/srvd/srvd_impl.dart
@@ -167,13 +167,13 @@ class SrvdImpl
   }
 
   Future<void> notificationHandler(AtNotification n) async {
-    if (!wellFormedRequest(n)) {
-      logger.shout('Un-handled notification key: ${n.key}');
-      return;
-    }
-
-    logger.shout('Notification key: ${n.key}');
     try {
+      if (!wellFormedRequest(n)) {
+        logger.shout('Un-handled notification key: ${n.key}');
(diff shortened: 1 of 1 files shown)


cconstab:
## Fix: Prevent srvd crash on stale session notifications

### Problem
The srvd daemon was crashing when receiving notifications for sessions that had already completed. This occurred due to a race condition where:
1. A session completes and is cleaned up
2. A late notification arrives for that completed session ID
3. The `wellFormedRequest` method throws an `ArgumentError` for the unknown session ID
4. The exception was not caught, causing the daemon to crash

### Root Cause
In commit `60bc7bddb` (Aug 21, 2025), the `wellFormedRequest` call in `notificationHandler` was placed outside the try-catch block. When it encountered unknown message types (like stale session IDs), it would throw an uncaught exception.

### Solution
Moved the `wellFormedRequest` call inside the existing try-catch block in `notificationHandler`. Now when stale or unknown notifications arrive:
- The `ArgumentError` is thrown by `wellFormedRequest`
- It's caught by the existing exception handler
- The error is logged with full stack trace
- The daemon continues processing other notifications without crashing

### Impact
- Prevents daemon crashes from race conditions with session cleanup
- Maintains full error logging for debugging
- No change to normal operation flow

cpswan (approved):

PR #2435, opened 2026-01-16
fix(noports_core): fix breaking changes from at_commons 5.8.0

Xlin123:
**- What I did**
 - fix breaking changes from at_commons 5.8.0
 - move dependency pin on at_commons to ^5.8.0
  
**- How I did it**
 - change variable types to Atsign where needed
 
**- How to verify it**
 - tests passing again..
fix(noports_core): fix breaking changes from at_commons 5.8.0

### DIFF
--- a/packages/dart/noports_core/pubspec.yaml
+++ b/packages/dart/noports_core/pubspec.yaml
@@ -11,7 +11,7 @@ dependencies:
   args: ^2.4.2
   at_chops: ^2.0.1
   at_client: ^3.10.0
-  at_commons: '>=5.6.1 <5.8.0'
+  at_commons: ^5.6.1
   at_cli_commons: ^3.0.1
   at_utils: ^3.0.19
   at_base2e15: ^1.0.0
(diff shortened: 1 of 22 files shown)


packages/dart/noports_core/pubspec.yaml
```diff
@@ -11,7 +11,7 @@ dependencies:
   args: ^2.4.2
   at_chops: ^2.0.1
   at_client: ^3.10.0
-  at_commons: '>=5.6.1 <5.8.0'
+  at_commons: ^5.6.1
```
cpswan: Unless you're sure that these changes work with older versions this should probably be bumped to ^5.8.0

Xlin123: oh, good catch, thanks

cpswan (dismissed):

packages/dart/noports_core/pubspec.yaml
cpswan: A bunch of the downstream packages are now failing at `pub get` as I think they'll also need to be bumped, and have newer `pubspec.lock` checked in

cpswan: Looks like we also have issues with Docker pulls for the e2e tests. I'd suggest that we add a step that logs into Docker (as we have when we build images) so that the `docker pull` is authenticated and less subject to rate limiting.

cpswan (approved):

PR #2525, opened 2026-02-20
fix: shell startup script error with arguments fixed

### LINKED ISSUE #2522: The sshnpd.sh script fails to translate the required arguments when using TMUX or CRON
### Describe the bug

When using the universal installer and targeting CRON or TMUX the "" around the arguments are interpreted by the shell and the sshnpd is passed arguments that it cannot make sense of.

The issue is nuanced as echoing the command line that is parsed looks just fine, but the shell interprets the delegate_policy=""  and then "$delegate_policy" as an empty process argument and the sshnpd errors out. Equally if you put in "-p (at)cconstab" that gets parsed as one argument to the process and  not two so sshnpd gives a weird error too.

### Steps to reproduce

1. First install in the normal way but on a machine/container that does not have systemd 
2. see the sshnpd not running
3. run the sshnpd.sh by hand and scratch your head for a while

### Expected behavior

The script should run with TMUX/CRON installs..

cconstab:
**- What I did**
fixed issue with staruop scripts
**- How I did it**
moved to conditional statements 
**- How to verify it**
tested locally and with new install

shell startuo scripts error with arguments fixed

### DIFF
--- a/packages/dart/sshnoports/bundles/shell/headless/srvd.sh
+++ b/packages/dart/sshnoports/bundles/shell/headless/srvd.sh
@@ -10,7 +10,9 @@ v="-v"              # Comment to disable verbose logging
 
 sleep 10 # allow machine to bring up network
 export USER="$user"
-while true; do
-	"$binary_path"/srvd -a "$atsign" -i "$internet_address" "$v"
-	sleep 10
+    set -- -a "$atsign"
+    [ -n "$internet_address" ] && set -- "$@" -i "$internet_address"
+    [ -n "$v" ] && set -- "$@" "$v"
+    "$binary_path"/srvd "$@"
+    sleep 10
 done
--- a/packages/dart/sshnoports/bundles/shell/headless/srvd.sh
+++ b/packages/dart/sshnoports/bundles/shell/headless/srvd.sh
@@ -9,8 +9,8 @@ v="-v"              # Comment to disable verbose logging
 # END METADATA
 
 sleep 10 # allow machine to bring up network
-while true; do
 export USER="$user"
+while true; do
     set -- -a "$atsign"
     [ -n "$internet_address" ] && set -- "$@" -i "$internet_address"
     [ -n "$v" ] && set -- "$@" "$v"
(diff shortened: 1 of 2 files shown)


packages/dart/sshnoports/bundles/shell/headless/srvd.sh
```diff
@@ -10,7 +10,9 @@ v="-v"              # Comment to disable verbose logging
 
 sleep 10 # allow machine to bring up network
 export USER="$user"
-while true; do
-	"$binary_path"/srvd -a "$atsign" -i "$internet_address" "$v"
-	sleep 10
+    set -- -a "$atsign"
```
cpswan, line 14: What's the thinking around removing the keep alive loop?

gkc: +1

cconstab: Good catch.. I edited the copy of this file in the build dir until I noticed that is all .gitnored (doh) I obviously messed up with the cut/paste

cpswan: Is there another commit inbound to put it back?

cconstab: I thoughtbis there already ?

cconstab: 9c5a98ac6c25ab03d3451be5a1a2bb9344f8fedd

packages/dart/sshnoports/bundles/shell/headless/srvd.sh
```diff
@@ -9,6 +9,7 @@ v="-v"              # Comment to disable verbose logging
 # END METADATA
 
 sleep 10 # allow machine to bring up network
+while true; do
 export USER="$user"
```
gkc, line 12: Plz move the export to before the while loop

packages/dart/sshnoports/bundles/shell/headless/srvd.sh
gkc: it is

gkc (changes requested):

cconstab (commented):
added this while

packages/dart/sshnoports/bundles/shell/headless/srvd.sh
cconstab: will do

cpswan (approved):

cpswan (approved):

gkc (approved):

PR #2636, opened 2026-04-03
refactor: replace String uses of atsigns with Atsign and renamed the …

### LINKED ISSUE #2536: Refactor the flutter noports app code to use `Atsign` instead of `String` type for atSign variables

Refactor the flutter noports app code to use `Atsign` instead of `String` type for atSigns, so we don't have to worry everywhere about whether there is an `@` at the start or not

Refactor the flutter noports app code to use `Atsign` instead of `String` type for atSigns, so we don't have to worry everywhere about whether there is an `@` at the start or not

_No response_

### Additional context

_No response_

CurtlyCritchlow:
…atsign spelling used.

**- What I did**
-REFACTOR: replaced the use of `String` atsign with the `Atsign` class.
-REFACTOR: where a String atsign is hardcoded or retrieved for packages `.toAtsign()` method is applied to convert it to the `Atsign` class.
-REFACTOR: rename all use of the old 'atSign, AtSign' spelling with new 'atsign, Atsign' spelling.
closes #2536 
**- How I did it**

There are numerous file changes but they are mostly: 

- replacing `String` with `Atsign`, 
- replacing `atSign` with `atsign`, 
- replacing `AtSign` with `Atsign`
- using `.toAtsign()` where our packages still returns atsign as a `String` or when we're converting hard coded or user input `String` to `Atsign`
- Minor code refactoring as a result of  the aforementioned replacements

**- How to verify it**
Run the app and confirm that the app functions the same, and a visual inspection of the code.


packages/dart/npt_flutter/lib/features/policy/models/policy.dart
```diff
@@ -85,47 +100,48 @@ class RoleInProgress {
...
   FetchedRole({
     required this.id,
     required super.name,
     required super.description,
-    required super.daemonAtSigns,
+    @JsonKey(name: 'daemonAtSigns') required super.daemonAtsigns,
     required super.devices,
     required super.deviceGroups,
-    required super.userAtSigns,
+    @JsonKey(name: 'userAtSigns') required super.userAtsigns,
```
CurtlyCritchlow, line 145: the json names remains the same to ensure the data already stored in the atServer can be retrieved.

packages/dart/npt_flutter/lib/features/policy/models/policy.dart
```diff
@@ -35,6 +36,21 @@ part 'policy.g.dart';
...
+  Atsign fromJson(String json) {
+    return json.toAtsign();
+  }
+
+  @override
+  String toJson(Atsign object) {
+    return object;
+  }
+}
+
```
JeremyTubongbanua, line 53: I'm having trouble understanding what this class is doing

I notice it's only used once in the FetchedRole in policy

JeremyTubongbanua (commented):
See comment

packages/dart/npt_flutter/lib/features/policy/models/policy.dart
JeremyTubongbanua: Is this converting a json array (in String format) to an Atsign array (like `['@jeremy', '@curtly']` to `[AtsignInstance, AtsignInstance]`) ?

CurtlyCritchlow: Yes, it implements JsonConverter, it tells FetchedRole when it receives the atsign as a string fromJson (the atServer) converts it to an Atsign by using the `.toAtsign()` method. the `toJson` returns the object because `Atsign` is technically a `String` so it wouldn't throw an error. Without this converter  fromJson wouldn't know how to get `Atsign` from a `String`. If you look at policy.g.dart you'll see how the converter is being applied.

CurtlyCritchlow: yes, it converts `String` to `Atsign` even when nested.

JeremyTubongbanua: Thanks @CurtlyCritchlow !

JeremyTubongbanua (approved):

PR #2627, opened 2026-03-30
feat: auto select best RV

srieteja:
Closes **https://github.com/atsign-foundation/noports/issues/294**

**- What I did**
- Allow noports to auto-select best RV when not provided through args

**- How I did it**
- Introduced `LatencyChecker` uses hard-coded RV IPs to create a dummy socket and calculate client/device latencies to all available RVs.
- Client sends a notification to the device requesting its latencies to all available RVs. 
- Then the client does the same calculation of RV latencies.
- Then selects the RV with the best average of client latency and device latency.

**- How to verify it**
- Tests pass here
- Manually: try running sshnp without the `-r` param and noports should pick up the best RV automatically
- Example from my testing:
For a device running in San Jose and the client running in India:
``` 
INFO|2026-03-31 17:33:24.708027|AtLatencyChecker|Client RV latencies: {@rv_ap: 47, @rv_eu: 241, @rv_am: 216, @rv_oc: 137} 
INFO|2026-03-31 17:33:24.708042|AtLatencyChecker|Daemon RV latencies: {@rv_ap: 179, @rv_eu: 145, @rv_am: 30, @rv_oc: 164} 
INFO|2026-03-31 17:33:24.708050|AtLatencyChecker|Selecting best RV: @rv_ap 
```
**- Description for the changelog**

feat: auto select best RV

PR #2641, opened 2026-04-07
feat: auto-select best RV contd. (srvd discover mechanism)

srieteja:
**- What I did**
- Added a `discover` mechanism for `srvd`. This is a notification exchange between `sshnp` and `srvd` that allows the client to figure out the `srvd` IP Address. This replaces the hard-coded IPs for relays in [PR](https://github.com/atsign-foundation/noports/pull/2627).

**- How I did it**
- `relay_selector.dart` (replaces rv_selector.dart)
    - Rewritten as a proper class that handles the full RV selection flow
       - Fetches the standard relays from URL (or uses a list of srvd atsigns if provided)
       - Sends the `discover` request to each RV in parallel and fetches their IPs
       - Measures latency from client to each of the RVs
       - Sends the list of IPs to the device and fetches the latency device -> RVs
       - Returns the RV with the best combined average latency
- `sshnpd_channel.dart`
   - renamed getRvLatencyDevice -> fetchDeviceRelayLatencies
   - The above-mentioned method now accepts a map of {@srvd -> {IP, port}}
   - The same method now throws FormatException when it receives a malformed latency-check response from the device
- `sshnoports/create_sshnp.dart` 
   - `--srvd` is now optional and accepts a comma-separated list of RVs. RelaySelector is invoked when no single RV is specified.

**- How to verify it**
- Run sshnp as you normally do, except do NOT pass the --srvd arg. sshnp should work as it normally does.

introduce a discover mechanism for srvd to allow clients to figure out the relay's IP


packages/dart/noports_core/lib/src/common/relay_selector.dart
```diff
@@ -0,0 +1,199 @@
...
+
+class RelaySelector with AtClientBindings {
+  @override
+  late final AtClient atClient;
+
+  @override
+  final AtSignLogger logger = AtSignLogger('RelaySelector');
+
+  final String rvServerListUrl =
+      'https://atsign-foundation.github.io/noports/standard_relays.json';
```
gkc: I would rather we have two copies of this json
1. `https://atsign-foundation.github.io/noports/root.atsign.org/standard_relays.json`
2. `https://atsign-foundation.github.io/noports/proxy0001.atsign.org/standard_relays.json`

gkc: i.e. construct the url using the `rootDomain` (but don't include `proxy:` in the constructed URL); this removes need for hard-coded links between `rootDomain` and `standard_relays.json`

packages/dart/noports_core/lib/src/sshnp/util/relay_selector.dart
```diff
@@ -0,0 +1,199 @@
...
+      }) async {
+    List<Atsign> toCheck = [];
+
+    if (rvAtSigns != null && rvAtSigns.isNotEmpty) {
+      toCheck = rvAtSigns;
+    } else {
+      // fetch a list of available RVs from [rvServerListUrl]
+      final standardRelaysRaw = await _fetchStandardRelays();
+      defaultRvList =
+      standardRelaysRaw['standard_relays'] as Map<String, dynamic>;
```
gkc: format

packages/dart/noports_core/lib/src/sshnp/util/relay_selector.dart
```diff
@@ -0,0 +1,199 @@
...
+    logger.info('Fetched latencies for client -> RV: $clientLatency');
+
+    channel ??= SshnpdDefaultChannel(
+      atClient: atClient,
+      params: params,
+      sessionId: Uuid().v4(),
+      namespace: DefaultArgs.namespace,
+    );
+
+    final deviceLatency = await channel.fetchDeviceRelayLatencies(rvIpMap);
```
gkc: Re-order things so that
- the request is sent to the daemon first
- then we do our checking on the client side
- then we wait for the daemon response

packages/dart/noports_core/lib/src/sshnp/util/relay_selector.dart
```diff
@@ -0,0 +1,199 @@
...
+        throw TimeoutException(
+          'Timed out waiting for discover response from $rvAtSign',
+        );
+      },
+    );
+  }
+
+  /// Accepts two maps of RV atsigns to latency in milliseconds, and returns
+  /// the atsign of the RV with the lowest combined average latency.
+  String _lowestAverageLatency(
```
gkc: Should be Atsign return type

packages/dart/noports_core/lib/src/sshnp/util/relay_selector.dart
```diff
@@ -0,0 +1,199 @@
...
+      client.close();
+    }
+  }
+
+  Future<Map<String, dynamic>> _requestRelayIpAddress(Atsign rvAtSign) async {
+    Completer<Map<String, dynamic>> completer = Completer();
+
+    // The srvd will respond with a key 'discover.sshrvd' sharedBy the srvd
+    // e.g. @client:discover.sshrvd@rv_am
+    String regex = 'discover\\.${Srvd.namespace}$rvAtSign';
```
gkc: change topic of the request to 'discover_request' and topic of the response to 'discover_response'

More generally, run all of this with the same atSign for client, relay and daemon to make sure there are no infinite loops caused by requests and responses having the same notification topic.

- "Server" subscribes to `foo.bar.baz` for requests
- "Client" subscribes to `foo.bar.baz` for responses
- "Client" sends notification on `foo.bar.baz`
- "Server" sends response on `foo.bar.baz`
- Now the "Server" receives that message and thinks it's another request

Check your interaction between client(s) and daemon for the same problem

packages/dart/noports_core/lib/src/sshnp/util/relay_selector.dart
```diff
@@ -0,0 +1,199 @@
...
+
+    late StreamSubscription<AtNotification> subscription;
+    subscription = subscribe(regex: regex, shouldDecrypt: true).listen((
+      notification,
+    ) {
+      if (!completer.isCompleted &&
+          notification.from == rvAtSign &&
+          notification.value != null) {
+        final ipAddress = jsonDecode(notification.value!);
```
gkc: make both the request and the response have a json structure; it allows easier later extensibility if we need it

gkc: eg. request value can just be json-encoded `{}` to start with since we don't currently need to do any more; but you could be explicit and say `{"items":["ipaddr","port"]}` and in the future maybe we might have `{"items":["ipaddr","port","fqdn","features"]}`

response is already a map and its structure should reflect the request i.e. a json whose top-level fields have the same names as the items in the request

packages/dart/noports_core/lib/src/sshnpd/sshnpd_impl.dart
```diff
@@ -617,8 +617,23 @@ class SshnpdImpl
...
+    if (notification.value == null) {
+      logger.warning(
+        'No srvd IPs provided in latency check request. Ignoring.',
+      );
+      return;
+    }
+    final Map<String, dynamic> rvServers;
+    try {
+      rvServers = jsonDecode(notification.value!) as Map<String, dynamic>;
+    } on FormatException {
```
gkc: `catch(e)`

packages/dart/noports_core/lib/src/sshnpd/sshnpd_impl.dart
```diff
@@ -628,8 +643,7 @@ class SshnpdImpl
       ..metadata = (Metadata()
         ..isPublic = false
         ..isEncrypted = true
-        ..ttl =
-            10000 // allow only ten seconds before this record expires
+        ..ttl = 10000
```
gkc: ttls are completely irrelevant to pure-messaging notifications; what matters is ttln

gkc (changes requested):

packages/dart/noports_core/lib/src/sshnpd/sshnpd_impl.dart
srieteja: Addressed through ef7c6aca94b3cd5cb149ce4d86704b4fff95f511

packages/dart/noports_core/lib/src/sshnp/util/relay_selector.dart
srieteja: Addressed as part of ef7c6aca94b3cd5cb149ce4d86704b4fff95f511

packages/dart/noports_core/lib/src/sshnp/util/relay_selector.dart
srieteja: request/response prefix added to communication between sshnp -> srvd and sshnp -> sshnpd as part f ef7c6aca94b3cd5cb149ce4d86704b4fff95f511

packages/dart/noports_core/lib/src/sshnp/util/relay_selector.dart
srieteja: Addressed through ef7c6aca94b3cd5cb149ce4d86704b4fff95f511

packages/dart/noports_core/lib/src/sshnp/util/relay_selector.dart
srieteja: Re-ordering has been done through ef7c6aca94b3cd5cb149ce4d86704b4fff95f511

packages/dart/noports_core/lib/src/common/relay_selector.dart
srieteja: domain-based URL construction has been implemented as part of 4d9b201f523c798c82f530ed12fe3b14a78430cd

packages/dart/noports_core/lib/src/sshnp/util/relay_selector.dart
srieteja: All files part of this impl have been formatted using dart format now.

packages/dart/noports_core/lib/src/sshnpd/sshnpd_impl.dart
srieteja: TTLs have been removed from this impl

srieteja:
Tests on the way.

packages/dart/noports_core/lib/src/sshnp/models/sshnp_arg.dart
```diff
@@ -230,7 +230,10 @@ class SshnpArg {
   static const srvdArg = SshnpArg(
     name: 'srvd',
     abbr: 'r',
-    help: 'atSign of srvd daemon',
+    help:
+        'The Socket Rendezvous (SRVD) atSign to use. Omit to auto-select '
```
gkc: "The Relay"

packages/dart/noports_core/lib/src/sshnp/models/sshnp_params.dart
```diff
@@ -202,7 +202,6 @@ class NptParams extends ClientParamsBase
     try {
       AtUtils.fixAtSign(clientAtSign);
       AtUtils.fixAtSign(sshnpdAtSign);
-      AtUtils.fixAtSign(srvdAtSign);
```
gkc, line 205: You need the same logic here as you have in the npt - i.e. fix srvdAtSign if it is non-empty and is not a list

gkc (changes requested):
Couple of small changes to make then this is good to go.

packages/dart/sshnoports/bin/npt.dart
```diff
@@ -78,8 +79,10 @@ void main(List<String> args) async {
       parser.addOption(
         'srvd',
         abbr: 'r',
-        mandatory: true,
-        help: 'The socket rendezvous\'s atSign',
+        mandatory: false,
+        help: 'The Socket Rendezvous (SRVD) atSign to use. Omit to auto-select '
```
gkc: Just say "The Relay" rather than "Socket Rendezvous (SRVD)"

gkc (changes requested):
one more

packages/dart/sshnoports/bin/npt.dart
srieteja: fixed through [bae2452](https://github.com/atsign-foundation/noports/pull/2641/commits/bae245253b533c6e3a5d8f72beaf8cdc0176fc3d)

packages/dart/noports_core/lib/src/sshnp/models/sshnp_arg.dart
srieteja: [bae2452](https://github.com/atsign-foundation/noports/pull/2641/commits/bae245253b533c6e3a5d8f72beaf8cdc0176fc3d)

packages/dart/noports_core/lib/src/sshnp/models/sshnp_params.dart
srieteja: Removed the fixAtSign calls from the NptParams constructor body — they were no-ops since the      
 return values were never assigned. This is acceptable for now because npt.dart's arg parser 
 already fixes the atSigns before constructing NptParams.                                          
                                                                                                 
 The underlying issue is a Dart limitation: clientAtSign, sshnpdAtSign, and srvdAtSign must be     
 forwarded to the superclass constructor via super. parameters, which means they cannot be 
 pre-processed before being passed up. 

A proper fix (likely a factory constructor) will be included in the upcoming tests PR (which I'm currently working on)

srieteja: Reverted the removal to ensure the tests, and added the "fix srvdAtSign if its non-empty and not a list" logic too. Although, the fixAtsign return as never assigned, these fixAtsign calls are acting as a guard.

gkc (approved):

PR #2656, opened 2026-04-14
fix: Better handling of sudo for universal.sh

### LINKED ISSUE #2645: Linux universal.sh no longer prompts to run as sudo
Linux universal.sh used to prompt you to run as sudo, but now it doesn't.

Running `./universal.sh` results in this error:

`sed: couldn't open temporary file /etc/systemd/system/sshnpd.service.d/sediSMJ2p: Permission denied`

There are two paths we can take:

1. Add back the prompting to run as sudo
2. If the prompt to run as sudo was intentionally removed, then we can fix the docs to say "you may need to run as `sudo`"

Additional context and full logs: https://github.com/atsign-foundation/noports/issues/2572#issuecomment-4200353163

cpswan:
Fixes #2645 cc @kimdimick 

**- What I did**

* Removed `sudo` from script, so it's use choice to use it
* Refined the suggest_sudo function
* Corrected handling of `--no-sudo` flag
  * Ensure that `--no-sudo` isn't run with `sudo`

**- How I did it**

Iterative test loops on Ubuntu VM.

Verified that:

* `sh universal.sh --no-sudo` installs locally (no apt package, no systemd config)
* `sudo sh universal.sh` installs apt package, configures systemd and removes any local install
* `sh universal.sh` suggests running with sudo

**- How to verify it**

Paste `universal.sh` from this PR branch into a test environment and run it.

fix: Better handling of sudo for universal.sh

### DIFF
--- a/packages/dart/sshnoports/bundles/universal.sh
+++ b/packages/dart/sshnoports/bundles/universal.sh
@@ -671,7 +671,7 @@ suggest_sudo() {
     echo "This script requires root privileges to continue."
   fi
   echo "Please re-run this script with sudo"
-  exit $?
+  exit 1
 }
 
 check_ssh_keys() {
--- a/packages/dart/sshnoports/bundles/universal.sh
+++ b/packages/dart/sshnoports/bundles/universal.sh
@@ -3,7 +3,7 @@
 # SCRIPT METADATA
 # DO NOT MODIFY/DELETE THIS BLOCK
 script_version="3.2.0"
-sshnp_version="5.14.10"
+sshnp_version="5.14.13"
 repo_url="https://github.com/atsign-foundation/sshnoports"
 # END METADATA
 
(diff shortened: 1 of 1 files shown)


packages/dart/sshnoports/bundles/universal.sh
```diff
@@ -656,22 +653,24 @@ get_atsign() {
...
+    echo "We recommend that you install to systemd, which requires root"
+    echo "privileges to write the unit file"
+    echo
+    echo "Use --no-sudo to install without systemd integration (NOT RECOMMENDED)"
+    echo
+  else
+    echo "This script requires root privileges to continue."
+  fi
+  echo "Please re-run this script with sudo"
   exit $?
```
gkc: should this be exit 1?

cpswan: Yes

JeremyTubongbanua (approved):
lgtm

PR #2682, opened 2026-05-06
Sshnpd doctor updater

LilianGRD:
**- What I did**
Replaced the interactive binary-download flow in the `sshnpd doctor` version check with non-interactive, platform-aware update advice. When an outdated version is detected, the doctor now tells the user exactly which command to run instead of prompting them to download and replace a binary.
**- How I did it**
Added two abstract methods to `PlatformUtils`: `detectPackageManagerInstall()` and `getRecommendedInstallAdvice()`, then implemented them for macOS, Linux, and Windows.
- `detectPackageManagerInstall()` probes the system for known package managers (Homebrew on macOS; dpkg/apt, rpm/dnf/yum, apk, opkg on Linux) and returns the exact upgrade command if the package was installed that way.
- `getRecommendedInstallAdvice()` falls back to platform-appropriate guidance (Homebrew tap, apt/dnf/yum repo setup docs, or the MSI download page on Windows) when no package manager install is detected.
- Removed the entire `_performUpdate()` method, all `stdin` prompts, and the interactive yes/no flow from `version_check.dart`.
- Fixed `diagnostic_runner.dart` so warnings are always displayed even when failures are also present (previously warnings were suppressed if any check failed).

**- How to verify it**
run sshnpd.dart --doctor with or without noports package manager install and check the output.
sshnpd doctor` version check no longer attempts to download or install updates automatically. When a newer version is available it now detects how the package was installed and advises the correct upgrade command (e.g. `brew upgrade noports`, `sudo apt upgrade noports`) or links to the appropriate installation guide for the platform.


cconstab:
@LilianGRD is this still needed to be merged ?

LilianGRD:
> @LilianGRD is this still needed to be merged ?

@cconstab this should be backlogged I think because the features are not up to date and it does a lot of things it should not that I needed to remove but I have put this aside a little.
I’ll come back to it when I have the bandwidth

cconstab:
Mark the PR as not ready esle it could get merged by accident

PR #2704, opened 2026-05-21
feat(npe2e): fresh set of Atsigns, retry and timeout mechanisms

### LINKED ISSUE #2703: npe2e flaking
Add better error catching and retry mechanisms

JeremyTubongbanua:
closes #2703 

**- What I did**
- `runWithRetry` function that runs a test with `maxRetries` (defaults to 3)
- `--max-retries`, maximum # of retries per test, defaults to 3
- `--test-timeout-seconds`, the amount of seconds to wait before giving up on an individual test, defaults to 300s.

- Changed the set of Atsigns running in core_tests, relay_tests and policy_tests
- Atsigns follow this format: `@npe2e_org_<NAME>_jttest`. These Atsigns belong to `root.atsign.org`.

**- How I did it**
- Using Claude

**- How to verify it**
- Ran it on my own machine (using Docker and my own prod Atsigns)
- CI tests pass

feat(npe2e): fresh set of Atsigns, retry and timeout mechanisms


gkc:
@JeremyTubongbanua checks failing?

JeremyTubongbanua:
I fed it a fresh set of Atsigns and that seemed to be the fix.

Either the client or daemon has been problematic, and that might be because they would go under VERY heavy notifications since we originally use the same client and daemon Atsigns for all three test types.

Now, each test type (core_tests, relay_tests, and policy_tests) each use a different set of client and daemon Atsigns.

JeremyTubongbanua:
I think the move here is to merge this PR for now (should be sufficient for 2-3 weeks)

Then will pick up work to implement virtualenv so that a fresh set of local Atsigns are generated every run.

gkc (approved):

PR #2761, opened 2026-07-08
test(npe2e): support running e2e packs against a local atStack

gkc:
## What I did

Enable running the npe2e end-to-end packs against a local atStack, so an atServer
that hasn't been released to production can be tested against the current NoPorts
client and daemon. No effect on CI, which runs against `root.atsign.org`.

## How I did it

- Added a repo-root `.dockerignore`. The npe2e image builds (and the published
  sshnpd image) build with the repo root as their Docker context. It excludes
  melos-managed `pubspec_overrides.yaml` — which pins dependencies to
  host-absolute paths that don't exist inside the container and break the
  in-container `dart pub get` — plus non-source directories. It keeps
  `pubspec.lock` (needed by `dockerhub_sshnpd.yml`'s
  `dart pub get --enforce-lockfile`) and `tools/` (used by
  `Dockerfile.sshnpd-slim`), and is a no-op on a clean CI checkout where the
  overrides are gitignored.
- Injected `--add-host <name>:host-gateway` into each daemon container in
  `startDockerDaemonsParallel`, for every `127.0.0.1 *.atsign.zone` entry in the
  host's `/etc/hosts`. A container can't reach the host on 127.0.0.1, so this
  lets the dockerised daemons resolve the developer's local test domains back to
  the host. It's scoped to `*.atsign.zone`, so unrelated `/etc/hosts` entries and
  deliberate black-holes (e.g. `127.1.1.1 unreachable.atsign.zone`) are left
  alone, and it returns nothing on a clean CI checkout.
- Added `tests/npe2e/README.md` documenting the local-atStack workflow: the
  `/etc/hosts` overrides, starting the atStack and the `srvd` relay (with a
  host-resolvable `-i`, no `--443`), running each pack, and clearing per-run
  state (run directory and leftover containers) between runs on the same commit.

## How to verify it

Against a local atStack on `vip.ve.atsign.zone` (atDirectory on :64, atServers
for @client / @service / @relay, and
`srvd -a @relay -i vip.ve.atsign.zone --root-domain vip.ve.atsign.zone`):

```
dart run tests/npe2e/bin/core_tests.dart \
  --client-atsign @client --daemon-atsign @service --relay-atsign @relay \
  --root-domain vip.ve.atsign.zone --base-directory npe2e_core_tests --batch-size 5
```

Result: core pack 73/73 passed. CI continues to run against `root.atsign.org`
unchanged — both new behaviours are no-ops on a clean checkout.

## Description for the changelog

Support running the npe2e e2e packs against a local atStack.

### DIFF
--- a/tests/npe2e/lib/core_tests/core_tests_docker_utils.dart
+++ b/tests/npe2e/lib/core_tests/core_tests_docker_utils.dart
@@ -52,48 +52,6 @@ Future<List<DockerImage>> ensureDockerDaemonsBuiltParallel({
   );
 }
 
-// Local-run aid: when the npe2e tests run against atServers on the developer's
-// own machine (reached via `/etc/hosts` overrides such as
-// `127.0.0.1 vip.ve.atsign.zone`), mirror those overrides into the daemon
-// containers. A container can't reach the host on 127.0.0.1, so each mapped
-// name is re-pointed at `host-gateway`, Docker's route back to the host.
-//
-// Scoped to `*.atsign.zone` names mapped to 127.0.0.1: that covers the atsign
-// test domains without touching unrelated /etc/hosts entries, and it leaves
-// deliberate black-holes (e.g. `127.1.1.1 unreachable.atsign.zone`) alone. On a
-// clean CI checkout there are no such entries, so this returns nothing (the
-// harness normally runs against the public root.atsign.org).
-List<String> hostGatewayAddHostArgs() {
-  final File hostsFile = File('/etc/hosts');
-  if (!hostsFile.existsSync()) {
-    return const <String>[];
-  }
-
-  final Set<String> names = <String>{};
-  for (String line in hostsFile.readAsLinesSync()) {
-    final int hash = line.indexOf('#');
-    if (hash >= 0) {
-      line = line.substring(0, hash);
-    }
-    final List<String> parts = line.trim().split(RegExp(r'\s+'));
-    if (parts.length < 2 || parts.first != '127.0.0.1') {
-      continue;
-    }
-    for (final String name in parts.skip(1)) {
-      if (name.endsWith('.atsign.zone')) {
-        names.add(name);
-      }
-    }
-  }
-
-  final List<String> args = <String>[];
-  for (final String name in names) {
-    args.add('--add-host');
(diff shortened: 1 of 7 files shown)


tests/npe2e/lib/core_tests/core_tests_docker_utils.dart
```diff
@@ -52,6 +52,48 @@ Future<List<DockerImage>> ensureDockerDaemonsBuiltParallel({
...
+// own machine (reached via `/etc/hosts` overrides such as
+// `127.0.0.1 vip.ve.atsign.zone`), mirror those overrides into the daemon
+// containers. A container can't reach the host on 127.0.0.1, so each mapped
+// name is re-pointed at `host-gateway`, Docker's route back to the host.
+//
+// Scoped to `*.atsign.zone` names mapped to 127.0.0.1: that covers the atsign
+// test domains without touching unrelated /etc/hosts entries, and it leaves
+// deliberate black-holes (e.g. `127.1.1.1 unreachable.atsign.zone`) alone. On a
+// clean CI checkout there are no such entries, so this returns nothing (the
+// harness normally runs against the public root.atsign.org).
```
JeremyTubongbanua: If we're copying the host machine `/etc/hosts` to the docker containers, we may also need to do this under `policy_tests/` and `relay_tests/`

JeremyTubongbanua (dismissed):

tests/npe2e/lib/core_tests/core_tests_docker_utils.dart
gkc: > If we're copying the host machine `/etc/hosts` to the docker containers, we may also need to do this under `policy_tests/` and `relay_tests/`

Yep you're right, thank you.

gkc: Done in eb575424b — promoted `hostGatewayAddHostArgs()` to the shared `docker_utils.dart` and applied it to every atStack-facing container: policy daemon + policy server, and relay daemon + self-relay + client + prime-client

gkc:
sigh. dartssh2 new version has broken our build. will create a separate PR for that fix

JeremyTubongbanua (approved):

PR #2764, opened 2026-07-09
test(ci): repoint legacy e2e_all job to healthy npe2e atServers

srieteja:
## What I did

Repointed the legacy `e2e_all` CI job off the degraded `@e2e_all_v2_*` atServers onto the healthy `@npe2e_org_*` set, so the check goes green again. `e2e_all` had been red on every branch including trunk since ~June 13, 2026 (~3.5 weeks); the root cause is external atServer degradation, not repo code (evidence below). This is a stopgap — the `@e2e_all_v2_*` atServers still need to be repaired or `e2e_all` given its own dedicated atSigns (tracked separately).

## How I did it

`.github/workflows/e2e_all.yaml`, `e2e_all` job only:

- Repoint the 7 positional atSigns to `@npe2e_org_*` (keeping `@rv_am` as the relay). Their atKeys already exist as GitHub secrets — no new provisioning.
- `needs: npe2e` serializes this job after the npe2e matrix (which tears its Docker daemons down, `removeWhenStopped: true`) so the shared atSigns aren't double-booked within a run.
- `!cancelled()` on the `if:` so a flaky npe2e leg can't skip `e2e_all` via the `needs` dependency.

## How to verify it

Evidence the failure is atServer-side, not code:

- The trunk green→red boundary (`aafbadb` → `6d538e7`) contains only a Svelte dependabot bump — no test/harness/product change. The version matrix (`5.9.4/5.11.2/5.13.0`) has been stable in `main.sh` since Nov 2025.
- In the same workflow run, `npe2e (core_tests)` — a port of the same scenarios running the same client versions, but against `@npe2e_org_*` — passes while `e2e_all` on `@e2e_all_v2_*` fails.

Result after repointing: the full legacy suite passed **81/81 in ~6 min** (was 48/81, ~24 min on `@e2e_all_v2_*`). Reviewers can confirm on this PR's `e2e_all` check.

### DIFF
--- a/.github/workflows/e2e_all.yaml
+++ b/.github/workflows/e2e_all.yaml
@@ -25,10 +25,14 @@ jobs:
     # atServer-side, not code. `needs: npe2e` serializes after the npe2e matrix so
     # the shared atSigns aren't double-booked (npe2e daemons removeWhenStopped).
     needs: npe2e
-    # Don't run on PRs from a fork or Dependabot as the secrets aren't available
+    # Don't run on PRs from a fork or Dependabot as the secrets aren't available.
+    # !cancelled() lets this run once npe2e *completes* even if a leg failed
+    # (npe2e is flaky on shared infra today) — we still serialize via `needs` to
+    # avoid double-booking the shared atSigns, but a policy_tests flake must not
+    # skip our hypothesis probe.
     if:
-      ${{ github.event.pull_request.head.repo.fork == false && github.actor !=
-      'dependabot[bot]'}}
+      ${{ !cancelled() && github.event.pull_request.head.repo.fork == false &&
+      github.actor != 'dependabot[bot]'}}
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/e2e_all.yaml
+++ b/.github/workflows/e2e_all.yaml
@@ -20,6 +20,11 @@ on:
 
 jobs:
   e2e_all:
+    # HYPOTHESIS TEST (do not merge): repointed from the degraded @e2e_all_v2_*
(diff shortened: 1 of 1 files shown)


srieteja:
@JeremyTubongbanua tagging you here for visibility

.github/workflows/e2e_all.yaml
```diff
@@ -20,10 +20,19 @@ on:
 
 jobs:
   e2e_all:
-    # Don't run on PRs from a fork or Dependabot as the secrets aren't available
+    # HYPOTHESIS TEST (do not merge): repointed from the degraded @e2e_all_v2_*
+    # atServers to the known-healthy @npe2e_org_* set to confirm the failures are
+    # atServer-side, not code. `needs: npe2e` serializes after the npe2e matrix so
+    # the shared atSigns aren't double-booked (npe2e daemons removeWhenStopped).
```
JeremyTubongbanua: If this is ready for merge, then these comments are stale

JeremyTubongbanua (changes requested):

JeremyTubongbanua:
I'd be okay to merge this for now. We know that e2e_all_v2 Atsigns aren't working properly

JeremyTubongbanua:
But the plan afterwards are to retire e2e_all tests later on

srieteja:
This isn't ready to be merged yet. Will bring this upto speed in some time

.github/workflows/e2e_all.yaml
srieteja: addressed in [9015786](https://github.com/atsign-foundation/noports/pull/2764/commits/901578651a1f85ed6e6c7bd564c83afd3d09833c)

JeremyTubongbanua (approved):

PR #2768, opened 2026-07-10
ci: probe atServer uptime before/after each e2e run

srieteja:
Add tests/npe2e/bin/atserver_info.dart, a zero-dependency probe that resolves each atSign's atServer via the atDirectory and runs the unauthenticated `info` verb (version + uptime). Wire it as before/after steps in both the e2e_all and npe2e jobs, over each job's involved atSigns.

If an AFTER uptime is lower than its BEFORE, that atServer restarted mid-run (OOM/crash) — direct evidence for the older-client e2e failures and an early-warning signal for the shared @rv_am relay that npe2e also depends on.

**- What I did**

**- How I did it**

**- How to verify it**


JeremyTubongbanua (approved):

srieteja:
This is just a diagnostic probe. Not an actual fix. Hence closing the PR

PR #2838, opened 2026-08-18
fix: reconnect NoPorts Desktop tunnels after a device reboot

### LINKED ISSUE #2789: NoPorts Desktop tunnel fails to reconnect after the device reboots
### Describe the bug

NoPorts Desktop tunnel fails to reconnect after the device reboots. The device I am running SSHNPD on is an Ubuntu 24.04.3 device with WiFi and a USB-C Ethernet adaptor. I'm seeing this on the 15.15.1 sshnpd release and the Mac version 1.9.3 of the NoPorts Desktop. 

### Steps to reproduce

1. First I startup the NoPorts Desktop and select my preferred Atsign 
2. Then I startup the RDP profile for the device with the name kdacer315
3. And then I verify the RDP sesion is usable.
4. Then I reboot the device kdacer315
5. Then I login to the device kdacer315 to ensure RDP is active
6. Then I try to start up the RDP session and I note in the NoPorts Desktop app window the connection indicator is spinning, but the Mac Menu Bar menu indicated the connection is active.
7. To get the RDP NoPorts tunnel to reconnect I need to close the NoPorts Desktop app and restart it.
8. There isn't a way to force a connection to close when it can't reconnect.

### Expected behavior

I expected the RDP NoPorts tunnel to reconnect without having to close the NoPorts Desktop application. Also I expected there would be a way to force a connection to close if needed.

### Screenshots

I'm attaching the NoPorts Desktop Log file. The a 2 minute 50 second video is too large for GitHub I guess. This is a link to the file on our Google Drive.

Video file Link on our Google Drive>
https://drive.google.com/file/d/1Gz4a66gjf4aebfGTnqnob44g8iR-ySCk/view?usp=sharing

[NoPorts-Logs-1784737264968.txt](https://github.com/user-attachments/files/30279208/NoPorts-Logs-1784737264968.txt)

_No response_

### Additional context

_No response_

srieteja:
`_NptImpl` completed its `_completer` only on the success paths, so `Npt.done` never completed when startup threw. ProfileBloc awaited it in a `finally` without first calling `npt.close()`, which parked the event handler forever: the profile froze on ProfileStarting, the keep-alive loop stopped iterating, and `_onStop` refused to act because it only accepted ProfileStarted. A tunnel whose daemon rebooted could only be recovered by quitting and relaunching the app.

The npt binary already worked around this - bin/npt.dart calls `await npt.close()` in every catch before its own `await npt.done`. The bloc copied the loop's shape and omitted the close.

noports_core:
- `run`/`runInline` close the session on their way out when they throw, so `done` always completes and callers need not call `close` themselves.
- Complete via `close()` rather than the completer directly in the `sc.done` callback, so closing a started session cannot later double complete.

npt_flutter:
- Collapse `_onStart` and `_retryConnection` into one loop plus a single-attempt `_runSession`, removing ~200 lines of near-duplicate code (the duplication is why the close was missed in one copy). The npt is now always closed and both stream subscriptions cancelled per attempt.
- Keep-alive retries startup failures too, matching the npt binary. This is the reboot case: the daemon is simply absent when the retry fires.
- ProfileStarting is stoppable from both the run button and the tray, via a new `_activeNpt`/`_stopRequested` pair - `_onStop` is a separate handler and cannot reach `_onStart`'s locals, and invalidating the running cubit cannot help when the SocketConnector does not exist yet.
- Drop `TimedOutSocketConnector`; it was never instantiated, so both branches testing for it were dead. Startup timeouts arrive as a thrown TimeoutException.

Fixes #2789

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -53,7 +53,7 @@ jobs:
 
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@5595ccaf912efad79be6eef63a5619ff05969be3 # v4.37.6
+        uses: github/codeql-action/init@ff2f1c621b7f889edc0d3c761ac2e6a3f8cdb0dd # v4.37.7
         with:
           languages: ${{ matrix.language }}
           # If you wish to specify custom queries, you can do so here or in a config file.
@@ -91,6 +91,6 @@ jobs:
           npm run build
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@5595ccaf912efad79be6eef63a5619ff05969be3 # v4.37.6
+        uses: github/codeql-action/analyze@ff2f1c621b7f889edc0d3c761ac2e6a3f8cdb0dd # v4.37.7
         with:
           category: "/language:${{matrix.language}}"
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -67,6 +67,6 @@ jobs:
 
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@5595ccaf912efad79be6eef63a5619ff05969be3 # v4.37.6
+        uses: github/codeql-action/upload-sarif@ff2f1c621b7f889edc0d3c761ac2e6a3f8cdb0dd # v4.37.7
         with:
           sarif_file: results.sarif
--- a/apps/admin/admin_api/pubspec.lock
+++ b/apps/admin/admin_api/pubspec.lock
@@ -608,7 +608,7 @@ packages:
       path: "../../../packages/dart/noports_core"
       relative: true
     source: path
-    version: "6.12.1"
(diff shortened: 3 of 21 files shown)


srieteja:
Superseded by #2932, which carries this same fix against current `trunk`.

The production and test sources on both branches are byte-identical (`npt.dart`, `npt_done_test.dart`, `profile_bloc.dart`, `favorite.dart`), so nothing is lost. #2932 additionally has two test-harness fixes this branch lacks: the bloc providers now wrap `MaterialApp` rather than sitting inside it (`App.log` resolves `LogsCubit` off the Navigator's context, an ancestor of `home`), and the poll helper yields to the real event loop via `tester.runAsync`.

This branch had also drifted behind `trunk` far enough to be actively regressive: merging it would have taken `noports_core` 6.15.0 -> 6.14.2 and rolled `tests/npe2e/pubspec.lock` back (at_chops 3.7.0 -> 3.0.0, socket_connector 2.6.0 -> 2.4.1, noports_core path dep 6.15.0 -> 6.12.2).

`Fixes #2789` has been moved onto #2932, along with the fuller rationale from this description.

PR #2839, opened 2026-08-18
fix: share event logging config once as cached key instead of on every heartbeat

### LINKED ISSUE #2744: Windows event viewer displays "{"eventLoggingConfig":null} " every few minutes
### Describe the bug

Windows event viewer displays the following  every few minutes. The @21cluster is my policy manager atsign.

INFO|2026-06-29 11:46:03.711332| sshnpd |config received from @21cluster: {"eventLoggingConfig":null} 
SHOUT|2026-06-29 11:46:03.711332| sshnpd |Config from policy: @55commercial:config.kdtpe16.devices.policy.sshnp@21cluster : {"eventLoggingConfig":null} 
SHOUT|2026-06-29 11:46:03.711332| sshnpd |No eventLoggingConfig 
INFO|2026-06-29 11:46:03.842927| sshnpd |config received from @21cluster: {"eventLoggingConfig":null} 
SHOUT|2026-06-29 11:46:03.842927| sshnpd |Config from policy: @55commercial:config.kdtpe16.devices.policy.sshnp@21cluster : {"eventLoggingConfig":null} 
SHOUT|2026-06-29 11:46:03.842927| sshnpd |No eventLoggingConfig

### Steps to reproduce

1. First I setup the policy manager atSign using the sshnpd.yaml file on a windows 11 pro system
2. Then I restarted the NoPorts service
3. And then I checked the windows Event Viewer for possible errors in the configuration sshnpd.yaml file.
4. The policy manager is running on a Linux system.

### Expected behavior

I didn't expect to see any errors I only was expecting to see the heartbeat messages.

### Screenshots

Image

_No response_

### Additional context

_No response_

### LINKED ISSUE #2794: Policy service re-sends the event-logging config on every daemon heartbeat — share it once as a cached key instead
### Summary

The policy service replies to **every** daemon heartbeat with its event-logging config. The heartbeat is every 5 minutes, so at the common ratio of 1 policy atSign per 100 daemons the policy atServer carries **57,600 notifications/day** (28,800 in, 28,800 out) before a single session happens — roughly **97% of everything that atServer does**. An auth check, by comparison, costs 6 notifications.

The re-send conveys nothing. Share the config **once** as a cached key instead.

### Why the current re-send is redundant

Two facts make this stronger than "the config changes rarely":

1. **The config is stored in an immutable atKey.** `getOrCreateConfig` creates `config.events.logging.sshnp@events` with `..metadata.immutable = true` and a randomly-generated topic (`event_mixins.dart:33`). For a given events atSign it *cannot* change — the atServer rejects the overwrite.
2. **The policy reads it once, ever.** `elc` is assigned in `NPAImpl.run()` (`npa_impl.dart:94`) and never re-read; there is no refresh path. So within a policy process lifetime the value being re-sent 288×/day/daemon is provably constant. It can only differ across a policy restart pointed at a different events atSign.

The heartbeat payload is also heavier than it needs to be: it's the full `pingResponse` — device name, group, two version strings, eight feature flags, auth modes, `permitOpen`, `npCpVersion`, `publicSigningKeyUri`. It's built once in the constructor (`sshnpd_impl.dart:156`) and never mutated, so every heartbeat after the first is byte-identical.

### Proposed design

On first contact, the policy shares the config at:

```
@daemon:config.events.logging.sshnp@policy    ttr = -1, ccd = true, ttl = null
```

`ttr = -1` means the daemon's own atServer keeps a `cached:` copy indefinitely. The daemon then reads the config from its own atServer with **zero cross-atServer traffic**, and persists nothing itself — it just checks whether it has the key. The policy needs no per-daemon delivery state and no device inventory.

The heartbeat survives as a **presence beacon only** — lighter payload, and its interval becomes an admin-UI decision rather than a config-delivery one.

### Why it works — verified against at_server

Edits still propagate, and the fan-out is driven by who holds a cached copy, which the atServer already tracks:

1. Policy `update:`s the source → `notify(...)` is called unconditionally with the metadata (`abstract_update_verb_handler.dart:151`).
   *(The `ttr! > 0` guard just above it is a validation check rejecting updates where `sharedBy` isn't the current atSign — not a gate on the notify.)*
2. `ttr:-1:ccd:<bool>` goes on the wire whenever `atMetaData.ttr != null` — no `> 0` condition (`notification_manager_impl.dart:306-308`).
3. Daemon's atServer: `validateTTR(-1)` returns `-1`, non-null, so `notify_verb_handler.dart:157` does not early-return — it falls through to `_storeCachedKey`, and **if the cached key already exists it merges metadata and updates it** (`:186-195`).
4. `ccd:true` → deleting the source removes the cached copy (`:172-178`).
5. Daemon reads via `lookup:` → `_fetchDataOwnedByOtherAtSign` serves from `cached:@daemon:…@policy` (`lookup_verb_handler.dart:227-232`), and `cacheManager.get` honours `ttr == -1` as cache-forever (`cache_manager.dart:259-260`). It falls through to a remote lookup only if the copy is absent or `bypassCache` is set.

The periodic cache-refresh job skips `ttr == -1` (`cache_manager.dart:86-87`) — correct, and irrelevant, since propagation is push-on-update.

### Changes needed

Both halves already exist and already use this exact key shape — this is precisely how **npevents → policy** works today. The fix extends it one hop.

- [ ] `event_mixins.dart:78-84` — `shareEventLoggingConfigWithAtsigns` already writes `$theirAtsign:config.$namespace$myAtsign`. Add `..metadata.ttr = -1` and `..metadata.ccd = true`. This also improves the existing npevents→policy hop, which currently re-fetches across atServers on every read.
- [ ] `npa_impl.dart:15` — `NPAImpl` mixes in `AtEventLogger` only; `shareEventLoggingConfigWithAtsigns` lives on `AtEventListener`. Add the mixin, or move the method.
- [ ] `sshnpd_impl.dart` — at startup call `getEventLoggingConfig(atSign: policyManagerAtsign!, namespace: DefaultArgs.eventLoggingNamespace)`. `SshnpdImpl` already mixes in `AtEventLogger`, so this method is already available — no new API. On `AtKeyNotFoundException`, signal the policy (see below).
- [ ] `npa_impl.dart:106-130` — stop replying unconditionally to every heartbeat.
- [ ] Slim the heartbeat payload and raise `DefaultSshnpdArgs.policyHeartbeatFrequency` (`default_args.dart:72`) once the presence SLA is agreed.

### First-contact trigger

Something has to trigger the initial share. Keep it stateless on both sides: **the daemon checks its own cached copy at startup and only signals the policy if it's absent** (a flag on the heartbeat, or a one-shot request). Policy shares, daemon retries. No per-daemon delivery state on the policy, no persistence on the daemon, and a policy restart costs nothing because nobody is tracking anything.

### Notes / risks

- **There's an existing scar from this mechanism in-tree.** `device_info` is shared with `ttr:-1, ccd:true` (`sshnpd_impl.dart:1801-1804`), and the client reads it with `bypassCache = true` plus a comment about "probably a cached key which should have been deleted" (`sshnpd_channel.dart:395-400`). Stale cached shared keys have bitten us before. An immutable config is far lower risk, but keep `ccd:true` and give the daemon a `bypassCache` escape hatch.
- **Cache refresh inherits notification delivery semantics**, not strong consistency. If a daemon's atServer is unreachable when the config changes, the refresh rides the normal notification retry path. Fine for a value that only changes on operator action.
- **The heartbeat interval is now purely a presence decision.** It currently doubles as daemon liveness for the admin UI — `onDaemonEvent` builds a `DaemonHeartbeat` feed from it (`admin/impl.dart:167-177`). That's the only thing that needs deciding before picking a new interval.
- Two bits of dead code in this path, worth removing while we're here: `NPAImpl` builds an event map from every heartbeat and discards it (`npa_impl.dart:107-111`), and `PolicyInfoRpcRequestHandler` takes an `elc` parameter it never uses (`npa_impl.dart:175, 184`).

### Expected result

Per 100 daemons, steady state:

| | now | with fix, 5 min beacon | with fix, 60 min beacon |
|---|---:|---:|---:|
| policy atSign, notifications/day | 57,600 | 28,800 | 2,400 |
| return leg (config out) | 28,800 | 0 | 0 |

Removing the reply alone halves it; the interval change takes it to a **~96% cut**.

### Related

- Fixes #2744 — the customer-visible symptom of this. The daemon logs `Config from policy: … {"eventLoggingConfig":null}` at SHOUT every 5 minutes, which lands in the Windows event viewer. No repeated config notification, no repeated log line.
  - Side observation from that issue's log: the config arrives **twice**, ~130ms apart. Worth confirming whether that's two `sshnpd` processes on the same atSign — every process on an atSign receives every notification for it, since the monitor is opened with no server-side regex — or a genuine double-send from the policy.

JeremyTubongbanua:
closes https://github.com/atsign-foundation/noports/issues/2744
closes https://github.com/atsign-foundation/noports/issues/2794

**- What I did**

`event_mixins.dart` — Added ttr = -1 and ccd = true to the key metadata in shareEventLoggingConfigWithAtsigns. This makes the recipient's atServer cache the config indefinitely and delete the cache when the source is deleted.

`npa_impl.dart` (policy side):
- Added AtEventListener mixin so the policy can call shareEventLoggingConfigWithAtsigns
- Heartbeat handler no longer replies with config on every heartbeat. Instead, it only shares config (via cached key) when the daemon explicitly requests it via needEventLoggingConfig: true
- Removed dead code: the unused event map construction, and the unused elc field on PolicyInfoRpcRequestHandler

`sshnpd_impl.dart` (daemon side):
- At startup, tries to read the cached event logging config from the policy via getEventLoggingConfig
- On each periodic heartbeat, retries fetching the config if elc is still null
- Heartbeat payload includes needEventLoggingConfig: true when elc is null
- handlePolicyConfigNotification log levels changed from SHOUT → INFO (fixes #2744's noisy Windows event viewer entries)

**- How I did it**

- Claude Opus 4.6 [1m]

**- How to verify it**


JeremyTubongbanua:
Once merged to trunk, this needs to be a MINOR release

- sshnpd — fetches cached config at startup, adds needEventLoggingConfig flag to heartbeat, fixes SHOUT→INFO log levels
- npa (policy) — stops replying to every heartbeat, shares config via cached key on request

JeremyTubongbanua:
---

Re #2744 — No more SHOUT-level "No eventLoggingConfig" in Windows Event Viewer

Daemon log:
```
INFO|2026-08-18 11:23:53.898311| sshnpd |No event logging config from @tastelessbanana yet
INFO|2026-08-18 11:23:53.899245| sshnpd |Sending heartbeat to policy service @tastelessbanana
INFO|2026-08-18 11:23:54.286938| sshnpd |Daemon is running
```


- Config absence is logged at INFO, not SHOUT — Windows Event Viewer won't pick it up.
- Zero instances of SHOUT|...|Config from policy: or SHOUT|...|No eventLoggingConfig in the entire daemon log.

Re: #2794 — Policy no longer re-sends config on every heartbeat

Policy log:
```
INFO|2026-08-18 11:23:54.424860|PolicyServiceWithAtClient|Received device heartbeat from @soccer99
```

- Policy received the heartbeat but did not send a config notification back.
- Zero instances of Sending config notification in the entire policy log.
- The old behavior (SHOUT|...|Sending config notification $configKey : {"eventLoggingConfig":null}) is completely eliminated.

---

gkc (dismissed):

srieteja (changes requested):
Nice change — swapping the per-heartbeat push for a cached pull is the right shape, and the key names line up correctly on both sides (`@daemon:config.events.logging.sshnp@policy`). The `ttr=-1`/`ccd=true` caching behaviour is already verified against at_server in #2794, so I'm taking that as given. Comments below are mostly about rollout and the fact that the new code path isn't exercised yet.

## Blocking

**1. Old daemon + new policy silently loses event logging**

`npa_impl.dart:110` only shares when the heartbeat carries `needEventLoggingConfig == true`. Daemons that haven't been upgraded never set that flag, and the push path they used to rely on is gone — so they get nothing. Since the policy service is upgraded centrally and daemons live on customer devices, the normal rollout order is exactly the one that breaks: every existing daemon goes dark on events until it's updated.

Either we make the rollout order explicit (policy last), or the policy shares once per previously-unseen `n.from` regardless of the flag — memoize the set of atSigns already shared with so it stays a one-shot per daemon.

**2. The new path isn't verified yet**

The logs in the PR description show:

```
INFO|... sshnpd |No event logging config from @tastelessbanana yet
```

plus zero `Sending config notification` on the policy side. That means the test policy had `elc == null`, so `shareEventLoggingConfigWithAtsigns` never fired and the cached-key read was never exercised — everything demonstrated so far is the *absence* path.

CI doesn't cover it either: `npe2e policy_tests` passed under the old notify behaviour too, so it isn't touching the changed code, and there's no `test/npa`.

Could you post a run with the policy started under `--event-logging-atsign`, showing the daemon log `Fetched event logging config: <elc>`? That's the one thing I'd want to see before approving.

**3. #2744's own scenario still logs on a 5-minute cadence**

In the reported setup the policy has no event logging, so `elc` stays null forever and the timer at `sshnpd_impl.dart:319` re-runs `_fetchEventLoggingConfig()` every period — emitting `No event logging config from … yet` at INFO (`:1912`) plus a wasted cross-atServer `lookup:` each time. The reporter's Event Viewer capture contains INFO lines, so dropping from SHOUT to INFO doesn't make it invisible.

Suggest logging on state *change* only: once on the first miss, and again on the transition to found. If the intent is for this PR to close #2744, this one matters.

## Worth fixing before merge

**Heartbeat is now behind an unbounded network call.** At startup `_fetchEventLoggingConfig()` is awaited before the first `_sendHeartbeatToPolicy()` and before `Daemon is running` (`sshnpd_impl.dart:312-326`), and the timer body does the same. A slow or hung fetch delays the presence beacon the admin UI's heartbeat feed depends on — which #2794 calls out separately. Swapping the order is free: heartbeat first, then fetch.

**Up to 5 minutes of dropped events on a fresh daemon.** It starts, misses, and waits a full period before trying again. Two options:
- After a heartbeat that set `needEventLoggingConfig: true`, retry the fetch once a few seconds later. No new assumptions.
- Have the daemon `subscribe` to `config\.events\.logging\.sshnp@policy` and set `elc` on notification. Cheaper, and it also handles the re-point case below — but worth confirming the monitor surfaces that notification to the client first, since there's no precedent in-tree (`NPAImpl.run()` reads its own config with a plain `get`, never a subscribe).

**Config changes now require a daemon restart.** The fetch only runs while `elc == null`. Given `getOrCreateConfig` sets `immutable = true`, that's fine for a fixed events atSign — but a policy restarted against a *different* events atSign won't reach running daemons. Not a bug, just worth a CHANGELOG line.

**Version bump / CHANGELOG / lockfile sync aren't in the diff.** Agreed on cutting a release once this lands — one thing to flag: dropping `required this.elc` from `PolicyInfoRpcRequestHandler`, which is exported via `lib/npa.dart`, is source-breaking for anyone constructing it. Might warrant more than a minor.

## Nits

- `npa_impl.dart:108-116` — worth wrapping the whole `.listen` body in a try/catch. `jsonDecode(n.value!)` assigned to `Map<String, dynamic>` throws on a malformed sender, and the `await shareEventLoggingConfigWithAtsigns(...)` is unprotected; either becomes an unhandled async error in the listener.
- `_sendHeartbeatToPolicy` already guards `policyManagerAtsign == null` internally, but `_fetchEventLoggingConfig` uses `policyManagerAtsign!` and pushes the check out to both call sites. Moving it inside drops both.
- `npa_impl.dart:99` still uses `logger.shout('Fetched AtEventLogger config …')` — looks like it was missed by the shout→info sweep on the daemon side.
- Deferred #2794 items (slimmer heartbeat payload, higher `policyHeartbeatFrequency`, a `bypassCache` escape hatch) are fine to punt, but the traffic saving actually realised here is roughly 50%, not the ~96% the issue projects. Worth stating in the PR body so it doesn't read as fully done.

JeremyTubongbanua:
npevents.log (@events_jttest)

```
- 10:59:58 - Created new EventLoggingConfig {atSign: @events_jttest, topic: 44f3qkbv.events.logging.sshnp, ttln: 3600000}
- 10:59:59 - Shared config with @policy02_jttest
- Rest is Monitor heartbeats (service stayed healthy)
```

npp.log (@policy02_jttest)

```
- 11:00:08 - Fetched AtEventLogger config from @events_jttest
- 11:00:27 - First device heartbeat from @device_jttest, shared EventLoggingConfig with @device_jttest (only happens once)
- 11:05:27, 11:10:27, 11:15:27... - Subsequent heartbeats from @device_jttest every 5 min, but no re-sharing of config (confirms _sharedElcWith dedup set works)
```

sshnpd.log (@device_jttest)

```
- 11:00:26 - "No event logging config from @policy02_jttest yet" (logged once)
- 11:00:26 - First heartbeat sent to policy service
- 11:05:27 - "Fetched event logging config: {atSign: @events_jttest, topic: 44f3qkbv.events.logging.sshnp, ttln: 3600000}" (config arrives on next poll cycle)
- 11:05:27+ - Heartbeats continue every 5 min, no repeated "no config" messages
```

The logs confirm all three blocking fixes are working:
1. Config sharing happens only once per daemon (not on every heartbeat)
2. Config propagates end-to-end within one heartbeat cycle
3. "No event logging config" message appears exactly once, not repeatedly

packages/dart/noports_core/lib/src/npa/npa_impl.dart
```diff
@@ -102,31 +104,19 @@ class NPAImpl with AtClientBindings, AtEventLogger implements NPA {
...
-          configKey,
-          configValue,
-          checkForFinalDeliveryStatus: false,
-          waitForFinalDeliveryStatus: false,
-          ttln: Duration(hours: 1),
-        );
+        if (n.value == null) return;
+
+        if (elc != null && !_sharedElcWith.contains(n.from)) {
+          _sharedElcWith.add(n.from);
```
srieteja: The atSign is recorded as served before the share is attempted:

`shareEventLoggingConfigWithAtsigns` is a bare `atClient.put` with no error handling of its own — the immutable-swallow at `event_mixins.dart:58` is in `getOrCreateConfig`, a different method, so nothing absorbs a failure on this path. If that put throws, the daemon stays permanently marked as served and is never retried: dark until the policy process restarts.

Would suggest something like this:
```dart
try {
  logger.info('Sharing event logging config with ${n.from}');
  await shareEventLoggingConfigWithAtsigns(...);
  _sharedElcWith.add(n.from);
} catch (e) {
  logger.warning('Failed to share event logging config with ${n.from}: $e');
}
```

packages/dart/noports_core/lib/src/sshnpd/sshnpd_impl.dart
```diff
@@ -308,11 +310,19 @@ class SshnpdImpl
...
     Timer.periodic(
       DefaultSshnpdArgs.policyHeartbeatFrequency,
-      (_) async => await _sendHeartbeatToPolicy(),
+      (_) async {
+        if (elc == null && policyManagerAtsign != null) {
+          await _fetchEventLoggingConfig();
+        }
+        await _sendHeartbeatToPolicy();
+      },
     );
```
srieteja: **Five minutes of events dropped on every daemon start.**

Your log measures it: miss at 11:00:26, config arrives at 11:05:27. The cause is the ordering here — the policy shares its config *in response to a heartbeat* (`npa_impl.dart:111`), so at startup the daemon fetches before anything has been shared, misses, sends the heartbeat that triggers the share, then doesn't look again for a full period. The config is sitting in its atServer seconds after the miss.

Two smaller things live in the same block: `_fetchEventLoggingConfig()` is awaited *before* the first `_sendHeartbeatToPolicy()` and before `Daemon is running`, so a slow fetch delays the presence beacon; and the `policyManagerAtsign` null check is duplicated here even though `_sendHeartbeatToPolicy` already guards internally.

All three collapse into one cycle:

```dart
static const _configRetryDelay = Duration(seconds: 5);

/// One policy cycle: beacon first, then close the gap on the config.
///
/// The policy shares its event logging config in response to a heartbeat, so
/// when we don't have it yet it lands moments after ours — re-check then,
/// rather than waiting out a full period.
Future<void> _policyCycle() async {
  if (policyManagerAtsign == null) return;
  await _sendHeartbeatToPolicy();
  if (elc != null) return;

  await Future.delayed(_configRetryDelay);
  await _fetchEventLoggingConfig();
}
```

```dart
await _policyCycle();
Timer.periodic(DefaultSshnpdArgs.policyHeartbeatFrequency, (_) => _policyCycle());
```

packages/dart/noports_core/lib/src/npa/npa_impl.dart
JeremyTubongbanua: Fixed by moving `_sharedElcWith.add(n.from)` after the successful `await shareEventLoggingConfigWithAtsigns()` call, wrapped in try/catch so a failed put doesn't permanently blackhole that daemon.

packages/dart/noports_core/lib/src/sshnpd/sshnpd_impl.dart
JeremyTubongbanua: Fixed by introducing _policyCycle() that sends heartbeat first, then retries the config fetch after a short delay instead of waiting a full period.

srieteja (approved):

PR #2841, opened 2026-08-19
fix: Make --help/--version output help2man-friendly so man pages render correctly

### LINKED ISSUE #2650: All of the man pages, but at_activate in the 5.14.13 release have issues when compared to the --help output
### Describe the bug

It looks like the np_admin man page has the "-a --atsign" listed twice. The second "-a --atsign" I think should be -r --root-server. Also, the description for the first "-a --atsign" isn't correct.

extracted from the np_admin man page:

-a,  --atsign  (mandatory)     The  atSign  to  use  -r,  --root-server
       Root server (e.g., [root.atsign.org:64](http://root.atsign.org:64/))

              (defaults to "[root.atsign.org:64](http://root.atsign.org:64/)")

       -v,  --verbose                More  logging  -d,   --legacy-root-domain
       Deprecated   abbreviation   option   for  --root-server  -b,  --bind-ip
       Bind to something other than the default

              (defaults to "127.0.0.1")

       -p, --bind-port             Bind to something other  than  the  default
       port

              (defaults to "3000")

       --help Usage instructions

       -a,  --atsign  (mandatory)     The  atSign  to  use  -r,  --root-server
       Root server (e.g., [root.atsign.org:64](http://root.atsign.org:64/))

### Steps to reproduce

1. First I "man np_admin"
2. Then I reviewed th page and compaired it to the --helpoutput.

### Expected behavior

I expected the page to be similar to the help output with more details.

### Screenshots

_No response_

_No response_

### Additional context

_No response_

cconstab:
- [x] Bug fix

### What this PR does

Fixes #2650 — all of the generated man pages (except `at_activate`'s) were mangled: every page was titled `VERSION(1)`, options ran together as plain text (making descriptions appear against the wrong flags), and `np_admin`'s option list appeared twice with a stray `Could not find an option named "--version".` at the bottom.

Man pages are generated in `multibuild.yaml` by `help2man`, which runs each binary with `--version` and `--help` and parses the output. All of the reported symptoms were reproduced locally with `help2man 1.49.3` and traced to three independent root causes:

1. **`--version` printed `Version : 5.14.13 (core: 6.12.0)` to stderr.** `help2man` titles the man page using the first word of the `--version` output — hence every page was `VERSION(1)`. GNU convention is `progname version` on stdout.
2. **`np_admin`, `npevents`, `npp_atserver` and `npp_file` had no `--version` flag at all.** `help2man` runs `--version` first; the resulting unknown-option error dumped the full usage to stderr, and (because we pass `--no-discard-stderr`) that dump was swept into the man page alongside the real `--help` output — that's the duplicated option list.
3. **`ArgParser.usage` output isn't help2man-parseable.** It places abbreviated options (`-a, --atsign`) at column zero with no `Options:` header, so `help2man` doesn't recognise them as options and flows them together as prose.

### Changes

**New shared helpers**

- `noports_core/lib/src/common/cli_help.dart` (exported from `utils.dart`):
  - `binaryName` — the running program's file name, derived from `Platform.script` so it always matches what the user invoked (works for compiled binaries and `dart run`).
  - `formatCliHelp(...)` — wraps an `ArgParser.usage` string in the GNU `Usage:` / description / `Options:` layout, with the options block indented two spaces so `help2man` parses each option into the OPTIONS section. Unit tested in `noports_core/test/cli_help_test.dart`.
- `sshnoports/lib/src/print_version.dart` — `printVersion()` now emits `sshnpd 5.15.1 (core 6.12.1)`-style output to stdout by default (with an `IOSink` parameter for error paths).

**Binary entrypoints** (all 12 shipped binaries, plus `npp_client`)

- `--help` and `--version` now print to **stdout** and exit 0 (GNU convention; previously most printed to stderr).
- Usage-on-error still goes to **stderr**, now in the same formatted layout, followed by the error message. Exit codes are unchanged throughout.
- Added `--version` to `np_admin`, `npevents`, `npp_atserver`, `npp_file` (via a new flag on `NPAParams.parser`).
- `npp_file --help` previously **crashed** (unhandled `ArgumentError` for the missing mandatory `--yaml`); it now has proper help/version/error handling.
- `np_admin` handles `--help`/`--version` before its port-bind sanity check (it used to bind a socket before printing help). It reports the `noports_core` version, since `admin_api` has no release version of its own — flagging this as a discussion point.
- `noports`: help/version are handled before the ASCII banner, and ANSI escape sequences are suppressed when stdout is not a terminal (`chalk.level = 0`) — its man page was previously full of raw escape codes, and piping `noports --help` into anything else had the same problem.

**Man page boilerplate**

- New `packages/dart/sshnoports/man/<binary>.h2m` include files (one per binary in `DBINS`, including `at_activate`) providing proper `NAME` one-liners — addressing the "you can't tell which man page you're looking at" comment — plus `EXAMPLES` (taken from the docs), `SEE ALSO` cross-references and `REPORTING BUGS`.
- `multibuild.yaml` passes `--include man/${BINARY}.h2m` to `help2man`. `--no-discard-stderr` is retained since `at_activate` (built from `at_onboarding_cli`, unchanged here) is known to work with it.

### Before / after (np_admin, rendered with man)

Before:

```
NAME
       np_admin - manual page for np_admin Usage:
DESCRIPTION
       --help Usage instructions
       -a, --atsign (mandatory)    The atSign to use -r, --root-server
       Root server (e.g., root.atsign.org:64)
       ...entire option list repeated a second time...
       Could not find an option named "--version".
```

After:

```
NAME
       np_admin - NoPorts policy administration web app and API
SYNOPSIS
       np_admin [options]
DESCRIPTION
       NoPorts admin: serves the NoPorts policy administration web app and its REST API.
OPTIONS
       -a, --atsign (mandatory)
              The atSign to use
       -r, --root-server
              Root server (e.g., root.atsign.org:64) (defaults to "root.atsign.org:64")
       ...
EXAMPLES
       np_admin -a @alice_policy
SEE ALSO
       npp(1), npp_atserver(1)
```

### Testing

- `dart analyze`: no issues in `sshnoports` and `admin_api`; no new issues in `noports_core` (the 13 pre-existing infos/errors there are all in files this PR doesn't touch).
- `dart format`: clean on every touched file that was format-clean before (three files were already unformatted on trunk; only my insertions follow format style there to keep the diff reviewable).
- New unit tests for `formatCliHelp` pass.
- Compiled `sshnpd`, `srvd`, `npt`, `srv`, `npevents`, `npp_file`, `noports` and `np_admin` locally (macOS/arm64) and generated all their man pages with the exact CI command (`help2man -N --no-discard-stderr --include man/<bin>.h2m`): every page now has correct title, NAME, SYNOPSIS, DESCRIPTION and a properly itemised OPTIONS section, with no duplication and no stray error text.
- Verified `--help`/`--version` exit 0 on stdout, and that a bad flag still exits 1 with usage on stderr.
- Grepped the repo for consumers of the old `Version : ` output format — none found.

### Notes for reviewers

- Man page generation only runs on Linux release builds, so the CI on this PR won't exercise `help2man`; the local renders above used the same command and help2man version as CI.
- Behaviour change worth calling out: `--help` output moved from stderr to stdout for most binaries (this is the GNU convention and is required for help2man; error-path usage remains on stderr).


cconstab:
Pushed a follow-up commit (a7c159db0) after testing the generated pages on a Linux machine: sshnpd's OPTIONS section was still partly mangled. Its usage embeds option *group headings* from package:config ("Runtime Options", "atSign Options", ...), and help2man absorbs the options that follow a heading into a run-together text block when they share the heading's paragraph.

`formatCliHelp` now emits any column-zero line not starting with `-` as its own paragraph (flush-left, trailing colon), so the groups render as sub-headings inside OPTIONS and every option gets its own entry. Covered by a new unit test; re-verified the rendered sshnpd.1 locally. Only sshnpd uses group headings today, so the other pages are unchanged.

cconstab:
Tested by hand on Linux machine and the results now look good..

```
NP_ADMIN(1)                                                                                User Commands                                                                               NP_ADMIN(1)

NAME
       np_admin - NoPorts policy administration web app and API

SYNOPSIS
       np_admin [options]

DESCRIPTION
       NoPorts admin: serves the NoPorts policy administration web app and its REST API.

OPTIONS
       --help Usage instructions

       -a, --atsign (mandatory)
              The atSign to use

       -r, --root-server
              Root server (e.g., root.atsign.org:64) (defaults to "root.atsign.org:64")

       -v, --verbose
              More logging

       -d, --legacy-root-domain
              Deprecated abbreviation option for --root-server

       -b, --bind-ip
              Bind to something other than the default (defaults to "127.0.0.1")

       -p, --bind-port
              Bind to something other than the default port (defaults to "3000")

       --version
              Print version

EXAMPLES
       np_admin -a @alice_policy

REPORTING BUGS
       Report bugs at <https://github.com/atsign-foundation/noports/issues>

SEE ALSO
       npp(1), npp_atserver(1)

       Full documentation at <https://docs.noports.com/>

np_admin 6.12.1                                                                             August 2026                                                                                NP_ADMIN(1)
```

cconstab:
Resolved the np_admin versioning discussion point (commit pushed): admin_api now uses the same `build_version` setup as the sshnoports package, with its pubspec version set to the release version. `np_admin --version` reports `np_admin 5.15.1 (core 6.12.1)`, consistent with the other binaries, and the man page footer/header follow.

**Release process note:** `apps/admin/admin_api/pubspec.yaml` now needs its version bumped alongside `packages/dart/sshnoports/pubspec.yaml` and `bundles/universal.sh` when cutting a release.

cconstab:
caught an issue in testimg by hand that np_admin did not have a version number, so looked odd in man pages

cconstab:
To test yourself on a Linux machine

## Prerequisites

```bash
sudo apt-get update && sudo apt-get install -y help2man
# Dart SDK if not already installed:
sudo snap install dart    # or use the apt repo from dart.dev
```

## 1. Get the branch

```bash
git clone https://github.com/atsign-foundation/noports.git
cd noports
git checkout 2650-fix-man-page-generation
```

(If already cloned: `git checkout 2650-fix-man-page-generation && git pull`)

## 2. Compile the core binaries

```bash
cd packages/dart/sshnoports
dart pub get --enforce-lockfile
dart run build_runner build --delete-conflicting-outputs

mkdir -p sshnp
for BINARY in at_activate noports npevents npp npp_atserver npp_file npt srv sshnp sshnpd srvd; do
  dart compile exe bin/${BINARY}.dart -o sshnp/${BINARY}
done
```

> Note: the loop lists the binaries literally rather than via a `$BINS`
> variable so it works in zsh as well as bash (zsh does not word-split
> unquoted variables).

## 3. Compile np_admin

```bash
cd ../../../apps/admin/admin_api
dart pub get --enforce-lockfile
dart run build_runner build --delete-conflicting-outputs   # generates lib/src/version.dart
dart compile exe bin/np_admin.dart -o ../../../packages/dart/sshnoports/sshnp/np_admin
cd ../../../packages/dart/sshnoports
```

## 4. Generate the man pages (exact CI command)

```bash
for BINARY in at_activate noports npevents npp npp_atserver npp_file npt srv sshnp sshnpd srvd np_admin; do
  help2man -N --no-discard-stderr --include man/${BINARY}.h2m \
    sshnp/${BINARY} -o sshnp/${BINARY}.1
done
```

## 5. Review the pages

```bash
man ./sshnp/np_admin.1    # the page from the original bug report
man ./sshnp/sshnpd.1      # has option group headings (fixed in a7c159db0)
man ./sshnp/noports.1     # previously full of raw ANSI escape codes
man ./sshnp/npp_file.1    # --help previously crashed outright
```

(Use `man -l` instead of `man` on older man-db versions.)

### What to look for

- [ ] Page header/footer shows the tool's name, e.g. `SSHNPD(1)` — **not** `VERSION(1)`
- [ ] NAME has a real one-line description, e.g. `np_admin - NoPorts policy administration web app and API`
- [ ] SYNOPSIS and DESCRIPTION sections present
- [ ] Every option is its own OPTIONS entry — no run-together text like
      `The atSign to use -r, --root-server ...`
- [ ] sshnpd: group headings (`Runtime Options:`, `atSign Options:` ...) are
      sub-headings, and `-k`, `-a`, `-m`, `-p`, `-d`, `-s` each have their own entry
- [ ] No duplicated option list; no `Could not find an option named "--version".`
- [ ] EXAMPLES / SEE ALSO / REPORTING BUGS sections present
- [ ] Footer version reads `<name> 5.15.2 (core 6.12.2)`

## 6. Spot-check CLI behaviour

```bash
sshnp/sshnpd --version           # sshnpd 5.15.2 (core 6.12.2), stdout, exit 0
sshnp/np_admin --version         # np_admin 5.15.2 (core 6.12.2)
sshnp/np_admin --help; echo $?   # formatted Usage/Options on stdout, exit 0
sshnp/npp_file --help            # no crash / stack trace
sshnp/noports --help | cat -v    # no ASCII banner, no ANSI escape codes
sshnp/srvd --bogus; echo $?      # usage on stderr, exit 1
```

## Optional: install the pages properly

```bash
sudo cp sshnp/*.1 /usr/local/share/man/man1/
sudo mandb
man sshnpd
```

## Notes

- `at_activate`'s output is unchanged by the PR (it comes from
  `at_onboarding_cli`; its page was already fine — the `.h2m` just adds a
  NAME line).
- Man pages are only generated in Linux release builds, so CI on the PR
  does not exercise `help2man` — this local run is the real verification.

apps/admin/admin_api/pubspec.yaml
```diff
@@ -1,6 +1,7 @@
 name: admin_api
-description: A sample command-line application.
-version: 1.0.1
+description: NoPorts admin API server
+# Keep in step with packages/dart/sshnoports/pubspec.yaml at release time
+version: 5.15.2
```
JeremyTubongbanua: I disagree with this change. Then we'll have to remember to bump policy version each time we bump sshnoports

JeremyTubongbanua (changes requested):
I think np_admin should remain having its own separate version

apps/admin/admin_api/pubspec.yaml
cconstab: No problem with that but it is released as a versioned package and currently has no version. Suggestions welcome...

JeremyTubongbanua: I currently see 1.0.1 in trunk , so should be 1.1.0 in this PR

JeremyTubongbanua (changes requested):
Also @cconstab
- sshnoports needs a version minor bump as new commands were introduced
- sshnoports CHANGELOG update
- noports_core CHANGELOG update

cconstab:
> I think np_admin should remain having its own separate version

It has nver had a version number so where do you suggest we start?

Also we version the package not bins hence my logic of alignment. 

Could you see what others suggest?

JeremyTubongbanua:
> > I think np_admin should remain having its own separate version
>
> It has nver had a version number so where do you suggest we start?

I am confused because I see `version: 1.0.1` in pubspec.yaml in trunk.. Sorry if I'm on the wrong page. What do you mean?

cconstab:
> > > I think np_admin should remain having its own separate version
> > 
> > 
> > It has nver had a version number so where do you suggest we start?
> 
> I am confused because I see `version: 1.0.1` in pubspec.yaml in trunk.. Sorry if I'm on the wrong page. What do you mean?

Gotcha ! Thanks

apps/admin/admin_api/pubspec.yaml
cconstab: Done in d386b180c — admin_api is now versioned independently at 1.1.0 (minor bump for the new --help/--version handling), and the lockstep comment is gone. np_admin reports `np_admin 1.1.0 (core 6.13.0)`; the `(core ...)` suffix still ties it to the release train.

cconstab:
@JeremyTubongbanua all three review points addressed in d386b180c:

- **np_admin has its own version again**: 1.0.1 → 1.1.0, no lockstep with the release version. It reports `np_admin 1.1.0 (core 6.13.0)`.
- **sshnoports minor bump**: 5.15.2 → 5.16.0, with a CHANGELOG entry covering the man page fixes, the new `--version` flags, the `npp_file --help` crash fix, the `noports` ANSI/banner fix, and the stdout convention change.
- **noports_core CHANGELOG updated** — and since this PR adds public API there (`binaryName`/`formatCliHelp` in utils, `--version` on `NPAParams.parser`), core also got a minor bump 6.12.2 → 6.13.0 by the same logic as the sshnoports bump; the entries sit under the new heading since v5.15.2/6.12.2 is already cut as a pre-release. Shout if you'd rather core stayed at 6.12.2.

All three generated `version.dart` files and both lockfiles regenerated; analyze clean, tests pass, and compiled binaries verified to report the new versions.

JeremyTubongbanua (approved):

PR #2859, opened 2026-08-26
fix(sshnpd): reject non-2048-bit client ephemeral PK before RSA encrypt

cconstab:
## Summary

Fixes a heap buffer overflow in the C daemon's RVD session-key setup (`setup_rvd_session_encryption` in `packages/c/sshnpd/src/handler_commons.c`).

The path validated only the **client-supplied label** `clientEphemeralPKType == "rsa2048"` (`strncmp(pk_type, "rsa2048", 7)`), never the **actual modulus** of `clientEphemeralPK`. `atchops_rsa_encrypt()` takes no output-buffer-size argument and writes exactly the modulus length into the fixed `malloc(BYTES(256))` buffers used for the encrypted session AES key and IV. An **authenticated** client supplying a key with a larger real modulus (e.g. RSA-4096 → 512 bytes) labeled `"rsa2048"` overflows those 256-byte heap allocations with attacker-influenced content — a trust-boundary crossing (tunnel client → memory corruption / potential RCE on the daemon host).

## Fix

Enforce a true 2048-bit modulus (`ac.n.len == 256`) immediately after `atchops_rsa_key_populate_public_key` and before either `atchops_rsa_encrypt` call. A single guard covers both call sites (AES key + IV) since both use the same populated key. A mislabeled oversized key is now rejected cleanly through the existing error path (buffers freed, non-zero return).

## Severity

High (authenticated). `atchops_rsa_key_populate_public_key` imposes no modulus restriction, so the only barrier was the label string.

## Related

- Root-cause API issue upstream: atsign-foundation/at_c#701 — `atchops_rsa_encrypt` has no ciphertext buffer-size parameter, so **every** fixed-buffer caller shares this hazard. This PR is the correct call-site defense regardless of the upstream API change, and remains valid even after at_c is hardened (input validation belongs at the trust boundary).
- The same vulnerability + fix in the ESP32/Arduino port: atsign-foundation/at_Arduino_NoPorts#10.

### DIFF
--- a/packages/c/sshnpd/src/handler_commons.c
+++ b/packages/c/sshnpd/src/handler_commons.c
@@ -375,18 +375,32 @@ int setup_rvd_session_encryption(cJSON *payload, unsigned char **session_aes_key
       }
 
       // The "rsa2048" type above is client-supplied and only names the type;
-      // it does not constrain the actual modulus. atchops_rsa_encrypt writes
-      // exactly the modulus length into the fixed BYTES(256) buffers below, so
-      // a key whose real modulus is larger than 2048 bits (e.g. RSA-4096 ->
-      // 512 bytes) would overflow those heap allocations. Enforce a true
-      // 2048-bit modulus before any encryption is attempted.
-      if (ac.n.len != 256) {
-        atlogger_log(LOGGER_TAG, ATLOGGER_LOGGING_LEVEL_ERROR,
-                     "client ephemeral pk modulus is not RSA-2048 (n.len=%zu)\n", ac.n.len);
-        atchops_rsa_key_public_key_free(&ac);
-        free(*session_aes_key);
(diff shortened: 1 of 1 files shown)


cconstab:
Updated with the fix from atsign-foundation/at_Arduino_NoPorts#11: the original strict `n.len != 256` check rejected every legitimate client, because a real RSA-2048 modulus always has its MSB set and DER INTEGER encoding therefore prepends a 0x00 pad byte (parsed `n.len` == 257). The check now compares the count of significant modulus bytes (leading zeros stripped), which is also exactly how mbedtls sizes the ciphertext that the `BYTES(256)` buffers must hold — so the overflow guard's security intent is preserved while standard clients pass.

cpswan (approved):

PR #2860, opened 2026-08-26
feat(c): Add twinned session key (twinKeys) support to C sshnpd and srv

cconstab:
## What

Brings the C daemon up to parity with the Dart sshnpd for encrypted rvd traffic: when the client requests `twinKeys`, separate AES-256-CTR keys are used for each direction of the TCP channel — **C2D** (client→daemon) for decryption and **D2C** (daemon→client) for encryption.

Reference implementations used: the Dart `sshnpd_impl.dart` / `srv_impl.dart` (canonical protocol) and the at_Arduino_NoPorts relay (`_create_enc_dec` direction semantics, which this follows; its daemon-side deviations — always-twinning regardless of the request flag, and legacy response field names in twin mode — were deliberately **not** copied).

### sshnpd
- Advertise `"twinKeys": true` in the ping response `supportedFeatures`
- `handle_npt_request` / `handle_ssh_request` honour the request's `twinKeys` flag and generate a second session key + IV for the D2C direction (reusing `setup_rvd_session_encryption`, called once per direction)
- Response payload uses `aesKeyC2D`/`ivC2D`/`aesKeyD2C`/`ivD2C` field names when twinned (matching the Dart daemon), legacy `sessionAESKey`/`sessionIV` otherwise

### srv
- `create_encrypter_and_decrypter` now takes both key pairs; the decrypter always uses the C2D key, the encrypter uses the D2C key when provided, falling back to C2D (legacy single-key mode)
- Control channel accepts both `connect:key:iv` and `connect:keyC2D:ivC2D:keyD2C:ivD2C` per-socket messages
- Standalone srv binary reads `RV_AES_C2D`/`RV_IV_C2D` (what current Dart `Srv.exec` sets — it previously only read the legacy `RV_AES`/`RV_IV`) with fallback, plus optional `RV_AES_D2C`/`RV_IV_D2C`
- Variable naming is direction-explicit and symmetric (`*_c2d_*` / `*_d2c_*`), mirroring the Dart `aesC2D`/`aesD2C` convention

### Drive-by fixes
- `connect:no:encrypt` detection compared a string literal instead of the variable (always false)
- Repaired the srv test build, broken for some time: stale `srv/stream.h` include, wrong `srv_lib` target name, renamed atlogger API, unterminated buffers in `test_stream_encrypt`

### Tests
- New `test_twin_keys`: single-key round trip, both twinned directions against simulated client-side transformers, and asserts D2C traffic is **not** decryptable with the C2D key
- Note: CI (`c_unit_tests.yaml`) currently only runs the sshnpd tests; the srv tests run with `-DSRV_BUILD_TESTS=ON` — worth adding to the workflow as a follow-up

## Backwards compatibility

Both directions hold: clients only send `twinKeys: true` after seeing the feature flag in the ping response; a legacy client that omits it gets the old single-key behaviour and field names unchanged. The Dart client accepts either naming (`sessionAESKey ?? aesKeyC2D`).

## Interaction with #2859

Verified no conflict (`git merge-tree` clean). Complementary: the modulus-length guard in #2859 lives inside `setup_rvd_session_encryption`, which this PR calls once per direction — so the guard automatically covers the new D2C bundle, and both call sites here already propagate its failure.

## Verification

- `dev` and `strict` cmake presets build clean
- sshnpd unit tests 5/5 (run exactly as CI does), srv tests 2/2

### DIFF
--- a/packages/c/justfile
+++ b/packages/c/justfile
@@ -59,16 +59,18 @@ strict-docker +ARGS='':
 
 memcheck +ARGS='': build-dev
   #!/usr/bin/env bash
-  file="valgrind-sshnpd-$(date +%s).log"
-  # child-silient-after-fork=yes: ignore session forks, we don't care
-  valgrind --log-file="logs/$file" \
+  mkdir -p logs
+  # Session forks (the srv processes) run most of the session encryption
+  # code, so trace them too: each process writes its own logs/vg-<pid>.log.
+  # After a run: grep -A4 "LEAK SUMMARY" logs/vg-*.log
+  valgrind --log-file="logs/vg-%p.log" \
   --tool=memcheck \
   --leak-check=full \
   --show-leak-kinds=all \
-  --child-silent-after-fork=yes \
+  --suppressions=valgrind.supp \
   {{dev_dir}}/sshnpd/sshnpd -a $TO_ATSIGN -m $FROM_ATSIGN -d $DEVICE_NAME \
   --permit-open "*:*" -v {{ARGS}}
-  echo "memcheck log saved to: $file"
+  echo "memcheck logs saved to: logs/vg-<pid>.log (one per process)"
 
 # CONTAINER COMMANDS
 
--- a/packages/c/srv/include/srv/params.h
+++ b/packages/c/srv/include/srv/params.h
@@ -9,8 +9,10 @@
 
 typedef struct {
   char *rvd_auth_string;
-  char *session_aes_key_string;
-  char *session_aes_iv_string;
+  char *session_aes_key_c2d_string;
+  char *session_aes_iv_c2d_string;
+  char *session_aes_key_d2c_string;
+  char *session_aes_iv_d2c_string;
 } srv_env_t;
 
 /**
@@ -31,8 +33,15 @@ typedef struct {
   int timeout;
 
   char *rvd_auth_string;
-  char *session_aes_key_string;
-  char *session_aes_iv_string;
+
+  // Session encryption keys (base64). When twinned keys are in use, the C2D
+  // (client to daemon) key decrypts inbound traffic and the D2C (daemon to
+  // client) key encrypts outbound traffic. When the D2C strings are NULL the
+  // C2D key is used in both directions (legacy single-key sessions).
+  char *session_aes_key_c2d_string;
+  char *session_aes_iv_c2d_string;
+  char *session_aes_key_d2c_string;
+  char *session_aes_iv_d2c_string;
 } srv_params_t;
 
 /**
--- a/packages/c/srv/include/srv/srv.h
+++ b/packages/c/srv/include/srv/srv.h
@@ -155,6 +155,33 @@ int server_to_socket(const srv_params_t *params, const char *auth_string, chunke
 int aes_ctr_crypt_stream(const chunked_transformer_t *self, size_t len, const unsigned char *input,
                          unsigned char *output);
 
-int create_encrypter_and_decrypter(const char *session_aes_key_string, const char *session_aes_iv_string,
+/**
+ * @brief create a single AES-CTR stream transformer from a base64 key and iv
+ *
+ * @param aes_key_base64 the base64 encoded AES-256 key
+ * @param aes_iv_base64 the base64 encoded 16 byte iv
+ * @param transformer the transformer to initialize (free with mbedtls_aes_free(&transformer->aes_ctr.ctx))
+ * @return int 0 on success, non-zero on error
+ */
+int create_transformer(const char *aes_key_base64, const char *aes_iv_base64, chunked_transformer_t *transformer);
+
+/**
+ * @brief create the encrypter and decrypter for one bridged connection
+ *
+ * The decrypter always uses the C2D (client to daemon) key, since it decrypts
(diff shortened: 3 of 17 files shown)


cpswan (dismissed):

JeremyTubongbanua (approved):

JeremyTubongbanua (changes requested):
Hey @cconstab , some leaks Claude caught:

```
`handle_ssh_request.c` has two early-return error paths (C2D setup failure at line 83 and D2C setup failure at line 94) that both return without freeing envelope or rvd_auth_string, unlike handle_npt_request.c which correctly cleans up both in its equivalent paths. The C2D leak is pre-existing, but the D2C leak is new code in this PR that copied the broken pattern instead of the correct one from the npt handler. Add `cJSON_Delete(envelope)` and the conditional free(rvd_auth_string) to both error paths in handle_ssh_request.c, matching what handle_npt_request.c already does. Everything else in the PR looks correct.

Here's the side-by-side confirmation:

handle_ssh_request.c -- C2D failure (line 83-86):
if (res != 0) {
    atlogger_log(LOGGER_TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "Failed to setup rvd session encryption");
    return;  // leaks: envelope, rvd_auth_string
}

handle_ssh_request.c -- D2C failure (line 94-101):
if (res != 0) {
    atlogger_log(LOGGER_TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "Failed to setup rvd session d2c encryption");
    free(session_aes_key_c2d);
    free(session_iv_c2d);
    free(session_aes_key_c2d_base64);
    free(session_iv_c2d_base64);
    return;  // leaks: envelope, rvd_auth_string
}

handle_npt_request.c -- C2D failure (line 102-109) -- correct:
if (res != 0) {
    atlogger_log(LOGGER_TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "Failed to setup rvd session encryption\n");
    cJSON_Delete(envelope);           // <-- present
    if (authenticate_to_rvd) {
        free(rvd_auth_string);        // <-- present
    }
    return;
}

handle_npt_request.c -- D2C failure (line 117-128) -- correct:
if (res != 0) {
    ...
    cJSON_Delete(envelope);           // <-- present
    if (authenticate_to_rvd) {
        free(rvd_auth_string);        // <-- present
    }
    return;
}

Both SSH handler error paths are missing cJSON_Delete(envelope) and the conditional free(rvd_auth_string). Both also omit the trailing \n in the log format string (minor, but the npt handler has it). The C2D leak is pre-existing; the D2C leak is new code that copied the broken pattern. Everything else checks out.
```

cconstab:
Good catch — fixed in the latest commit. Both error paths in `handle_ssh_request.c` now free `envelope` and (conditionally) `rvd_auth_string`, matching `handle_npt_request.c`, and the missing trailing newlines on those log lines are added. This also fixes the pre-existing C2D-path leak, not just the new D2C one. Builds clean on dev+strict presets, 5/5 sshnpd and 2/2 srv tests pass.

cconstab:
Follow-up to the review: did a full sweep of every allocation/error path in the files this PR touches, plus a cppcheck pass matching the CI workflow (`f5531e897`). Found and fixed, beyond the two paths flagged:

**handler_commons.c**
- `verify_envelope_signature_from`: three error paths leaked `buffer`, the RSA public-key context, and/or the atkey
- `setup_rvd_session_encryption`: unknown ephemeral PK type leaked the freshly generated session key + IV and left dangling out-pointers
- `send_success_payload`: `keyname` leaked when shared-key creation failed (the `goto` skipped the block that frees it), and — more significant — failures from atkey creation, `atkey_to_string`, and `atclient_notify` were logged but **not propagated**, so the caller believed the success response was delivered when it wasn't

**srv.c**
- both daemon-side runners proceeded with **uninitialized AES contexts** when transformer creation failed (UB); they now return the error
- a malformed `connect` control message now skips that request instead of running garbage crypto; `connect:no:encrypt` previously passed uninitialized transformers (UB) — now passes NULL with `socket_to_socket` made NULL-tolerant
- classic realloc leak in `process_multiple_requests` (cppcheck)

**side.c**
- unchecked `malloc` → memset on NULL under OOM (cppcheck)

cppcheck is now clean for srv + sshnpd; dev/strict builds pass, 5/5 sshnpd and 2/2 srv tests pass.

JeremyTubongbanua (dismissed):

cconstab:
Valgrind against live sessions on Linux caught one more — a per-connection leak that no static sweep would find, because it isn't a missing `free`: `socket_to_socket` tears down the second side thread with `pthread_cancel`, which lands inside `mbedtls_net_recv` (a cancellation point), so the thread dies holding its live 81-byte buffer (`BUFFER_LEN`) every time a connection closes. The same teardown also leaked the cancelled thread's **socket fd** — in a long-lived multi-mode srv that's one fd per tunnelled connection until ulimit.

Fixed in `side.c`/`srv.c`: side threads now use stack buffers (nothing to leak on cancellation), the cancelled thread is joined and both side sockets closed idempotently, and the send-retry loop got a real fix too — it decremented the remaining length on partial send but never advanced the pointer, i.e. latent stream corruption.

Builds clean, cppcheck clean, 5/5 + 2/2 tests pass.

JeremyTubongbanua (approved):

PR #2863, opened 2026-08-26
fix: sshnpd C daemon manager list handling and exact -m authorization match

cconstab:
## Problem

A user reported that `sshnpd -m` "works but seems to have a maximum limit of atSigns, then breaks."

Investigation showed the `-m` parser itself has no limit (verified under ASan/UBSan up to 1000 comma-separated atSigns). The real cause is in the C daemon's startup: `handle_username_keys` loops over all manager atSigns sharing the `username.<device>.sshnp` record, but returns the status of only the **last** iteration. `main()` treated any nonzero return as fatal. So:

- a bad atSign (unactivated / typo / unreachable) at the **end** of the `-m` list killed the daemon at startup, while
- the same bad atSign in the **middle** of the list was silently swallowed.

Appending one more atSign to a working list is exactly how you hit the first case, which presents as a "maximum limit."

## Fix

1. **`handle_username_keys` is now best effort, matching the Dart daemon** (`_shareUsername` / `_refreshDeviceEntry` wrap each share in try/catch): per-atSign failures are logged and skipped, and can never prevent daemon startup. `main()` no longer treats the call as fatal.

2. **`is_manager_atsign` now guarantees an exact canonical match.** The `-m` list is canonicalized at startup with `sshnpd_normalize_atsign` (same rules as `AtUtils.fixAtSign`: lowercase, dots stripped, single leading `@`, reserved/whitespace/control characters rejected). The requesting atSign from the notification is now passed through the same canonicalizer before comparison, so:
   - every valid spelling of an authorized atSign (`@Alice`, `@a.lice`) matches, per atProtocol identity rules;
   - near-misses (`@alice1`, `@malice`, `@alic`, ...) never match;
   - anything that is not a valid atSign fails closed.

## Security review

- The authorization gate in `daemon.c` runs before any handler dispatch and uses the server-authenticated `notification->from`; handlers never take the requester identity from the payload.
- Fail-closed behavior preserved: NULL/empty sender, empty/NULL manager list, and policy-mode all still deny.
- Canonicalization is identity-preserving per protocol (parity with `AtUtils.fixAtSign`), bounds-checked into a fixed buffer, and rejects over-length input.
- Envelope signature verification (`publickey@<from>`) is untouched.
- The best-effort change affects only outbound sharing of the username record, not authorization.

## Tests

- New unit tests in `test_manager_auth.c`: canonical equivalence, near-miss denial, malformed/over-length senders fail closed.
- Full C test suite passes (5/5 via ctest).
- Parser separately fuzzed with 1–1000 atSigns under AddressSanitizer/UBSan: no errors.

## Note (not in this PR)

`device_info_last_sent` is declared and NULL-checked but never allocated, so `send_next_device_info` has been dead code since fc9b307a8 — the C daemon never proactively refreshes `device_info` entries. Worth a separate issue.

### DIFF
--- a/packages/c/sshnpd/src/handler_commons.c
+++ b/packages/c/sshnpd/src/handler_commons.c
@@ -4,6 +4,7 @@
 #include "atchops/rsa.h"
 #include "atclient/notify.h"
 #include "atclient/notify_params.h"
+#include "sshnpd/authorization.h"
 #include "sshnpd/params.h"
 #include "sshnpd/sshnpd.h"
 #include <atchops/constants.h>
@@ -30,11 +31,19 @@ bool is_manager_atsign(const sshnpd_params *params, const char *atsign) {
     return false;
   }
 
+  // The manager list was canonicalized at startup; canonicalize the requesting
+  // atSign the same way so the comparison is always exact (case, dots, '@'
+  // prefix). Anything that isn't a valid atSign fails closed.
+  char normalized[SSHNPD_ATSIGN_BUFFER_LEN];
+  if (sshnpd_normalize_atsign(atsign, normalized, sizeof(normalized)) != 0) {
+    return false;
+  }
+
   for (size_t i = 0; i < params->manager_list_len; i++) {
     if (params->manager_list[i] == NULL) {
       continue;
     }
-    if (strcmp(atsign, params->manager_list[i]) == 0) {
+    if (strcmp(normalized, params->manager_list[i]) == 0) {
       return true;
     }
   }
--- a/packages/c/sshnpd/include/sshnpd/device_info.h
+++ b/packages/c/sshnpd/include/sshnpd/device_info.h
@@ -19,6 +19,8 @@ struct sshnpd_device_info_state {
   uint8_t attempts;
 };
 
+// Shares (or un-shares) the username with each manager atSign. Best effort:
+// per-atSign failures are logged and skipped; always returns 0.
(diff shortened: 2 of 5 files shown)


packages/c/sshnpd/src/handler_commons.c
```diff
@@ -30,11 +31,19 @@ bool is_manager_atsign(const sshnpd_params *params, const char *atsign) {
     return false;
   }
 
+  // The manager list was canonicalized at startup; canonicalize the requesting
+  // atSign the same way so the comparison is always exact (case, dots, '@'
+  // prefix). Anything that isn't a valid atSign fails closed.
+  char normalized[SSHNPD_ATSIGN_BUFFER_LEN];
+  if (sshnpd_normalize_atsign(atsign, normalized, sizeof(normalized)) != 0) {
```
JeremyTubongbanua, line 38: While `sizeof(normalized)` works perfectly fine, a convention we've adopted in the SDK is to use `sizeof(char) * SSHNPD_ATSIGN_BUFFER_LEN`. In the event that normalized changes from a static-allocated array to a dynamic allocated array (char *ptr), then the size would be incorrect, and we would get very bad errors.

JeremyTubongbanua (approved):
Just one nit, but it's fine, there are likely other instances in csshnpd where the same feedback applies which I think should be done in a later PR

cconstab:
> Just one nit, but it's fine, there are likely other instances in csshnpd where the same feedback applies which I think should be done in a later PR

I like the nit! I will fix in another PR as you suggest

PR #2864, opened 2026-08-26
feat: adjustableTimeout support in csshnpd

### LINKED ISSUE #2831: NoPorts Desktop can't connect to an OpenWRT csshnpd daemon

cconstab:
> [!NOTE]
> **Ready for review.** First of the stacked chain #2864 → #2865 → #2867 (merge in that order; GitHub retargets each PR as its base merges). CI is green. End-to-end verification of the whole feature set is being run on the `cconstab/c-2831-integration` branch, which merges this stack plus #2870.

## What this does

Fixes #2831 (option 3, as agreed in the issue discussion): csshnpd now honors the `timeout` value from npt session requests and advertises the `adjustableTimeout` daemon feature, so clients that send a non-default timeout (e.g. NoPorts Desktop) pass the pre-flight feature check and can connect.

## Key insight

The issue discussion estimated 1–2 days because it assumed "supporting timeout" meant per-byte idle tracking across the srv relay threads. Checking the Dart reference implementation shows the actual semantics are much simpler: `SocketConnector`'s timeout is a **no-connections grace timer** — close the srv if no session is established within the timeout, and close after the last session ends. There is no idle-byte enforcement anywhere in the Dart implementation either.

## Changes

- **`packages/c/srv/src/srv.c`** — the real work. Multi-mode control channel now polls with `mbedtls_net_recv_timeout` (1s interval) instead of blocking forever in `mbedtls_net_recv`, tracks active socket-to-socket sessions with a mutex-protected counter, and exits once there have been no active sessions for `params->timeout` seconds. Slightly gentler than Dart (which closes the instant the last session ends) to avoid killing the srv between a client's back-to-back reconnects, but still an honest implementation of the feature.
- **`packages/c/srv/src/params.c` / `params.h`** — `timeout` now has a default (30s, matching `DefaultArgs.srvTimeoutInSeconds`); previously it was uninitialized stack memory in the sshnpd path (harmless only because nothing read it).
- **`packages/c/sshnpd/src/handle_npt_request.c`** — parses `timeout` (milliseconds) from the npt payload, rounds up to whole seconds, defaults to 30s when absent.
- **`packages/c/sshnpd/src/run_srv_process.c` / `.h`, `handle_ssh_request.c`** — thread the timeout through; the ssh flow passes the default since ssh_request payloads don't carry a timeout.
- **`packages/c/sshnpd/src/main.c`** — advertise `adjustableTimeout: true` in the ping response.

## Side benefit

This also fixes a real resource leak: every abandoned or completed npt session previously left a forked srv process blocked forever on the control-channel recv (the process only died if the relay closed the socket). On long-running OpenWRT devices those accumulate. Now they exit on their own.

## Testing so far

- `cmake --build` clean, all 5 existing unit tests pass.
- Smoke test: standalone `srv --multi --timeout 3` against a local listener connects its control channel and exits cleanly with `No connections for 3 seconds - closing srv` after exactly 3.0s (previously: blocked forever).

- [ ] End-to-end test: NoPorts Desktop / `npt` client with a non-default timeout against this daemon via a real relay
- [ ] Verify sessions survive longer than the timeout while traffic is flowing, and that back-to-back reconnects within the timeout work
- [ ] Test on an OpenWRT target
- [ ] Consider whether `handle_npt_request` should cap absurdly large timeout values

### DIFF
--- a/packages/c/sshnpd/src/handle_npt_request.c
+++ b/packages/c/sshnpd/src/handle_npt_request.c
@@ -13,6 +13,7 @@
 #include <sshnpd/handle_ssh_request.h>
 #include <sshnpd/handler_commons.h>
 #include <sshnpd/run_srv_process.h>
+#include <srv/params.h>
 #include <stdlib.h>
 #include <string.h>
 #include <sys/wait.h>
@@ -161,9 +162,20 @@ void handle_npt_request(atclient *atclient, sshnpd_params *params, bool *is_chil
 
     const bool multi = true;
 
+    // The npt session request carries the timeout in milliseconds; the srv
+    // works in whole seconds, so round up (DaemonFeature.adjustableTimeout)
+    int timeout_seconds = SRV_DEFAULT_TIMEOUT_SECONDS;
+    cJSON *timeout_json = cJSON_GetObjectItem(payload, "timeout");
+    if (cJSON_IsNumber(timeout_json) && cJSON_GetNumberValue(timeout_json) > 0) {
+      timeout_seconds = (int)((cJSON_GetNumberValue(timeout_json) + 999) / 1000);
+      if (timeout_seconds < 1) {
+        timeout_seconds = 1;
+      }
+    }
+
     run_srv_process(rvd_host_str, rvd_port_int, requested_host_str, requested_port_int, authenticate_to_rvd,
-                    rvd_auth_string, encrypt_rvd_traffic, multi, session_aes_key_c2d, session_iv_c2d, session_aes_key_d2c,
-                    session_iv_d2c);
+                    rvd_auth_string, encrypt_rvd_traffic, multi, timeout_seconds, session_aes_key_c2d, session_iv_c2d,
+                    session_aes_key_d2c, session_iv_d2c);
 
     *is_child_process = true;
 
--- a/packages/c/sshnpd/src/handle_npt_request.c
+++ b/packages/c/sshnpd/src/handle_npt_request.c
@@ -167,7 +167,14 @@ void handle_npt_request(atclient *atclient, sshnpd_params *params, bool *is_chil
     int timeout_seconds = SRV_DEFAULT_TIMEOUT_SECONDS;
     cJSON *timeout_json = cJSON_GetObjectItem(payload, "timeout");
     if (cJSON_IsNumber(timeout_json) && cJSON_GetNumberValue(timeout_json) > 0) {
-      timeout_seconds = (int)((cJSON_GetNumberValue(timeout_json) + 999) / 1000);
+      double timeout_ms = cJSON_GetNumberValue(timeout_json);
+      // Clamp before the cast: an out-of-range double -> int conversion is UB.
+      // The cap matches the npt client's "never" timeout (-T 0 -> 365 days),
+      // the largest value a well-behaved client sends.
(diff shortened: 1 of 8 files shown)


packages/c/sshnpd/src/handle_npt_request.c
```diff
@@ -161,9 +162,20 @@ void handle_npt_request(atclient *atclient, sshnpd_params *params, bool *is_chil
...
+    // works in whole seconds, so round up (DaemonFeature.adjustableTimeout)
+    int timeout_seconds = SRV_DEFAULT_TIMEOUT_SECONDS;
+    cJSON *timeout_json = cJSON_GetObjectItem(payload, "timeout");
+    if (cJSON_IsNumber(timeout_json) && cJSON_GetNumberValue(timeout_json) > 0) {
+      timeout_seconds = (int)((cJSON_GetNumberValue(timeout_json) + 999) / 1000);
+      if (timeout_seconds < 1) {
+        timeout_seconds = 1;
+      }
+    }
+
```
JeremyTubongbanua, line 182: A very large timeout value in the JSON payload could overflow the double-to-int cast.

Not *really* worth fixing, since this code would only execute on a payload that would be coming from an authenticated Atsign.

But a quick check on the size of the string or something would be good.

JeremyTubongbanua (dismissed):

packages/c/sshnpd/src/handle_npt_request.c
cconstab: Fixed in d784df6a6: the requested timeout is now clamped before the cast, so the double -> int conversion can't go out of range. The cap is `SRV_MAX_TIMEOUT_SECONDS` = 365 days — chosen to match the expected maximum from the npt client, which sends exactly 365 days for a "never" timeout (`-T 0` -> `neverTimeoutDays` in npt.dart). So an oversized value degrades to the same behavior a "never" request gets, rather than wrapping. Verified the arithmetic on the edges: 30000ms -> 30s, 100ms -> 1s, 1e15ms -> 31536000s (clamped), exactly-365d -> unchanged.

JeremyTubongbanua (approved):

Issue #1295, opened 2024-08-29
Better defaults for sshnpd's --permit-open


gkc:
### Is your feature request related to a problem? Please describe.

Currently permitOpen defaults to `localhost:22,localhost:3389` which is overly restrictive for a typical 'bastion'-like daemon deployment.

### Describe the solution you'd like

IF a --permit-open is not specified by the user
AND a policy atSign IS specified
THEN default permitOpen to `*:*`

### Describe alternatives you've considered

Considered just defaulting to '*:*' in all cases but that might be overly permissive. In discussion, @cpswan and I felt that this proposed solution has the best balance between considerations of user experience and security.

cconstab:
If you define a policy server then default open?

gkc:
> If you define a policy server then default open?

Yes exactly

JeremyTubongbanua:
arch comments

- proposal is good if we want to eliminate administrator hassles, because if they want to permit open a port, they would have to both 1. manually configure permit open on sshnpd and 2. do policy stuff. 
- story behind of being *restrictive* on sshnpd permit open is to encourage the person using sshnpd to explicitly say what is allowed
- bastion in corporate environment is setting firewall rules one by one, whch is still the behaviour that will be exercised when using policy manager

gkc:
PR in review

PR #2867, opened 2026-08-26
feat: policy service (--policy-manager) support in csshnpd

cconstab:
> [!NOTE]
> **Ready for review.** Stacked on #2865 → #2864; merge in order — this PR’s diff is clean against #2865’s branch. Full CI runs once the bases merge and this retargets to trunk. End-to-end verification of the whole feature set is being run on the `cconstab/c-2831-integration` branch.

## What this does

Part 3 of 3 for #2831: enables `--policy-manager` / `-p` in csshnpd and implements the NPA auth-check RPC protocol, matching the Dart daemon and the Arduino daemon (used as the C reference).

## Protocol

- Request: notification `@policy:request.<reqId>.auth_checks.__rpcs.sshnp@daemon`, value `{"reqId":<epoch µs>,"payload":{"daemonAtsign","daemonDeviceName","daemonDeviceGroupName","clientAtsign"}}`, encrypted, 30s expiry. `reqId` is epoch-microseconds like Dart's `AtRpc` (a boot counter would collide with the policy server's 30s request mutex across daemon restarts — a real bug in the Arduino implementation this port avoids).
- Response: `ack` resets the 10s clock; `success` carries `{authorized, message, permitOpen[]}`; `nack`/`error`/timeout/garbage ⇒ **fail closed**. Only the configured policy atSign may answer — the sender is checked before the reqId, since the key pattern alone binds a response to nobody.

## Semantics (follows the Dart daemon where it and Arduino differ)

- At least one of `--manager` / `--policy-manager` required; **managers short-circuit** the policy check (Dart behavior; Arduino ignores the manager list in policy mode).
- Policy `permitOpen` is enforced **in addition to** the daemon's own `--permit-open` (Dart ANDs them; Arduino replaces). With `-p` and no explicit `--permit-open`, the daemon's own list defaults to `*:*` (Dart #1295), so policy is the effective ACL out of the box.
- Matching grammar: `host:port`, `*` wildcards, split on last colon, malformed entries skipped, `authorized:true` + empty `permitOpen` denies everything.
- Device heartbeat `<device>.devices.policy` (the ping response JSON) sent to the policy service at startup and every 5 minutes.
- In policy mode the monitor subscribes unfiltered — RPC response keys don't contain the `<device>.sshnp@` substring the normal filter uses.

## Design note: synchronous wait

csshnpd's main loop is single-threaded, so while an auth check is in flight the daemon reads the monitor looking for the RPC response and **drops unrelated notifications** it encounters (logged), like the Arduino daemon's single in-flight slot. The policy decision gates every notification type (ping, sshpublickey, ssh_request, npt_request), same as Dart's pre-dispatch `authCheck`. Denials are a silent drop + log, consistent with csshnpd's existing unauthorized handling. Not ported (yet): Dart's 15s per-client decision cache, `--device-group` (sends `__none__`), and the `config` push subscription for event logging (csshnpd has no event logging).

## Testing so far

- New `test_policy` unit test: response parsing (success/ack/nack, reqId mismatch, garbage, missing fields) and `permitOpen` matching (wildcards, malformed entries, empty-list-denies, `*:*`). `leaks --atExit`: 0 leaks.
- `test_manager_auth` updated: managers now short-circuit when a policy manager is configured (the old test asserted the unimplemented-policy stub that denied everyone).
- All 7 unit tests pass; CLI sanity: `-p @policy` accepted, permit-open defaults to `*:*`, arg validation intact.

- [ ] End-to-end against a real policy service (`npp_file` with a test policy YAML): manager short-circuit, authorized/denied clients, permitOpen narrowing, timeout fail-closed
- [ ] Confirm the RPC response key form from a real Dart policy service (code matches on the `.auth_checks.__rpcs` infix so both the `.sshnp`-suffixed and unsuffixed forms match)
- [ ] Verify heartbeat appears in the policy service's device view
- [ ] Consider a short per-client decision cache (Dart uses 15s) if ping+npt bursts double-hit the policy service
- [ ] OpenWRT target build/test

### DIFF
--- a/packages/c/sshnpd/CMakeLists.txt
+++ b/packages/c/sshnpd/CMakeLists.txt
@@ -22,6 +22,7 @@ set(
   ${CMAKE_CURRENT_LIST_DIR}/src/main.c
   ${CMAKE_CURRENT_LIST_DIR}/src/params.c
   ${CMAKE_CURRENT_LIST_DIR}/src/permitopen.c
+  ${CMAKE_CURRENT_LIST_DIR}/src/policy.c
   ${CMAKE_CURRENT_LIST_DIR}/src/run_srv_process.c
 )
 
--- a/packages/c/sshnpd/include/sshnpd/handle_npt_request.h
+++ b/packages/c/sshnpd/include/sshnpd/handle_npt_request.h
@@ -1,8 +1,13 @@
 #ifndef HANDLE_NPT_REQUEST_H
 #define HANDLE_NPT_REQUEST_H
 #include "sshnpd/params.h"
+#include "sshnpd/policy.h"
 #include <atclient/monitor.h>
 
+// policy is the policy service's decision for this request, or NULL when the
+// requester was approved without a policy check (manager atsign, or no
+// policy service configured)
 void handle_npt_request(atclient *atclient, sshnpd_params *params, bool *is_child_process,
-                        atclient_monitor_message *message, atchops_rsa_key_private_key signing_key);
+                        atclient_monitor_message *message, atchops_rsa_key_private_key signing_key,
+                        const sshnpd_policy_decision *policy);
 #endif
--- a/packages/c/sshnpd/include/sshnpd/handle_ssh_request.h
+++ b/packages/c/sshnpd/include/sshnpd/handle_ssh_request.h
@@ -1,9 +1,14 @@
 #ifndef HANDLE_SSH_REQUEST_H
 #define HANDLE_SSH_REQUEST_H
 #include "sshnpd/params.h"
+#include "sshnpd/policy.h"
 #include <atclient/monitor.h>
 
+// policy is the policy service's decision for this request, or NULL when the
+// requester was approved without a policy check (manager atsign, or no
+// policy service configured)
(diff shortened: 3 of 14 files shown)


cconstab:
Added `3ad2239a8`: clients are now told **why** a request was denied, matching the Dart daemon — npt denials reply with `Connection to <host>:<port> denied based on daemon --permit-open [...]` / `... based on POLICY --permit-open [...]`, and ssh denials with `Daemon does not permit connections to localhost:<p>` / `Client is not permitted connections to localhost:<p>`. The ssh permit-open checks moved from the main loop into the handler where the sessionId (needed to address the reply) is available. Previously csshnpd silently dropped these and the client just timed out.

JeremyTubongbanua (dismissed):

cconstab:
@JeremyTubongbanua last one 🙏 — #2865 merged, so this PR is now retargeted to trunk (which auto-dismissed your approval, same GitHub behavior as before). The only change since your approval is the trunk merge-in (one help-text conflict in params.c resolved in trunk's favor). Once you re-approve and CI is green this completes the C daemon stack.

JeremyTubongbanua (approved):

PR #2869, opened 2026-08-26
feat: ephemeral tunnel key in csshnpd (PARKED - sshnp --no-et tunnel mode)

cconstab:
> [!IMPORTANT]
> **Parked — we suggest ignoring this need for now.** The ephemeral tunnel key only matters for `sshnp --no-encrypt-rvd-traffic` (the pre-E2EE legacy tunnel mode). On a modern stack there is no reason to disable socket E2EE, and the two real-world needs it appeared to serve are already covered without it:
> - `sshnp` with default settings (E2EE on) works against csshnpd today — clients only demand `ephemeralPrivateKey` when `--no-et` is passed
> - a non-standard sshd port is served by daemon-side `--local-sshd-port`, or by `npt -rp <port>` (which is what NoPorts Desktop does)
>
> Meanwhile this is the largest security/maintenance surface of the csshnpd parity PRs: per-session `authorized_keys` rewrites, an `ssh-keygen` exec (often absent on dropbear-only OpenWRT images), and private keys in responses. Not worth it for a legacy client mode with no known users on C daemons.
>
> **The implementation is complete and unit-tested, and stays here in case the need arises** — if a real `--no-et`-against-csshnpd use case turns up, this can be rebased and taken to end-to-end testing. It has been dropped from the `cconstab/c-2831-integration` test branch. Suggested documentation stance: *csshnpd supports sshnp with encrypted relay traffic (the default); `--no-et` tunnel mode is unsupported.*

---

## What this does

Implements the **ephemeral tunnel key** flow in `ssh_request` handling, which is what "supporting `sshnp --remote-sshd-port 2222`" actually requires on the daemon side.

## Why this would be the fix for --remote-sshd-port

`--remote-sshd-port` never travels to the daemon — the `SshnpSessionRequest` payload carries no sshd port at all. Instead, the Dart client (both openssh and dartPure impls) reaches the alternate port by making an ssh port-forward (`-L …:localhost:2222` / `startForwarding`) **inside an initial tunnel ssh session to the daemon's sshd**, authenticated with an ephemeral key that the daemon generates, authorizes, and returns as `ephemeralPrivateKey` in the session response. csshnpd never returned that key, so tunnel-mode clients fail with `Expected an ephemeral private key from device daemon, but it was not set`.

Two behavioral notes (true against Dart daemons too, this is client-side design):
- The tunnel session (and therefore `--remote-sshd-port`) is only used when `encryptRvdTraffic == false` (`--no-et`). With e2ee on (the default), the user's ssh lands directly on the **daemon's** `--local-sshd-port` and the client's `--remote-sshd-port` is a no-op.
- The daemon's own srv always bridges to `--local-sshd-port`; csshnpd already supports that flag.

## Changes (ports Dart `startDirectSsh` exactly)

- **`ephemeral_key.c` (new)**:
  - keypair generation via the same ssh-keygen invocations as Dart's `LocalSshKeyUtil` (`-t ed25519 -a 100` default; `-t rsa -b 4096` with `--ssh-algorithm ssh-rsa`), via fork/exec (no shell), session id validated to uuid characters first; transient key files unlinked immediately after reading
  - authorized_keys entry byte-matching Dart: `command="echo \"ssh session complete\";sleep 20",PermitOpen="localhost:<port>"[,extra] <pubkey> sshnp_ephemeral_<sessionId>` — reuses the existing `authorize_ssh_public_key`
  - removal 15 seconds after the response (same grace as Dart's `Timer`), swept from the daemon main loop; queue-full evicts the oldest key immediately
- **`handle_ssh_request.c`** — generate + authorize + schedule removal, pass the private key into the response; non-fatal on failure
- **`handler_commons.c`** — `send_success_payload` gains the `ephemeralPrivateKey` field (NULL/omitted for npt)
- **`params.c`** — `--ephemeral-permission` was parsed but dropped; now wired through, matching the Dart daemon's `--ephemeral-permissions`

## Testing done

- New `test_ephemeral_key`: generation round-trip (skipped cleanly when ssh-keygen is absent), transient-file cleanup, unsafe-session-id rejection (path traversal, shell metacharacters), exact entry format incl. extra permissions, removal preserving other entries, no-op removal, and sweep grace-period behavior. `leaks --atExit`: 0 leaks.
- All 10 csshnpd unit tests pass; build clean.

## Would still need before merging (if ever revived)

- [ ] E2E: `sshnp --no-et` (openssh impl) against this daemon — tunnel session comes up with the ephemeral key, forced command + PermitOpen restrictions hold
- [ ] E2E: `sshnp --no-et --remote-sshd-port 2222` with a second sshd on 2222 — user session lands on 2222
- [ ] E2E: `sshnp --ssh-client dart` variant
- [ ] Verify the authorized_keys entry is removed ~15s after connect, and that a client connecting at second 14 stays connected
- [ ] OpenWRT: confirm ssh-keygen availability on the target image (dropbear-only images may need openssh-keygen; the daemon degrades gracefully without it) and dropbear's handling of the forced-command/PermitOpen entry syntax

### DIFF
--- a/packages/c/sshnpd/CMakeLists.txt
+++ b/packages/c/sshnpd/CMakeLists.txt
@@ -13,6 +13,7 @@ set(
   ${CMAKE_CURRENT_LIST_DIR}/src/daemon.c
   ${CMAKE_CURRENT_LIST_DIR}/src/daemon_utils.c
   ${CMAKE_CURRENT_LIST_DIR}/src/device_info.c
+  ${CMAKE_CURRENT_LIST_DIR}/src/ephemeral_key.c
   ${CMAKE_CURRENT_LIST_DIR}/src/file_utils.c
   ${CMAKE_CURRENT_LIST_DIR}/src/handle_npt_request.c
   ${CMAKE_CURRENT_LIST_DIR}/src/handle_ping.c
--- a/packages/c/sshnpd/include/sshnpd/ephemeral_key.h
+++ b/packages/c/sshnpd/include/sshnpd/ephemeral_key.h
@@ -1,0 +1,71 @@
+#ifndef SSHNPD_EPHEMERAL_KEY_H
+#define SSHNPD_EPHEMERAL_KEY_H
+
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdio.h>
+
+// The ephemeral tunnel key flow (parity with the Dart daemon's
+// startDirectSsh): for each ssh_request session the daemon generates a
+// throwaway ssh keypair, authorizes the public key in authorized_keys with a
+// restrictive entry (forced command + PermitOpen to the local sshd only),
+// returns the private key to the client in the session response, and removes
+// the authorized_keys entry shortly afterwards. The client uses it to bring
+// up the initial tunnel ssh session, inside which --remote-sshd-port style
+// forwards are made by sshd itself.
+
+/**
+ * @brief Generate a throwaway ssh keypair for one session by running
+ * ssh-keygen (no shell involved; the session id is validated first).
+ *
+ * The key files are written under directory (created 0700 if needed), read
+ * into memory and unlinked before returning.
+ *
+ * @param directory directory for the transient key files, e.g. $HOME/.sshnp
+ * @param use_rsa true for rsa-4096, false for ed25519 (the default)
+ * @param session_id the session id (uuid); also used in the file name
+ * @param private_key_pem receives the malloc'd private key file contents
+ * @param public_key receives the malloc'd public key line
+ * @return int 0 on success
+ */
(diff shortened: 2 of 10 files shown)


cconstab:
Parking this per discussion: `--no-et` tunnel mode is a pre-E2EE legacy path with no known users against C daemons, and everything it appeared to be needed for is already covered — default-mode `sshnp` works against csshnpd today, non-standard sshd ports are handled by daemon-side `--local-sshd-port` or `npt -rp <port>` (which is what NoPorts Desktop does). Given the security/maintenance surface (per-session authorized_keys writes, ssh-keygen exec, private keys in responses) we suggest ignoring this need for now. The implementation is complete and unit-tested on this branch and can be revived if a real use case appears. Dropped from `cconstab/c-2831-integration`; the mergeable stack for #2831 remains #2864 → #2865 → #2867.

PR #2870, opened 2026-08-27
feat: 'proxy:' root domain support in csshnpd (atServer connections via 443)

cconstab:
> [!IMPORTANT]
> **Ready for review**, but merge only after atsign-foundation/at_c#702 lands and `packages/c/cmake/atsdk.cmake` is bumped to an at_c trunk hash that contains it — the runtime proxy connection support lives in at_c, so merging this first would ship a `proxy:` option that cannot connect. The integration branch currently pins atsdk to the at_c PR branch for testing.

## What this does

Ports the Dart `at_lookup` `proxy:` root-domain convention to csshnpd. With

```
sshnpd ... --root-domain "proxy:proxy0001.atsign.org:443"
```

the daemon makes **no atDirectory lookups at all**: the address of every atServer is deemed to be the reverse proxy host and port, exactly as `SecondaryUrlFinder.findSecondaryUrl` does in Dart. This lets the daemon run on networks where only port 443 egress is allowed.

## How

Small change, because the C atclient already has the right hook: setting `atserver_host`/`atserver_port` on `atclient_authenticate_options` makes `atclient_pkam_authenticate` skip the atDirectory query entirely (`atclient.c` step 5). So:

- `main.c` root-domain parsing recognizes the `proxy:` prefix (port parsing follows `AtRootDomain.parse`: `proxy:host` → port 64, `proxy:host:port` → given port) and, in proxy mode, sets the atServer override on **both** the monitor and worker authenticate options
- the monitor/worker reconnect paths reuse the same option globals, so reconnections inherit the proxy address
- lookups of other atsigns' public keys (`verify_envelope_signature_from`, manager key caching) go through the daemon's own atServer connection via `plookup`, so no other code path touches the atDirectory — verified by grep, `atdirectory` is only referenced from the authenticate path
- `--root-domain` help text documents the new form

## Testing so far

- Build clean, all unit tests pass.
- Not yet run against a real proxy — needs the e2e checklist below.

- [ ] E2E: daemon starts and authenticates via a real reverse proxy on 443 (`proxy:proxy0001.atsign.org:443`); log shows `atServer connections will go via the reverse proxy ...` and no atDirectory contact
- [ ] Monitor reconnect (leave idle >40s, and force a network blip) still goes via the proxy
- [ ] Regression: plain `--root-domain root.atsign.org:64` and default behavior unchanged
- [ ] Combined with the #2831 stack (timeout/ESCR/policy) on the integration branch

### DIFF
--- a/packages/c/sshnpd/src/main.c
+++ b/packages/c/sshnpd/src/main.c
@@ -26,6 +26,7 @@
 #include <sshnpd/daemon.h>
 #include <sshnpd/file_utils.h>
 #include <sshnpd/run_srv_process.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -233,13 +234,15 @@ int main(int argc, char **argv) {
       }
       snprintf(root_host, root_host_size, "%.*s", (int)host_len, root_domain_str);
       const char *port_str = colon_pos + 1;
-      root_port = (uint16_t)atoi(port_str);
-      if (root_port == 0) {
+      char *port_end = NULL;
+      long port_val = strtol(port_str, &port_end, 10);
+      if (port_end == port_str || *port_end != '\0' || port_val < 1 || port_val > UINT16_MAX) {
         atlogger_log(LOGGER_TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "Root domain port is not a valid number: %s\n", port_str);
         atcommons_memlist_failure_free(&memlist);
         res = 1;
         return res;
       }
+      root_port = (uint16_t)port_val;
     } else {
       // no port specified, use the default port
       snprintf(root_host, root_host_size, "%s", root_domain_str);
--- a/packages/c/sshnpd/src/main.c
+++ b/packages/c/sshnpd/src/main.c
@@ -201,20 +201,38 @@ int main(int argc, char **argv) {
   }
   memset(root_host, 0, sizeof(char) * root_host_size);
   uint16_t root_port = 0;
-  if (params.root_domain != NULL) {
+
+  // A 'proxy:' prefix on --root-domain (e.g. "proxy:proxy0001.atsign.org:443")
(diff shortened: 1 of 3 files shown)


cconstab:
at_c PR #702 has merged — the atsdk pin here is now bumped to the at_c trunk merge commit (`f7303da5`) in 03ace325c, so this PR no longer has any external branch dependency. Verified locally: fresh dev-preset build fetches the pinned SHA and `sshnpd --root-domain "proxy:proxy0001.atsign.org:443"` takes the reverse-proxy path.

cpswan:
@cconstab do you need me to do a release of at_c (bumping all the Conan stuff for SBOMs)?

cconstab:
Yes please

cconstab:
> @cconstab do you need me to do a release of at_c (bumping all the Conan stuff for SBOMs)?

Yes please

packages/c/sshnpd/src/main.c
```diff
@@ -201,30 +202,50 @@ int main(int argc, char **argv) {
...
+  // A 'proxy:' prefix on --root-domain (e.g. "proxy:proxy0001.atsign.org:443")
+  // means there is no atDirectory: every atServer connection goes to the given
+  // reverse proxy host and port instead. This matches the convention of the
+  // Dart at_lookup package, and lets a daemon run where only port 443 egress
+  // is allowed.
+  bool via_atserver_proxy = false;
+  const char *root_domain_str = params.root_domain;
+  if (root_domain_str != NULL && strncmp(root_domain_str, "proxy:", strlen("proxy:")) == 0) {
+    via_atserver_proxy = true;
+    root_domain_str += strlen("proxy:");
```
JeremyTubongbanua, line 215: TIL you can have a `const char *`pointer and still move the pointer around! (`root_domain_str += strlen("proxy:");`). But still illegal to change the data inside.

JeremyTubongbanua (approved):

PR #2876, opened 2026-08-29
feat: C daemon + srv security hardening (M1–M8, M11)

cconstab:
Security hardening for the C `sshnpd` daemon and `srv` relay, from a review of the C daemon + at_activate. One commit per finding (M1–M8).

## Findings fixed (this repo)
| ID | Severity | Summary |
|----|----------|---------|
| M1 | Medium | Reject malformed signed envelopes before deref — missing/mistyped `signature`/`hashingAlgo`/`signingAlgo`/`payload` crashed the parent daemon (remote authenticated DoS) |
| M2 | Low/Med | Inverted `&&`→`||` NULL guards in envelope extraction |
| M3 | Medium | NUL-terminate the 4096-byte relay control buffer (attacker-controlled over-read via strtok/`%s`) |
| M4 | Medium | Cap concurrent relay sessions (`SRV_MAX_SESSIONS=128`) — unbounded thread/fd creation from the control channel |
| M5 | Medium | Open `authorized_keys` per call instead of `freopen`+lock on a shared FILE* (fixed `funlockfile(NULL)` crash + dangling global UAF) |
| M6 | Medium | Reap all children in a `waitpid` loop; async-signal-safe SIGCHLD handler (zombie/PID-exhaustion) |
| M7 | Medium | Scope pushed ssh keys with `from="127.0.0.1,::1"` (were unrestricted + persisted past revocation) |
| M8 | Medium | Run the relay via **self-exec** (`--__srv-worker`) instead of an in-process fork, so it no longer shares an address space with the daemon's private keys / authenticated atServer sockets. Secrets via env, non-secrets via argv (matches the Dart/`srv` contract). Self-exec only — no external binary to plant/replace. |
| M11 | Low | Normalize the `--policy-manager` atSign like managers; a non-canonical value silently denied all non-manager clients (fails safe, but the policy service wouldn't work) |

## Testing
- Verified working over a real relay: **ssh + ESCR + twinned keys + encrypted** session.
- Author confirms the branch works in their environment. Reviewers/QA may still want to spot-check the **npt** path and the **legacy `RV_AUTH`** (non-ESCR) path, since M8 reroutes the launch for all session types.
- [x] Confirm M7's localhost restriction doesn't break any deployment that bridges sshd over a non-loopback local address.
- [x] Decide whether M4's session cap (128) is the right bound; a session recv-timeout was deliberately **not** added (would kill idle tunnels — needs a keepalive design).

## Not included (upstream)
- **M9/M10** (atauth enroll-response NULL-derefs / unchecked decrypt) live in the at_c SDK and need a separate upstream PR.
- The review's **High** items (world-readable `.atKeys`, CRAM/OTP in argv) are also at_c SDK.

Build: `sshnpd` + `srv` compile and link clean; self-exec worker dispatch smoke-tested.

## Follow-up (tracked separately)

- **#2877** — M8 passes the daemon's APKAM/PKAM **identity** key to `srv` (via env) so srv can answer the relay's ESCR challenge. This branch keeps that (net improvement over the old in-process fork, which inherited the daemon's *entire* address space), but the end state is to keep the identity key out of srv entirely (dedicated ESCR key / delegated ephemeral key / daemon-mediated signing). #2877 also captures a cross-protocol signing-oracle concern: ESCR and atServer PKAM auth sign with the same key + algorithm, and are kept safe only by message-format non-collision. Applies to both C and Dart.

### DIFF
--- a/packages/c/srv/src/srv.c
+++ b/packages/c/srv/src/srv.c
@@ -28,6 +28,14 @@ static pthread_mutex_t session_mutex = PTHREAD_MUTEX_INITIALIZER;
 static int active_sessions = 0;
 static time_t idle_since = 0;
 
+// Upper bound on concurrent socket-to-socket sessions per srv process. The
+// relay control channel is attacker-reachable and each connect: line spawns a
+// thread + two sockets, so without a cap a malicious relay can exhaust
+// threads/fds. 128 is well above any legitimate fan-out for one daemon.
+#ifndef SRV_MAX_SESSIONS
+#define SRV_MAX_SESSIONS 128
+#endif
+
 static void session_started(void) {
   pthread_mutex_lock(&session_mutex);
   active_sessions++;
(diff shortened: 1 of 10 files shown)


JeremyTubongbanua (approved):
LGTM

PR #2879, opened 2026-08-29
fix: uptake atchops_rsa_encrypt output-buffer bound (at_c#709)

### LINKED ISSUE #701: atchops_rsa_encrypt() has no output-buffer-size parameter → caller stack overflow with oversized modulus
## Summary

`atchops_rsa_encrypt()` writes exactly the key modulus length into the caller's `ciphertext` buffer but provides **no way for the caller to declare that buffer's size**. Any caller using a fixed-size buffer (the documented convention is 256 bytes for RSA-2048) will overflow it if the supplied public key's real modulus is larger than expected — e.g. an RSA-4096 key writes 512 bytes.

## Where

`packages/atchops/src/rsa.c` — `atchops_rsa_encrypt()` (~L252):

```c
int atchops_rsa_encrypt(const atchops_rsa_key_public_key *public_key, const unsigned char *plaintext,
                        const size_t plaintext_len, unsigned char *ciphertext) {
  ...
  // ~L330: mbedtls writes ctx.len (= modulus length) bytes into ciphertext,
  // with no output-size bound
  mbedtls_rsa_pkcs1_encrypt(&rsa, mbedtls_ctr_drbg_random, &ctr_drbg_ctx,
                            plaintext_len, plaintext, ciphertext);
```

`packages/atchops/src/atchops/rsa.h` (~L49) documents the hazard as an assumption rather than enforcing it:

> `@param ciphertext the ciphertext buffer to populate, assumed to be 256 bytes long for 2048 RSA modulus`

Note the asymmetry: `atchops_rsa_decrypt()` **does** take a `plaintextsize` bound, so the safe pattern already exists in the same file — encrypt just never got it.

## Impact

`atchops_rsa_key_populate_public_key()` accepts keys with moduli larger than 2048 bits (no size restriction), and `atchops_rsa_key_public_key.n.len` reflects the real modulus. A caller that:

1. accepts a public key from an untrusted/semi-trusted source, and
2. passes a fixed 256-byte stack buffer to `atchops_rsa_encrypt()`

gets a stack buffer overflow with attacker-influenced length when the key is larger than RSA-2048. On embedded targets (ESP32, etc.) with no stack canaries/ASLR by default this is a plausible path to memory corruption / RCE.

This is not hypothetical — the NoPorts Arduino daemon hit exactly this (authenticated client supplies `clientEphemeralPK`; only the `"rsa2048"` type *label* was checked, not the modulus). We patched it at the call site (atsign-foundation/at_Arduino_NoPorts#10), but the API shape means every fixed-buffer caller shares the hazard, so it's worth fixing at the source.

## Suggested fix

Add an output-buffer-size parameter and bounds-check before writing, mirroring `atchops_rsa_decrypt`:

```c
int atchops_rsa_encrypt(const atchops_rsa_key_public_key *public_key, const unsigned char *plaintext,
                        const size_t plaintext_len, unsigned char *ciphertext, const size_t ciphertext_size);
```

with `if (mbedtls_rsa_get_len(&rsa) > ciphertext_size) return <error>;` before the encrypt call. (Breaking signature change — could alternatively be introduced as an `_ex` variant to preserve source compat.) A minimal non-breaking stopgap would be to at least reject moduli whose length exceeds the historically-assumed 256 bytes.

## Propagation note

This needs to flow into the Arduino packaging (`at_client_arduino`, published as `atsign/at_client` on the PlatformIO registry) and a version bump for downstream consumers like at_Arduino_NoPorts.

cconstab:
Uptake PR for atsign-foundation/at_c#709 (fixes atsign-foundation/at_c#701 — `atchops_rsa_encrypt` gains a `ciphertext_size` bound and optional `ciphertext_len` out-param; breaking signature change).

## Changes
- Both call sites in `handler_commons.c` pass the new arguments. These encrypt the session AES key/IV to the **client-supplied** `clientEphemeralPK` — the exact hazard shape from the issue — so the SDK-level bound now backs the existing call-site modulus pre-check (kept as defense in depth for its clearer error).
- atsdk pin: **PLACEHOLDER** on the at_c#709 PR branch head (`05295c51`) so CI can prove the uptake compiles.

## Before undrafting
- [x] at_c#709 merged
- [x] at_c release cut containing it (breaking change → v0.4.0 candidate)
- [x] Bump the pin here to that release's tagged SHA and drop the PLACEHOLDER comment

Verified locally: fresh dev-preset configure fetches `05295c51` and sshnpd builds clean against the new signature.

### DIFF
--- a/packages/c/cmake/atsdk.cmake
+++ b/packages/c/cmake/atsdk.cmake
@@ -12,9 +12,11 @@ if(NOT atsdk_FOUND)
     FetchContent_Declare(
       atsdk
       GIT_REPOSITORY https://github.com/atsign-foundation/at_c.git
-      # at_c v0.3.11: adds the enroll-via-proxy fix (at_c#705) on top of the
-      # 'proxy:' root server support and connection CRLF fix from v0.3.10
-      GIT_TAG 06dbc19ce2d1b682b7b51b3a2b5fb4dfdf1b3763
+      # PLACEHOLDER - update before merge: pinned to the at_c#709 PR branch
+      # head (atchops_rsa_encrypt output-buffer bound, breaking signature
+      # change). Replace with the at_c release SHA that contains #709 once it
+      # merges and is released.
+      GIT_TAG 05295c51791a17a8118361cd02d6b5e7ed6b845e
     )
   endif()
   FetchContent_MakeAvailable(atsdk)
--- a/packages/c/cmake/atsdk.cmake
+++ b/packages/c/cmake/atsdk.cmake
@@ -12,11 +12,9 @@ if(NOT atsdk_FOUND)
     FetchContent_Declare(
       atsdk
       GIT_REPOSITORY https://github.com/atsign-foundation/at_c.git
-      # PLACEHOLDER - update before merge: pinned to the at_c#709 PR branch
-      # head (atchops_rsa_encrypt output-buffer bound, breaking signature
-      # change). Replace with the at_c release SHA that contains #709 once it
-      # merges and is released.
-      GIT_TAG 05295c51791a17a8118361cd02d6b5e7ed6b845e
+      # at_c v0.4.0: atchops_rsa_encrypt output-buffer bound (at_c#709,
+      # breaking signature change) on top of the v0.3.11 proxy/enroll support
+      GIT_TAG 295c5d2b43264ac57432d4ceceee7955986d32de
     )
   endif()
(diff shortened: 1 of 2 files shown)


srieteja (approved):

JeremyTubongbanua:
Hang on don't merge yet, I think I might remember why we did it this way

JeremyTubongbanua:
Oh never mind, it's good to merge

PR #2886, opened 2026-09-01
feat: end-to-end tests ve

JeremyTubongbanua:
## TLDR

- `e2e_all` (the older bash tests) -- uses VE
- `core_tests` (superset of `e2e_all`) -- runs against production
- `policy_tests` -- uses VE
- `relay_tests` -- runs against production

## What I did

Context: we have 4 end-to-end test "things": `e2e_all`, `core_tests`, `policy_tests` and `relay_tests`. `e2e_all` and `core_tests` test the same things but `e2e_all` is in bash and `core_tests` is in Dart.

- `policy_tests` and `e2e_all` use a virtual environment now, instead of using prod Atsigns. `relay_tests` and `core_tests` use prod Atsigns. The reasoning is `policy_tests` need a VE so that policy records can be written concurrently across PRs. `e2e_all` is also pretty stable as we've had it around for a while. `core_tests` and `relay_tests` will remain as probes for flakiness to see if VE is the solution to flakiness or something fishy is going on in prod atServer land.
- Removed matrix in the workflow. **TLDR: reason I removed it is cause it will give more clues.** The reason I did this is so we can be intentional in which tests run what. In this case, VE is running in policy_tests while every other suite still uses prod Atsigns. My plan here is to introduce VE slowly and experiment with the flakiness. If flakes still happen even with the policy_tests using VE, my hypothesis tells me test harness or GitHub is to blame. If flakiness stops for policy_tests, but flakiness continues for all else, then VE could be the solution to removing flakes and atServer stress test/examination will be required.

## How I did it

- Claude Opus 4.6

feat: end-to-end tests ve


tests/e2e_all/scripts/common/common_functions.include.sh
```diff
@@ -763,10 +763,15 @@ runDockerDaemon() {
...
   containerName="e2e_all-$deviceName"
+  local addHostArgs=""
+  if [ "$atDirectoryHost" != "root.atsign.org" ]; then
+    addHostArgs="--add-host $atDirectoryHost:host-gateway"
+  fi
   local dockerRunCommand="sudo docker run \
     --rm \
     -d \
     --name \"$containerName\" \
+    $addHostArgs \
```
JeremyTubongbanua, line 774: Docker daemons need to be able to access host machine's IP to access the VE container also running on the host. This adds `<host machine's IP> vip.ve.atsign.zone` to `/etc/hosts`

JeremyTubongbanua:
@gkc can I get your opinions? My thinking is:

1. we'll still see if we're getting flakes with 50/50 prod/ve coverage... if things are all good, we can move completely to VE
2. policy_tests definitely needs VE so AtRecords don't overlap each other between concurrent runs
3. I decided to give VE to e2e_all *anyways* since it's been around for a long time. Since e2e_all and core_tests both test the same things (core_tests tests MORE things, though), we'll get to see the performance after a long while between e2e_all and core_tests. My hypothesis is we won't see flakes in e2e_all anymore, but we will still see flakes in core_tests which will hint at us that there's a problem with atServers; and the problem does not lie in GitHub actions or in the tests retry mechanisms itself.
4. I kept `relay_tests` using prod Atsigns so that we test `@rv_am` and ensure older clients/daemons still work with what's in production, and is an overall good sense check

gkc:
@JeremyTubongbanua sgtm

riic10 (approved):

PR #2892, opened 2026-09-03
fix: srv segfault on connection teardown (pthread_join clobbered tid)

### LINKED ISSUE #2891: csshnpd 1.1.C segfaults when using NoPorts Desktop and setup a Web session with openWRT
### Describe the bug

The csshnpd 1.1.C segfaults when I use either FireFox or Safari, when using NoPorts Desktop and setup a Web session with openWRT. Chrome doesn't segfault and I don't see any segfaults with any of the web browsers if I use the command line NPT to setup the web interface connection. This only seems to happen with macOS 26.6.2 when I use Safari or Firefox. Windows 11 Pro using the same connection setup information in NoPorts Desktop doesn't cause a segfaullt with Firefox.

SegFault when using Safari on macOS with NoPorts desktop allowing for a connection to localhost:8081:

[Sep 2, 2026, 11:54:13 AM PDT] daemon.err: sshnpd[5983]: [0;32m[INFO][0m 2026-09-02 18:54:13.583061 | srv - run | Starting threads
[Sep 2, 2026, 11:54:13 AM PDT] daemon.err: sshnpd[5983]: [0;32m[INFO][0m 2026-09-02 18:54:13.583159 | srv - run | Sending auth string
[Sep 2, 2026, 11:54:13 AM PDT] kern.info: [ 1504.801361] sshnpd[6203]: segfault at 7f57b5f6cb74 ip 00007f57b608c1be sp 00007f57b5f8f880 error 6 in libc.so[661be,7f57b603a000+58000] likely on CPU 0 (core 0, socket 0)
[Sep 2, 2026, 11:54:13 AM PDT] kern.info: [ 1504.803120] Code: 40 40 01 c3 48 83 cf ff 50 67 e8 df 09 00 00 55 53 48 89 fb 48 81 ec a8 00 00 00 8b 05 6b ee 01 00 85 c0 74 2f b8 01 00 00 00 <89> 43 3c f0 83 0c 24 00 67 e8 a9 1b 00 00 48 39 c3 74 6f 48 81 c4

SegFault when using Firefox on macOS with NoPorts desktop allowing for a connection to localhost:8081:

[Sep 2, 2026, 12:18:16 PM PDT] daemon.err: sshnpd[5983]: [0;32m[INFO][0m 2026-09-02 19:18:16.662521 | srv - run | Sending auth string
[Sep 2, 2026, 12:18:16 PM PDT] kern.info: [ 2947.778521] sshnpd[6578]: segfault at 7f8e632a3b74 ip 00007f8e6335a1be sp 00007f8e632c6880 error 6 in libc.so[661be,7f8e63308000+58000] likely on CPU 1 (core 0, socket 1)
[Sep 2, 2026, 12:18:16 PM PDT] kern.info: [ 2947.780227] Code: 40 40 01 c3 48 83 cf ff 50 67 e8 df 09 00 00 55 53 48 89 fb 48 81 ec a8 00 00 00 8b 05 6b ee 01 00 85 c0 74 2f b8 01 00 00 00 <89> 43 3c f0 83 0c 24 00 67 e8 a9 1b 00 00 48 39 c3 74 6f 48 81 c4

### Steps to reproduce

1. First I I create a profile in the NoPorts Desktop to setup a connection with port 8081 to port 80 on localhost.
2. Then I make the connection to the openwrtvm and I see a segfault in the logs on the openwrtvm 
3. And then I tried using the same connection information with Google Chrome, Firefox and Microsoft Edge
4. Safari and Firefox running on macOS caused the segfault.

### Expected behavior

I expected the NoPorts Desktop to work like the command line NPT was well as not cause a segfault if I used Firefox of Safari.

### Screenshots

_No response_

_No response_

### Additional context

_No response_

cconstab:
## What

Fixes the `csshnpd` segfault reported in #2891 (web sessions to an OpenWRT daemon crashing with Safari/Firefox but not Chrome).

## Root cause

In `socket_to_socket` teardown ([srv.c](https://github.com/atsign-foundation/noports/blob/c1.1.0/packages/c/srv/src/srv.c#L451-L459)):

```c
int retval = 0;
res = pthread_join(tid, (void *)&retval);
```

`pthread_join` writes the joined thread's result — a `void *`, 8 bytes — through that pointer, but `retval` is a 4-byte `int`. Disassembly of the shipped c1.1.0 binary confirms `tid` sits directly after `retval` on the stack (`retval` at `rbp-0x10c`, `tid` at `rbp-0x108`), so the join zeroes the **low 32 bits of `tid`** (side threads exit with `pthread_exit(NULL)`).

The very next check, `pthread_equal(threads[0], tid)`, then reports *not equal* even when the joined thread **was** `threads[0]` — so the code falls into `tidx = 0` and calls `pthread_cancel(threads[0])`: **cancelling the thread it just joined**. musl unmaps a joined thread's stack, including its `struct pthread`, so the cancel's `a_store(&t->cancel, 1)` writes to unmapped memory → SIGSEGV. The kernel dump in #2891 matches exactly: fault address = dead descriptor + 0x3c (musl's `cancel` field offset), faulting code bytes decode to musl `pthread_cancel`.

### Why only Safari/Firefox on OpenWRT

The corrupted comparison only picks the wrong thread when **side a** (the local service socket, `threads[0]`) exits first. Safari/Firefox teardown makes the local web server close first; Chrome and CLI `npt` close from the client/relay side first, where "not equal" happens to be the correct answer. glibc builds suffer the same 8-byte clobber but cache joined stacks instead of unmapping, so they corrupt silently instead of crashing.

## Changes

- `retval` is now a `void *`, so `pthread_join` writes within bounds (the actual crash fix)
- check the `read()` of the exited-thread id; on failure cancel + reap both threads instead of joining an uninitialized `pthread_t`
- check `pipe()` for failure instead of using uninitialized fds
- auth-string send failure now exits via the cleanup path instead of returning directly (previously leaked the pipe fds and both sockets on every failed session)

## Testing

- Builds clean (macOS clang, full `sshnpd` target)
- Root cause verified by disassembly of the shipped `sshnpd-linux-x64-musl` c1.1.0 release binary
- A musl x64 test build for the reporter's OpenWRT VM to verify the segfault is gone

Fixes #2891


cconstab:
Validation of the regression test on musl (Alpine, clang, same toolchain as `Dockerfile.musl`):

| platform | pre-fix (trunk) | with fix |
|---|---|---|
| linux/amd64 (Kim's platform) | **SIGSEGV** (exit 139) in the service-closes-first ordering | all pass |
| linux/arm64 | passes (stack layout happens to be benign there — the clobber is UB, blast radius is compiler/arch dependent) | all pass |

So the test reproduces the exact #2891 crash on x86-64 musl and confirms the fix. Bug archaeology: the `int retval` passed to `pthread_join` dates to the original srv prototype (37312b460, 2024-02-08); it became an armed segfault when 85dc1172d (2024-02-15) moved the join above the `pthread_equal` check; shipped in every csshnpd release since c0.1.1 (2024-06-24).

cconstab:
The real Colin here.. Tested fix by hand using 

Docker  + Alpine + http server (python) + sshnpd musl (c1.1.0)

Connects once then seg fault of the exec'd sshnpd

Built new sshnpd musl ran that in the above set up and things work as they should.

This would have been a tough bug to find without Claude sluthing..

To make sure we do not regress a test put in place too..

cconstab:
@kimdimick confirmed this fix is working on OpernWRT

srieteja:
Reviewed at `d3234a95e7a12f312d6b213275bd665e4999985b`. The `void *retval` fix is correct — a genuine 8-byte write into a 4-byte slot. Three things around it; patch for 2 and 3 at the bottom.

**1. The srv ctest step this PR adds can't catch the bug it guards.** I built the new test against both the pre-fix and post-fix `srv.c`, on both libcs, x86-64:

| x86-64 | pre-fix | post-fix |
| --- | --- | --- |
| glibc — `ubuntu-latest`, the exact command added at `c_unit_tests.yaml:31` | **Passed** | Passed |
| musl — `alpine:3.21`, same source | **SEGFAULT** | Passed |

So on the runner it targets the new step is green either way. And it's unsound rather than merely blind: pre-fix `tidx == 0`, `threads[1]` is never cancelled, and `socket_to_socket` returns 0 while side b still holds `&sides[1]` into a stack the next iteration recycles — green, crash and hang are all reachable.

One container step on the same runner closes it — this is verbatim what produced the SEGFAULT above:

```yaml
- name: srv Unit CTest (musl x86-64)
  run: |
    docker run --rm -v "$PWD/packages/c:/src" alpine:3.21 sh -c '
      apk add --no-cache build-base clang cmake git python3 &&
      cmake -S /src/srv -B /build -DSRV_BUILD_TESTS=ON -DCMAKE_C_COMPILER=clang \
        -DCMAKE_C_FLAGS="-Wno-error -pthread" &&
      cmake --build /build -j"$(nproc)" &&
      ctest --test-dir /build/tests --output-on-failure --timeout 10'
```

Worth narrowing `test_socket_to_socket_teardown.c:12` to musl **x86-64** either way — per the validation table arm64 doesn't reproduce.

**2. `test_socket_to_socket_teardown.c:142,149`** join `runner_thread` with both conns still open (closed only at :173,:176), so the side threads never write a tid and `srv.c:470`'s `read` never returns. Forcing the bridge-failure branch on musl x86-64: `Timeout` at 10.02s, swallowing the assertion output. With the patch below, `Failed` in 0.01s.

**3. `srv.c:392-396`** returns without freeing side a, while the new adjacent `pipe()` path tears down correctly. One-liner — but don't fold it into `exit:`, which would free `sides[1]` uninitialized (`mbedtls_net_free` only guards `fd == -1`, so a zeroed `side_t` closes fd 0).

Both fixes, verified on the PR head — full `srv` suite 3/3 on musl x86-64 with it applied:

```diff
diff --git a/packages/c/srv/src/srv.c b/packages/c/srv/src/srv.c
index 23c130be4..46fdd3d7e 100644
--- a/packages/c/srv/src/srv.c
+++ b/packages/c/srv/src/srv.c
@@ -392,6 +392,8 @@ int socket_to_socket(const srv_params_t *params, const char *auth_string, chunke
   res = srv_side_init(&hints_b, &sides[1]);
   if (res != 0) {
     atlogger_log(TAG, ERROR, "Failed to initialize connection for side b\n");
+    // side a is already up and sides is stack-local, so the caller cannot free it
+    srv_side_free(&sides[0]);
     return res;
   }
 
diff --git a/packages/c/srv/tests/test_socket_to_socket_teardown.c b/packages/c/srv/tests/test_socket_to_socket_teardown.c
index ad6acf3b6..ce2f4f3ea 100644
--- a/packages/c/srv/tests/test_socket_to_socket_teardown.c
+++ b/packages/c/srv/tests/test_socket_to_socket_teardown.c
@@ -139,15 +139,13 @@ static int run_iteration(bool close_service_first) {
   relay_conn = accept_with_recv_timeout(relay_listener);
   if (service_conn < 0 || relay_conn < 0) {
     printf("accept failed\n");
-    pthread_join(runner_thread, NULL);
-    goto cleanup_listeners;
+    goto cleanup_session;
   }
 
   // Prove both side threads are up and bridging in both directions
   if (expect_bridged(relay_conn, service_conn, "ping") != 0 ||
       expect_bridged(service_conn, relay_conn, "pong") != 0) {
-    pthread_join(runner_thread, NULL);
-    goto cleanup_listeners;
+    goto cleanup_session;
   }
 
   // Tear down one side; socket_to_socket must join the exited thread, cancel
@@ -167,6 +165,21 @@ static int run_iteration(bool close_service_first) {
   }
 
   failures = 0;
+  goto cleanup_listeners;
+
+cleanup_session:
+  // Both conns must be closed before joining: while either is open its side
+  // thread stays parked in mbedtls_net_recv and never writes its id to the
+  // pipe, so socket_to_socket never returns and the join blocks forever.
+  if (service_conn >= 0) {
+    close(service_conn);
+    service_conn = -1;
+  }
+  if (relay_conn >= 0) {
+    close(relay_conn);
+    relay_conn = -1;
+  }
+  pthread_join(runner_thread, NULL);
 
 cleanup_listeners:
   if (service_conn >= 0) {
```

cconstab:
@srieteja thanks — all three points taken, landed in 4f4a95798:

1. **musl x86-64 CI step**: added to `c_unit_tests.yaml` verbatim (with a comment explaining *why* the glibc run can't catch this class of bug). Agreed the ubuntu-only step was decorative for this regression — green pre-fix and post-fix.
2. **Test hang on failure branches**: applied your `cleanup_session` patch as-is. Good catch — an assertion failure would have surfaced as an opaque 10s timeout instead of its own message.
3. **side-a leak on side-b init failure**: applied, including keeping it out of `exit:` for the uninitialized-`sides[1]` reason you gave.

Also narrowed the test header comment to musl x86-64 per your note. Full srv suite re-verified green on macOS; an alpine/amd64 (qemu) run of the amended branch is in flight as a double-check, and the new CI step will exercise it on every PR from here on.

cconstab:
Double checked by hand with...

```
git clone -q https://github.com/atsign-foundation/noports.git /tmp/np-old && cd /tmp/np-old
git checkout -q trunk                                                    # pre-fix srv code
git checkout origin/fix/srv-pthread-join-segfault -- packages/c/srv/tests/  # the new test suite

docker run --rm -v "$PWD/packages/c:/src" alpine:3.21 sh -c '
  apk add --no-cache build-base clang cmake git python3 &&
  cmake -S /src/srv -B /build -DSRV_BUILD_TESTS=ON -DCMAKE_C_COMPILER=clang \
    -DCMAKE_C_FLAGS="-Wno-error -pthread" &&
  cmake --build /build -j"$(nproc)" &&
  ctest --test-dir /build/tests --output-on-failure --timeout 10'
```
Fails
```
The following tests FAILED:
	  1 - test_socket_to_socket_teardown (SEGFAULT)
Errors while running CTest
```

and 

```
cd /tmp/np-old && git checkout -q fix/srv-pthread-join-segfault
docker run --rm -v "$PWD/packages/c:/src" alpine:3.21 sh -c '
  apk add --no-cache build-base clang cmake git python3 &&
  cmake -S /src/srv -B /build -DSRV_BUILD_TESTS=ON -DCMAKE_C_COMPILER=clang \
    -DCMAKE_C_FLAGS="-Wno-error -pthread" &&
  cmake --build /build -j"$(nproc)" &&
  ctest --test-dir /build/tests --output-on-failure --timeout 10'
```

passes

```
Test project /build/tests
    Start 1: test_socket_to_socket_teardown
1/3 Test #1: test_socket_to_socket_teardown ...   Passed    0.01 sec
    Start 2: test_stream_encrypt
2/3 Test #2: test_stream_encrypt ..............   Passed    0.00 sec
    Start 3: test_twin_keys
3/3 Test #3: test_twin_keys ...................   Passed    0.00 sec

100% tests passed, 0 tests failed out of 3

Total Test time (real) =   0.01 sec
```

TY @srieteja

srieteja (approved):

PR #2897, opened 2026-09-04
feat(npt_flutter): online sync only

JeremyTubongbanua:
**- What I did**

 Updates NoPorts Desktop to use the remote atServer as the source of truth instead of synchronizing application data through a local secondary.

 - Configures AtClient for remote-only data operations and replaces the synchronization service with a no-op implementation.
 - Removes sync listeners, progress state, notifications, and related UI wiring.
 - Preserves remote favorites before mutations and performs profile key discovery directly against the atServer.
 
feat(npt_flutter): online sync only


CurtlyCritchlow (changes requested):
update change log please.

Issue #2866, opened 2026-08-26
OpenWRT csshnpd won't start or restart if the device Atsign in added to the manager list by mistake.


kimdimick:
### Describe the bug

OpenWRT csshnpd won't start or restart if the device Atsign in added to the manager list by mistake.

### Steps to reproduce

1. First I created a list of Atsigns for the manager entry.
2. Then I pasted them in 
3. And then by mistake I added the device Atsign to the manager list.
4.  This is the manager list I used: @navy1no01_np,@intense56,@blackbagel42,@blushing1263,@lead2jf01_np,@lead2jf02_np,@radium7vz01_np,@radium7vz02_np,@shaggy69,@snow59bat,@voluntary96,@85amateur55,@21cluster,@31snowprime,@55commercial,@navy1no02_np

5. The device Atsign is: @navy1no02_np

### Expected behavior

I expected the LuCI web interface to tell me I used the device Atsign in the manager list and that it wasn't acceptable.

kimdimick:
This is the log created when I included the device Atsign in the manager list:

[Aug 26, 2026 at 1:29:46 PM PDT] daemon.info: sshnpd[6536]: permitting open:
[Aug 26, 2026 at 1:29:46 PM PDT] daemon.info: sshnpd[6536]: localhost:22
[Aug 26, 2026 at 1:29:46 PM PDT] daemon.info: sshnpd[6536]: localhost:3389
[Aug 26, 2026 at 1:29:46 PM PDT] daemon.info: sshnpd[6536]: localhost:443
[Aug 26, 2026 at 1:29:46 PM PDT] daemon.err: sshnpd[6536]: [0;32m[INFO][0m 2026-08-26 20:29:46.596705 | sshnpd - main | Using root_host: "root.atsign.org" and root_port: 64
[Aug 26, 2026 at 1:29:46 PM PDT] daemon.err: sshnpd[6536]: [0;32m[INFO][0m 2026-08-26 20:29:46.597690 | atclient | Missing atServer host or port. Using root.atsign.org:64 atDirectory to find atServer address
[Aug 26, 2026 at 1:29:47 PM PDT] daemon.err: sshnpd[6536]: [0;32m[INFO][0m 2026-08-26 20:29:47.594418 | connection | Connected
[...]

PR #2899, opened 2026-09-04
fix: C daemon filter device Atsign from manager Atsign list

cpswan:
One part of fixing #2866 

**- What I did**

Filter device Atsign from manager Atsign list.

**- How I did it**

`agy` prompt:

```txt
if the device atsign is included in a list of the manager atsigns after the
-m flag then the daemon fails to start. draft a change set that filters the
device atsign from the list and warns in the logs
```

**- How to verify it**

Params tests have been expanded.

fix: C daemon filter device Atsign from manager Atsign list

### DIFF
--- a/packages/c/sshnpd/src/main.c
+++ b/packages/c/sshnpd/src/main.c
@@ -84,6 +84,7 @@ int main(int argc, char **argv) {
 
   int res = 0;
   atlogger_set_logging_stream(stderr);
+  atlogger_set_logging_level(ATLOGGER_LOGGING_LEVEL_INFO);
 
   // setup initial values for global variables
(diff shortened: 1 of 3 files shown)


JeremyTubongbanua (changes requested):
Edge case: `sshnpd -a @device -m @device <...>`

When every manager gets filtered out and no `--policy-manager` is set, the daemon starts in a silently broken state where nothing can authorize. That should be caught and rejected at parse time.

cpswan:
@JeremyTubongbanua that edge case should be addressed now

JeremyTubongbanua (approved):

PR #2904, opened 2026-09-07
feat(srv): carry tunnel traffic with at_chops AES-CTR, OpenSSL-backed

srieteja:
## What

`srv` tunnel traffic moves from `package:cryptography`'s byte-at-a-time `DartAesCtr` to at_chops' OpenSSL-backed AES-CTR — 55–67× on the primitive. Falls back to pure-Dart `DartAesCtr` when libcrypto is absent; both branches produce the same keystream.

Depends on at_chops `^3.7.0` (published 2026-09-16) for `AesCtrFfiCipher.updateView`. No git pins or `dependency_overrides` — everything resolves from pub.dev.

## Key changes

- **New `aes_ctr_transformer.dart`** — `createAesCtrTransformer` (a `DataTransformer`) and `createAesCtrChunkTransformer` (a `ChunkTransformer`). `setAesCtrTransformer` picks between them per `Side`.
- **Payload-carrying sides use `ChunkTransformer`'s zero-copy `updateView`** instead of a `StreamController` + subscription.
- **Dropped a discarded per-byte HMAC-SHA256** — the decrypt side already passed `Mac.empty`, so it was computed and thrown away.
- **ESCR relay skips the per-chunk mutex**, not the controller, so #2915's backpressure handling is untouched.
- **Backend resolution is logged once per process**, so a host silently falling back to pure Dart is visible in the logs.

## Reviewer attention

- **Native lifecycle is hand-driven** — one `EVP_CIPHER_CTX` per tunnel, with two disposal paths, both deliberate: the `ChunkTransformer` is disposed by socket_connector's `_closeSide`, the `DataTransformer` on `StreamController.onCancel`. The header comment records why neither `stream.map` nor `async*`/`finally` works — `finally` waits on *source* close and would leak on a downstream cancel. A failed `transform` disposes immediately and stays disposed.
- **Key/IV validation moved to setup time**, so a misconfig can't surface mid-stream.
- **The pure-Dart fallback is gated behind a `@visibleForTesting` cipher factory**, so it is actually exercised on FFI hosts. The FFI-vs-Dart comparison tests skip *visibly* on a host without libcrypto rather than passing vacuously.
- **`SrvImplInline.run` keeps the `aesD2C == null` backwards-compatibility branch**, where both directions share the C2D key *and* IV — a two-time pad. Carried over unchanged from trunk (same semantics, fewer lines); this PR neither introduces nor fixes it. Worth its own issue, but the diff touches these lines, so it's called out rather than left to be rediscovered in review.

## Interop

This is a wire-format change in all but name, so it's verified at three levels:

1. **Unit** — `_legacyTransform` reproduces the pre-swap `DartAesCtr` keystream and asserts the new transformer matches it, on both the encrypt and decrypt halves.
2. **Cross-version e2e** — `npe2e (relay_tests)` runs this build against **released** peers: client and daemon at `v5.13.0` / `v5.14.13` / current, self-relay at `v5.10.0` / `v5.14.13` / current, every combination. A keystream regression fails here.
3. **`e2e_all`** plus `npe2e (core_tests)` / `(policy_tests)`.

All green.

### DIFF
--- a/packages/dart/noports_core/lib/src/srv/srv_impl.dart
+++ b/packages/dart/noports_core/lib/src/srv/srv_impl.dart
@@ -3,7 +3,7 @@ import 'dart:convert';
 import 'dart:io';
 import 'dart:typed_data';
 
-import 'package:at_chops/at_chops.dart';
+import 'package:at_chops/at_chops_ffi.dart';
 import 'package:at_utils/at_utils.dart';
 import 'package:dartssh2/dartssh2.dart';
 import 'package:meta/meta.dart';
@@ -539,17 +539,6 @@ class SrvImplDart implements Srv<SocketConnector> {
     required InternetAddress relayAddress,
     required Duration timeout,
   }) async {
-    DataTransformer? encrypter;
-    DataTransformer? decrypter;
-    if (aesC2D != null && ivC2D != null) {
-      encrypter = createEncrypter(aesC2D!, ivC2D!);
-      if (aesD2C == null) {
-        // Backwards compatibility - use the same key & iv
-        decrypter = createDecrypter(aesC2D!, ivC2D!);
-      } else {
-        decrypter = createDecrypter(aesD2C!, ivD2C!);
-      }
-    }
     // client side
     InternetAddress localAddress = await resolveRequestedLocalHost();
     SocketConnector sc = await SocketConnector.serverToSocket(
@@ -559,12 +548,21 @@ class SrvImplDart implements Srv<SocketConnector> {
       portB: streamingPort,
       verbose: Platform.environment['SRV_TRACE'] == 'true',
       logger: ioSinkForLogger(logger),
-      transformAtoB: encrypter,
-      transformBtoA: decrypter,
       multi: multi,
       timeout: timeout,
       beforeJoining: (Side sideA, Side sideB) async {
         logger.info('beforeJoining called');
+        // One connection per call to this function, so it's safe to
+        // construct the transformer pair here rather than hoisting it.
+        if (aesC2D != null && ivC2D != null) {
+          setAesCtrTransformer(sideA, aesC2D!, ivC2D!);
+          if (aesD2C == null) {
+            // Backwards compatibility - use the same key & iv
+            setAesCtrTransformer(sideB, aesC2D!, ivC2D!);
+          } else {
+            setAesCtrTransformer(sideB, aesD2C!, ivD2C!);
+          }
+        }
         // Authenticate the sideB socket (to the rvd)
         if (relayAuthenticator != null) {
           logger.info(
@@ -871,8 +869,8 @@ class SrvImplDart implements Srv<SocketConnector> {
             socketIVD2C = socketIVC2D;
           }
 
-          sideA.transformer = createEncrypter(socketAESKeyC2D, socketIVC2D);
-          sideB.transformer = createDecrypter(socketAESKeyD2C, socketIVD2C);
+          setAesCtrTransformer(sideA, socketAESKeyC2D, socketIVC2D);
+          setAesCtrTransformer(sideB, socketAESKeyD2C, socketIVD2C);
 
           logger.info(
             '_runClientSideMulti (_clientSideEncryptedSocket)'
@@ -1340,16 +1338,6 @@ class SrvImplDart implements Srv<SocketConnector> {
   Future<SocketConnector> _runDaemonSideSingle({
     required InternetAddress relayAddress,
   }) async {
-    DataTransformer? encrypter;
-    DataTransformer? decrypter;
-    if (aesC2D != null && ivC2D != null) {
-      decrypter = createDecrypter(aesC2D!, ivC2D!);
-      if (aesD2C == null) {
-        encrypter = createEncrypter(aesC2D!, ivC2D!);
-      } else {
-        encrypter = createEncrypter(aesD2C!, ivD2C!);
-      }
-    }
     InternetAddress localAddress = await resolveRequestedLocalHost();
 
     bool verbose = Platform.environment['SRV_TRACE'] == 'true';
@@ -1365,7 +1353,16 @@ class SrvImplDart implements Srv<SocketConnector> {
       );
     }
     Socket sideASocket = await Socket.connect(localAddress, localPort);
-    Side sideA = Side(sideASocket, true, transformer: encrypter);
+    Side sideA = Side(sideASocket, true);
+    // One connection per call to this function, so it's safe to construct
+    // the transformer pair here rather than hoisting it.
+    if (aesC2D != null && ivC2D != null) {
+      if (aesD2C == null) {
+        setAesCtrTransformer(sideA, aesC2D!, ivC2D!);
+      } else {
+        setAesCtrTransformer(sideA, aesD2C!, ivD2C!);
+      }
+    }
     unawaited(
       socketConnector.handleSingleConnection(sideA).catchError((err) {
         logSink.writeln(
@@ -1380,7 +1377,10 @@ class SrvImplDart implements Srv<SocketConnector> {
       );
     }
     Socket sideBSocket = await Socket.connect(relayAddress, streamingPort);
-    Side sideB = Side(sideBSocket, false, transformer: decrypter);
+    Side sideB = Side(sideBSocket, false);
+    if (aesC2D != null && ivC2D != null) {
+      setAesCtrTransformer(sideB, aesC2D!, ivC2D!);
+    }
 
     // Authenticate the sideB socket (to the rvd)
     if (relayAuthenticator != null) {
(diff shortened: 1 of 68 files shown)


cconstab:
Oh my you have done so much ! Huge TY !

cconstab:
@srieteja — we've been running the cross-repo end-to-end this PR lists as not-done-here, on an integration branch built for exactly that: **`cconstab/srv-throughput-integration`** = trunk + this PR, with the local-path at_chops override swapped for a git dependency on at_client_sdk#2220's branch, plus two backpressure fixes the testing forced out of hiding (below), plus a `+integration` version stamp so test binaries are distinguishable from the 5.16.0 release. Feel free to use it — it's what produced all the numbers here.

Setup: iperf3 through an npt tunnel, macOS client (Apple Silicon, Homebrew libcrypto) → own srvd → sshnpd on Linux (libcrypto.so.3), authenticated relay, identical path for every row. Receiver-side numbers throughout — on trunk the sender-side report is meaningless (see below).

## Net result — trunk vs integration branch, same path, all hops

| direction | trunk | integration (this PR + fixes) | improvement |
|---|---|---|---|
| forward | 233 Mbit/s | **1.84 Gbit/s** | **7.9×** |
| reverse (`-R`) | 170 Mbit/s | **1.35 Gbit/s** | **7.9×** |

Equally important, the honesty of the pipe: on trunk the iperf *sender* reported 1.81 Gbit/s forward and 3.29 Gbit/s reverse against those 233/170 real — i.e. ~1.5 GB and ~3.8 GB accumulated in srv/srvd heaps per 10-second run. On the integration branch sender and receiver agree within ~2% in both directions and RSS stays flat on all five processes.

The 7.9× real-tunnel gain vs the 55–67× primitive microbenchmark is Amdahl at work — with the cipher fast, the pipeline's other costs (socket copies, event loop) become the ceiling. The old cipher was the dominant cost; now it isn't.

## How it got there

| configuration | iperf receiver | iperf sender reported |
|---|---|---|
| trunk baseline, all hops (old cipher, no backpressure) | 233 / 170 Mbit/s | 1.81 / 3.29 Gbit/s (!) |
| this PR's srv on client, 5.15.2 release daemon-side, public relay (WAN — different path) | 3.85 Mbit/s | 9.46 Gbit/s (!) |
| all hops on this PR, own srvd, + socket_connector#37 | 1.12 Gbit/s | 2.01 Gbit/s |
| + noports#2915 (all fixes, all hops) | **1.84 Gbit/s** | 1.88 Gbit/s |
| same, reverse direction (`-R`) | 1.35 Gbit/s | 1.36 Gbit/s |

The sender/receiver gaps were real bytes: `Socket.add()` buffers unboundedly, so e.g. the ~11 GB difference in the WAN row was sitting in srv's heap. Two fixes came out of chasing that:

- **atsign-foundation/socket_connector#37** — flush-gated backpressure in the relay loop (4 MiB high-water mark per direction). Notably your transformer's `onPause`/`onResume` forwarding was load-bearing for this — the pause propagates through `createAesCtrTransformer` exactly as designed.
- **noports#2915** — the relay auth wrappers (srv authenticator + both srvd verifiers) re-emit the socket through a `StreamController` with no pause forwarding, which silently ate the backpressure one layer up. Independent of this PR, lands on its own.

With both in place, throughput *rose* from 1.12 → 1.84 Gbit/s — the GC churn from unbounded buffering was itself a bottleneck. Bonus datapoint for the interop criterion: the WAN row was effectively a mixed tunnel (new cipher client end, old `DartAesCtr` daemon end) through a production relay, and it carried traffic correctly — slowly, but correctly.

That closes out the e2e + throughput items this PR's description lists as not done here: real tunnel, authenticated relay, both directions, mixed ends, with libcrypto present on both hosts. (The scrubbed-`LD_LIBRARY_PATH` pure-Dart-fallback e2e remains untested.)

cconstab:
The headline comparison, now with both sides measured: **full stack on a single M1 Mac, everything on loopback** — sshnpd, own srvd, npt, both srvs; same machine, same test (`iperf3 -t 10 -R`), network removed from the equation entirely.

| stack | tunnel throughput (receiver) | sender vs receiver |
|---|---|---|
| release 5.15.2 (old cipher, no backpressure) | 343 Mbit/s | 750 vs 343 Mbit/s — 2.2× apart, ~500 MB buffered in-process |
| **integration branch (this PR + backpressure fixes)** | **9.76 Gbit/s** | **within 0.3%** (9.79 sender) |
| control: raw loopback, no tunnel | 85.8 Gbit/s | — |

**28.5× end-to-end**, and the tunnel now delivers ~11% of raw loopback throughput vs ~0.4% before.

Recordings: [integration branch](https://asciinema.org/a/F7RmrnwuomwCAEwc) · [release 5.15.2 + control run](https://asciinema.org/a/5qEvV4y4K8nBRVyA)

What the 10 Gbit/s run establishes beyond the ratio:

- **The stack's own ceiling is ~10 Gbit/s.** Every byte crosses six socket hops and four AES-CTR passes (encrypt + decrypt at each srv) and the pipeline sustains ~1.2 GB/s — so the 1.84 Gbit/s cross-machine figure above was the network path, not the software.
- **Backpressure is not the bottleneck even at loopback speeds** — sender/receiver converge at 10 Gbit/s with the default 4 MiB high-water mark, which retires the stop-and-go tuning concern noted in socket_connector#37. Steady 9.0–10.1 Gbit/s intervals, no sawtooth.
- This is also the easiest configuration for a reviewer to reproduce: one machine, no infrastructure.

cconstab:
### x64 cross-machine result: 3.5–4.0 Gbit/s

Tested the current (ChunkTransformer) form of this PR end-to-end on x64 Linux, both tunnel ends on the FFI cipher:

| stack (same path, x64 Linux, cross-machine) | tunnel throughput |
|---|---|
| software baseline (pure-Dart cipher) | ~450 Mbit/s |
| **this PR — FFI + ChunkTransformer, both ends** | **3.5–4.0 Gbit/s** |

≈ **8×**, with memory bounded and no crashes/hangs.

**What was tested:** integration branch `cconstab/srv-ffi-chunk-integration` = trunk + this PR, with git overrides pinning:
- `socket_connector` → atsign-foundation/socket_connector#39 (`feat/chunk-transformer`), which now also carries the 2.5.1 flush-gate fix (merged from trunk) — so ChunkTransformer runs on the fixed backpressure, no `StreamSink is bound` crash and no paused-relay hang.
- `at_chops` → at_client_sdk trunk (3.7.0, `updateView`).

**One gotcha worth flagging for the interop story:** the first run showed only ~450 Mbit/s because the daemon end was still an old `sshnpd` spawning an old `srv` — i.e. one end on the pure-Dart cipher. A tunnel is only as fast as its **slowest cipher end**, so a single non-FFI (or missing-libcrypto → silent pure-Dart fallback) end caps the whole thing. Rebuilding the daemon side took it 450 Mbit/s → 3.5–4.0 Gbit/s. Since the fallback is silent, this is easy to hit in the field; might be worth a one-line log of which cipher path each `srv` resolved to.

The `SocketException: reset by peer` at the end of each iperf3 run is benign teardown (iperf closing with data in flight), seen on both ends — unrelated to this PR.

3.5–4.0 Gbit/s is likely path/CPU-bound here rather than cipher-bound (the same FFI cipher did 9.76 Gbit/s on M1 loopback), so there's headroom above this number on a faster path.

srieteja:
> might be worth a one-line log of which cipher path each `srv` resolved to

Added in 3c712548d. 

**What it does**
- Pure-Dart fallback → `warning`, naming libcrypto *and* the cost: traffic is still encrypted correctly, but a tunnel runs at the speed of its slowest end, so this caps the link even when the far end has libcrypto.
- FFI path → `info`.
- Logged at both points that actually resolve a cipher, so the tunnel data path, the control channel and the inline path are all covered — `createAesCtrTransformer` resolves inside its returned closure at *stream* time, so building the transformer alone isn't what logs.

**Once per process, not per connection.** Worth stating why that's safe rather than just convenient: `AtPqc.aesCtrStreamCipher` returns null purely on `_aesCtrSupported`, a `static final` probed once, and a bad key/IV makes `AesCtrFfiCipher.fromLib` *throw* rather than return null. So a null means libcrypto and nothing else, and the answer can't flip mid-process — the latch can't report one backend and hide a different one.

JeremyTubongbanua (commented):
I think just update the noports_core version and CHANGELOG? or do that in another PR

cconstab (approved):

srieteja:
> I think just update the noports_core version and CHANGELOG? or do that in another PR

Let me do that in another PR @JeremyTubongbanua

packages/dart/noports_core/lib/src/srv/srv_impl.dart
```diff
@@ -269,42 +269,12 @@ class SrvImplInline implements Srv<SSHSocket> {
...
-        decrypter = (Stream<List<int>> stream) {
-          return algorithm.decryptStream(
-            stream,
-            secretKey: sessionAESKeyD2C,
-            nonce: sessionIVD2C,
-            mac: Mac.empty,
-          );
-        };
-      }
+      encrypter = createAesCtrTransformer(aesC2D!, ivC2D!);
```
JeremyTubongbanua, line 272: My agent found a P2 issue regarding socket teardown not disposing every cipher

For agents:

```
The new FFI transformer requires its stream to close or its subscription to be cancelled. Existing consumers do
 neither for outgoing encryption:

 - WrappedSSHSocket.close() and destroy() close only the underlying socket, leaving the encryption controller
   open.
 - Multi-session control channels likewise destroy the socket without closing controlSink or cancelling its
   encryption subscription.

 Consequently, native cipher contexts and buffers survive tunnel shutdown, relying on eventual garbage
 collection rather than deterministic disposal. Retained wrappers or heartbeat timers can keep them alive
 longer, increasing memory usage in long-running processes.

 Verified: Three targeted tests confirmed that the cipher remains usable after inline close(), inline destroy(),
 and daemon-multi connector shutdown. Wire controller/subscription cleanup into these teardown paths.
```

srieteja: Will be included in the PR that updates the version and CHANGELOG

PR #2915, opened 2026-09-09
fix: forward backpressure through relay auth stream wrappers

cconstab:
## Problem

The relay authenticator (`srv/relay_authenticators.dart`) and both relay auth verifiers (`srvd/relay_auth_verifiers.dart`) wrap the raw socket: they run their own `socket.listen(...)` and re-emit post-auth data into a `StreamController`, handing that controller's stream onward. The controllers were created without `onPause`/`onResume` wiring, so when the downstream consumer pauses its subscription, the raw socket keeps being read at full speed and the backlog accumulates in the controller without bound — the TCP window never closes and the far-end writer is never throttled.

This silently defeats any backpressure applied by the consumer. Observed with iperf3 through an authenticated relay (with socket_connector's backpressure fix, atsign-foundation/socket_connector#37, already in place): sender held a steady 2.01 Gbit/s while the receiver measured 1.12 Gbit/s — ~1 GB accumulated in srvd's heap in 10 seconds, inside the auth verifier's controller, exactly one layer above where socket_connector's pause stopped.

## Fix

Forward `onPause`/`onResume` from each wrapper's controller to the socket subscription, in all three places (srv-side authenticator, and both srvd-side verifiers: challenge/response and signature). Backpressure now propagates through the auth layer to the socket, and TCP throttles the sender.

With this fix plus socket_connector#37 on every hop, the same run converges: sender 1.88 Gbit/s vs receiver 1.84 Gbit/s (1.36/1.35 in reverse), RSS flat on srv, srvd and sshnpd throughout.

## Notes

- Independent of #2904; this bug predates it and this PR can land on its own. It only becomes *visible* at high throughput, which is why nobody hit it at the old cipher's rates.
- Analyzer output unchanged from trunk; all 380 noports_core tests pass.
- One further instance of the same pattern was found and deliberately left alone: `WrappedSSHSocket` (`srv_impl.dart:347`) pipes its encrypter with `encrypted.listen(underlyingSink.add)` — sshnp's in-process SSH path. A naive `addStream` fix would break its `flush()` (StateError while a sink is bound), so it deserves its own change.

### DIFF
--- a/apps/admin/admin_api/pubspec.lock
+++ b/apps/admin/admin_api/pubspec.lock
@@ -616,7 +616,7 @@ packages:
       path: "../../../packages/dart/noports_core"
       relative: true
     source: path
-    version: "6.14.0"
+    version: "6.14.1"
   openssh_ed25519:
     dependency: transitive
     description:
--- a/packages/dart/noports_core/CHANGELOG.md
+++ b/packages/dart/noports_core/CHANGELOG.md
@@ -1,3 +1,11 @@
+# 6.14.1
+
+- fix: the relay auth stream wrappers (the srv-side relay authenticator and
+  both srvd-side relay auth verifiers) now forward pause/resume from their
+  `StreamController` to the socket subscription, so backpressure applied by
+  the consumer of the wrapped stream reaches the socket and closes the TCP
+  window instead of buffering without bound in the wrapper
+
 # 6.14.0
 
 - fix: share event logging config once as cached key instead of on every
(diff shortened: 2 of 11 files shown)


gkc (changes requested):
lgtm

version bumps (for both the noports_core and sshnoports packages) and changelog entries please Colin

cconstab:
Done in b2409b99d — noports_core bumped to 6.14.1 and sshnoports to 5.16.1 (patch bumps: fix only, no API change), with changelog entries in each package, and the generated `version.dart` files bumped to match so `--version` reports the new versions.

gkc (approved):

PR #2916, opened 2026-09-09
fix: install root CA certificates in published runtime container images

cconstab:
## What I did and why

- Follow-up to #2907: new Dart versions no longer carry root certs, which is why the `noports_e2e_all_base_runtime` image was rebuilt with the `ca-certificates` package (and the e2e_all base runtime Dockerfile got it in 3e3a4332c).
- The production images built by `dockerhub_sshnpd.yml` (sshnpd, srvd, sshnp, npt, npp, at_activate, sshnpd-slim) have the same exposure: their `debian:stable-slim` runtime stages ship **zero** root certs (verified: `/etc/ssl/certs` is empty and `ca-certificates` is not installed in the pinned base), so the Dart binaries have nothing to validate atServer TLS connections against.
- This PR applies the same fix to those Dockerfiles:
  - install `ca-certificates` in every debian-slim runtime stage
  - for the scratch-based `Dockerfile.sshnpd-slim`, install `ca-certificates` in the build stage and explicitly `COPY /etc/ssl/certs` + `/usr/share/ca-certificates` into the final image — the currently pinned `dart:3.13.3` still provides certs via `/runtime/`, but newer Dart images drop them, so this protects the next dependabot dart bump
  - sync `tests/npe2e/tools/dockerfiles/Dockerfile.base.runtime` with the already-pushed base image (rebuilt with `ca-certificates` in #2907) so the file in git matches the published image
- Verified locally: built `Dockerfile.srvd` from this branch; the runtime image now has 301 certs in `/etc/ssl/certs` including `ca-certificates.crt`, and `srvd --help` runs.

fix: install root CA certificates in published runtime container images

### DIFF
--- a/packages/dart/sshnoports/tools/Dockerfile
+++ b/packages/dart/sshnoports/tools/Dockerfile
@@ -28,7 +28,7 @@ COPY --from=buildimage /app/packages/dart/sshnoports/bundles/core/docker/.startu
 RUN \
   set -eux ; \
   apt-get update ; \
-  apt-get install -y adduser openssh-server sudo iputils-ping iproute2 ncat telnet net-tools nmap iperf3 traceroute vim ; \
+  apt-get install -y adduser ca-certificates openssh-server sudo iputils-ping iproute2 ncat telnet net-tools nmap iperf3 traceroute vim ; \
   addgroup --gid ${GROUP_ID} ${USER} ; \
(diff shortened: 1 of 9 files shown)


cconstab:
reviewed by hand.. Close call... Maybe we should have some canary srvd's we use on a daily basis ?

JeremyTubongbanua (approved):

PR #2929, opened 2026-09-17
fix(sshnpd-slim): ship libcrypto so AES-CTR uses the OpenSSL FFI path

srieteja:
## What

`sshnpd-slim` could never use the AES-CTR OpenSSL FFI path from #2904, so the daemon always ran the slow pure-Dart cipher — silently.

The image is `FROM scratch` and takes its whole userland from the dart image's `/runtime/`, which ships no libcrypto. Scratch also has no `ld.so.cache`, so at_chops' bare `libcrypto.so.3` candidate had nothing to resolve against either.

## How

Stage Debian's `libcrypto.so.3` (and its zlib dependency) in the builder, copy them into the scratch stage, and point at_chops at libcrypto by absolute path via `AT_CHOPS_LIBCRYPTO_PATH`.

Same shape as 928c968da, which hand-copies `/etc/ssl/certs` for the same reason: `/runtime/` doesn't supply something we need, so we supply it.

**The rationale is inline in the Dockerfile**, at the code it explains — why the glob instead of a package name, why this stage, why the ordering matters, why the assertions are fatal, why deps go where `/runtime/` puts libc, and the alternative that was rejected.

Costs ~4–5 MB (~3 MB on armv7). The image already embeds OpenSSL statically inside `ssh`/`ssh-keygen`; a named `.so` is at least visible to scanners and to `sbom: true`.


srieteja:
Rebased onto \`st/ci-docker-build-check\` (#2931) so the new build gate actually exercises this Dockerfile before merge — previously #2931 built trunk's unfixed Dockerfile and this PR built nothing.

\`docker_build_check.yml\` shows in the diff only until #2931 merges. Merge #2931 first; this then reduces to the single Dockerfile change.

srieteja:
### Design rationale

Kept here rather than in the Dockerfile, which only carries the traps you'd hit editing those lines.

**Why stage libcrypto at all.** The scratch stage takes its whole userland from the dart image's `/runtime/`: loader, libc, libm, libpthread, libdl, libresolv, librt, libnss_dns, CA certs — no libcrypto. Scratch also has no `ld.so.cache`, so at_chops' bare `libcrypto.so.3` candidate has nothing to resolve against either. Both doors are shut, so the daemon always took the pure-Dart path.

**Why from this stage.** It's the same stage `/runtime/` comes from, so the glibc ABI matches by construction. Sourcing libcrypto from any other image breaks that guarantee silently. libcrypto only — at_chops' FFI path never references libssl.

**Why a glob, not a package name.** `dpkg -L libssl3` breaks on the trixie `libssl3t64` rename; `ldconfig -p` isn't on a non-root PATH and reports the cache rather than the file; a hardcoded `$TARGETARCH` triplet is three paths to hand-maintain. The glob is arch- and package-name-independent, and asserted non-empty so it can't fail open.

**Why deps can't sit beside libcrypto.** `AT_CHOPS_LIBCRYPTO_PATH` locates libcrypto; libcrypto's own `DT_NEEDED` are resolved by ld.so, which has no reason to look in `/usr/local/at/lib` and no cache to consult. They go into whichever directory `/runtime/` already keeps `libc.so.6` in — provably searched in this image, since the daemon resolves libc from it today. The path is derived at build time, so it's correct on every arch without a hardcoded triplet.

**Why `AT_CHOPS_LIBCRYPTO_PATH` over `LD_LIBRARY_PATH`.** Deterministic without an `ld.so.cache`, scoped to the one consumer, and on failure at_chops logs a named warning instead of degrading silently. `srv_impl.dart` spawns `srv` with `includeParentEnvironment: true`, so one `ENV` covers the daemon and the spawned `srv`.

**Why the checks are fatal.** This bug class degrades silently — the daemon keeps working, just slower. A base-image change that breaks the FFI path has to break the build. The scratch stage has no shell and can assert nothing, which is why it all lives in the builder. Please don't soften them to warnings.

**Why an explicit `STAGED_EXTRAS` rather than auto-staging.** Auto-staging whatever `readelf` asks for turns a loud assertion into "ship whatever it wants." An explicit list makes a new shared dependency show up in review as a diff. It found `libz.so.1`, then `libzstd.so.1` (OpenSSL 3.2+ links zstd for TLS certificate compression).

**Rejected: a shared `.so` from `static-openssh.sh`.** Dropping `no-shared` leaves the existing `--with-ldflags=-static` ssh link at the mercy of linker search order, ships Jan-2024 OpenSSL 3.2.1 with the patching burden on us, and saves nothing — we'd still stage, COPY and env-var the result.

**Trade-offs.** ~4–5 MB on amd64/arm64, ~3 MB on armv7, against a ~40–50 MB image. "Slim" here means no shell, no package manager, no distro userland — not zero shared libraries; the image already loads libc and friends from `/runtime/`. On security this is a net gain in transparency: the image *already* embeds OpenSSL 3.2.1 statically inside `ssh`/`ssh-keygen`, where no scanner sees it and nothing triggers a rebuild. A named `libcrypto.so.3` is visible to filesystem scanners and to the workflow's `sbom: true` provenance.

srieteja:
Built green on all three platforms via #2931 (run [35259964510](https://github.com/atsign-foundation/noports/actions/runs/35259964510)), which carries these commits plus the PR build gate.

Staged `libcrypto.so.3` `DT_NEEDED`, per platform:

| Platform | NEEDED |
|---|---|
| linux/amd64 | `libz.so.1 libzstd.so.1 libc.so.6` |
| linux/arm64 | `libz.so.1 libzstd.so.1 libc.so.6 ld-linux-aarch64.so.1` |
| linux/arm/v7 | `libz.so.1 libzstd.so.1 libc.so.6 ld-linux-armhf.so.3` |

The staged `libz.so.1` and `libzstd.so.1` need only libc and the loader, so the closure is closed rather than checked one level deep.

Two notes for the record:

- `libz.so.1` and `libzstd.so.1` were found by the gate, not by reading. Debian links OpenSSL 3.2+ against zstd for TLS certificate compression; neither library is in `/runtime/`. Without them the staged libcrypto would have been unloadable by `dlopen` — the exact silent pure-Dart fallback this PR exists to remove. A build-free merge would have shipped a no-op.
- `libatomic.so.1` on armv7, flagged as the likely failure when this was planned, did not appear. `STAGED_EXTRAS` stays at two entries; the assertion decides, so nothing speculative is in the image.

Still unproven, and deliberately so: that at_chops *selects* the FFI path at runtime. `_logCipherPath` fires during tunnel setup, so that needs a live tunnel from this image.

JeremyTubongbanua (dismissed):
i think the comments could be less verbose

srieteja:
The latest commits cut down the comments to a minimum @JeremyTubongbanua

cconstab (approved):

JeremyTubongbanua:
nice

PR #2931, opened 2026-09-17
ci: build the sshnpd container images on pull requests

srieteja:
## What

Builds the seven published sshnpd images on pull requests, with `push: false`.

## Why

`dockerhub_sshnpd.yml` runs only on trunk pushes and tags, and it builds with `push: true`. So the first build of a changed Dockerfile *was* the release build — breakage reached trunk, and the registry, before anyone saw it.

This came up in #2929: a fix to `Dockerfile.sshnpd-slim` had no way to be validated before merge.

## Matrix

| Coverage | Why |
|---|---|
| All 7 images, `linux/amd64` | Catches the cheap failures — Dockerfile syntax, a dependency that no longer resolves, a lockfile `--enforce-lockfile` rejects |
| `sshnpd-slim` also `linux/arm64,linux/arm/v7`, native arm runner | It's the only `FROM scratch` image and the only one that cross-builds static OpenSSH, so it's the only one whose build is genuinely arch-sensitive. The rest are the same `debian:stable-slim` runtime stage on every arch |

Image names and Dockerfile paths were diffed against `dockerhub_sshnpd.yml` — every published image is covered, no path drift.

## Cost and safety

- Path-filtered to the sources these images build from, plus `concurrency` cancel-in-progress and a `gha` cache.
- `permissions: contents: read`, no registry credentials, `push: false` — a fork PR can neither push nor reach secrets.

If the arm leg proves too slow in practice, dropping that one matrix row is a one-line change — but note it forfeits exactly the coverage that motivated this.

## Not covered

This proves the images *build*. It does not run them, so it would not catch a runtime regression such as a library present but never loaded. Worth adding later if it earns its keep.


srieteja:
Also carries #2929's commits (cherry-picked) so this PR's CI actually builds the fixed Dockerfile. Both branches sit on trunk independently; whichever merges first, the other drops the duplicate patch on rebase.

The gate has already paid for itself here — it caught missing `libz.so.1` and `libzstd.so.1` in the staged libcrypto, which would otherwise have shipped as a silently slow daemon.

cpswan:
@srieteja could the existing workflow not be modified to run with push set to false for certain conditions and true for the existing conditions? Then we would have far less mostly duplicate code to maintain.

